PR #9, opened 2017-09-12
remove extra `)`

Licenser:
There was an extra `)` that snuck in there, I hunted it down and caught it. the `)` wasn't hurt in the process it's now released into the wilds and lives a happy life in some Lisp code.


bogdandrutu (approved):

codefromthecrypt:
:) very entertaining

Issue #8, opened 2017-09-08
Do we need sampling to be more than one bit in trace-parent to define priority


codefromthecrypt:
The first draft of the http header spec had 140 comments in various areas. Not all came to closure. One that remained was if we do anything about vendor-specific sampling priority data, or relegate that as a stakeholder in the key/value aka baggage functionality.

Here's background from the original pull request (#1)

[One line of comment started from @bhs](https://github.com/TraceContext/tracecontext-spec/pull/1#discussion_r110552519) 
This discussion was about flags and why we had initially 4bytes allocated, more room implied problems with interpretation such as endianness. Bogdan had reserved extra space as some proposed they needed room for sampling. This was reduced to a single byte and a question raised about if we should extend if for sampling priority. This particular line of comment went quiet.

[Another started afterwards from @SergeyKanzhelev](https://github.com/TraceContext/tracecontext-spec/pull/1#discussion_r124341654)
Sergey suggested that since sampling implementation would be vendor-specific, why not use a separate header? For example, not all vendors do consistent sampling per trace (ex some do sampling again at span level). Yuri mentioned benefits of a consistent trace-level algo. Sergey agreed with this and also that a sample bit as a force-trace operation is useful. However, he believed this should be a "debug" flag, not a sampling one. He suggests the sampling info should be vendor-specific.

[After some chatter from Adrian,  @SergeyKanzhelev](https://github.com/TraceContext/tracecontext-spec/pull/1#issuecomment-312704758) maintained "I still think it would make sense to detach sampling/debug flag from the identity information."

[The last comment on this sampling was from @bhs](https://github.com/TraceContext/tracecontext-spec/pull/1#issuecomment-315392209), when mentioned that one can stash the sampling rate (or reciprocal sampling rate) if arbitrary key/value propagated tags were supported somehow, and that presence of a sampled bit could be ignored/overlooked when they don't match a model.

tylerbenson:
Overall I like the direction of this project.  I think there would be significant value in keeping the core context propagation simple, but allow it to be extended for additional details that aren't central to maintaining a consistent trace flow.  Examples that have been brought up are:
* a debug flag
* sampling flag (currently in the main header)
* sampling priority
* other vendor specific things (account scoping might be an example)

I think being able to handle these kind of things should be in the scope of this spec, but I am open to discussion on whether that is handled in one header or a separate standardized header (or set of headers).

bogdandrutu:
[...]

PR #12, opened 2017-09-15
Make TraceOptions optional.

bogdandrutu:
Based on the comments from the PR https://github.com/TraceContext/tracecontext-spec/pull/1 and the issue https://github.com/TraceContext/tracecontext-spec/issues/4 seems that not everybody uses the trace-options. 

When the trace-options are missing every bit has to have a default value that is used. In general this will not be missing but in cases where users are concerned about extra bytes on wire not sending the trace-options is acceptable.

### DIFF
--- a/HTTP_HEADER_FORMAT.md
+++ b/HTTP_HEADER_FORMAT.md
@@ -9,11 +9,13 @@ they're not being traced (useful for load balancers, etc.)
 
 ## Header name
 
-```Trace-Context```
+`Trace-Context`
 
 ## Field value
 
-```base16(<version>)-<version_format>```
+```
+base16(<version>)-<version_format>
+```
 
 The value will be US-ASCII encoded (which is UTF-8 compliant). Character `-` is
 used as a delimiter between fields.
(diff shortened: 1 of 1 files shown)


HTTP_HEADER_FORMAT.md
```diff
@@ -27,41 +26,41 @@ Is a 1-byte representing a 8-bit unsigned integer. Version 255 reserved.
...
-caller rather than strict rules to follow for 3 reasons:
+Controls tracing options such as sampling, trace level etc. It is a 1-byte representing a 8-bit 
+unsigned integer. The flags are recommendations given by the caller rather than strict rules to 
+follow for 3 reasons:
 
-1.  Trust and abuse.
-2.  Bug in caller
-3.  Different load between caller service and callee service might force callee
-    to down sample.
+1. Trust and abuse.
```
yurishkuro, line 58: It was never clear to me what this means. Maybe a bit of an explanation?

yurishkuro (approved):

HTTP_HEADER_FORMAT.md
```diff
@@ -27,41 +26,41 @@ Is a 1-byte representing a 8-bit unsigned integer. Version 255 reserved.
...
+2. Bug in caller
+3. Different load between caller service and callee service might force callee to down sample.    
 
-The behavior of other bits is currently undefined.
+##### Bits behavior definition (01234567):
+* The least significant bit (the 7th bit) provides recommendation whether the request should be 
+traced or not (```1``` recommends the request should be traced, ```0``` means the caller does not
```
codefromthecrypt: nit: curious why triple quoting things that aren't broken by newlines?

codefromthecrypt (approved):
looks good. this actually matches B3 better

HTTP_HEADER_FORMAT.md
codefromthecrypt: I believe this meant a bad actor can always say sampled for a ton of requests. If you don't trust the sampled bit, you can ignore it.

tylerbenson:
Why have trace options defined like this at all? I think the spec would be way better off by removing trace options from this header and pushing that functionality to a separate header that is more flexible.

codefromthecrypt:
My commentary is that the sampled bit being optional was a good thing as
fits existing use in B3, making it a pragmatic replacement. If we remove
the sampled bit, we couldnt use it without a yet to be defined replacement
header. If the proposal swings the way of removing the trace options bit
(again.. because we have discussed this for a year) I would suggest against
it.

codefromthecrypt:
Fwiw I opened up #8 to track the discussion around sampling priority (which while sounding similar is different than trace option 1). I could open another issue to collect the historical discusson around the flag, if helpful.

There is a lot of context in 1 bit and this single bit makes a relatively straightforward port of exisiting B3 code to this format. Adoptability of something that exists in dozens if not a hundred projects is fair game to retain imho. In our last workshop we realized people do need some incentive to change off B3 and making it easy is at least a lack of disincentive.

tylerbenson:
Ok, that's a fair argument.  Sorry I am pretty late to the game here.  Still trying to get my context fully spun up.

In that case, if the intent is for trace options to be extensible, we should clarify the process by which  additional flags are proposed and ratified.

mtwo:
> Still trying to get my **context** fully spun up

heh

bogdandrutu:
I will merge this initially because this moves us towards https://github.com/TraceContext/tracecontext-spec/pull/15

PR #18, opened 2017-09-23
Copies goal from http to top-level and copies rationale from google doc

codefromthecrypt:
This makes the goals and motivations more clear, so that future
revisions can understand what might be invalidated in the process of
change. The goal was copied from the http header definition, and the
why from the google doc created last year.

https://docs.google.com/document/d/1Mrw7hxVAkj7h98hvgRixDt1RrEJ5fsqqddPofIHRVvI/edit#
^^ shared with distributed-tracing google group

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # TraceContext Specs
 Specification for TraceContext propagation format.
 
+## Goal
+This specification defines formats to pass trace context information across systems for a HTTP request. Our goal is to share this with the community so that various tracing and diagnostics products can operate together, and so that services can pass context through them, even if they're not being traced (useful for load balancers, etc.).
+
 ## HTTP Format
 The HTTP format is defined [here](https://github.com/TraceContext/tracecontext-spec/blob/master/HTTP_HEADER_FORMAT.md)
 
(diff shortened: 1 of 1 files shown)


tsloughter (approved):

tsloughter (approved):

README.md
```diff
@@ -9,3 +12,10 @@ TODO: add link here
 
 ## Reference Implementations
 TODO: add link here
+
+## Why are we doing this?
+* If this becomes popular, frameworks and other services will automatically pass trace IDs through for correlated requests. This would prevent traces from hitting dead ends when a request reaches an un-instrumented service.
+* Once aligned on a header name, we can ask for a CORS exception from the W3C. This would allow browsers to attach trace IDs to requests and submit tracing data to a distributed tracing service.
+* Loggers can reliably parse trace / span IDs and include them in logs for correlation purposes.
+* Customers can use multiple tracing solutions (Zipkin + New Relic) at the same time and not have to worry about propagating two sets of context headers.
```
yurishkuro, line 20: The way I interpret this point is: trace-id is passed unchanged through apps using Zipkin and NewRelic.
  * corollary: both Zipkin and NewRelic must share the same trace-id
  * corollary: only tracing systems using the exact same format of trace-id can be compatible under this spec

codefromthecrypt:
>
>
>    - corollary: both Zipkin and NewRelic must share the same trace-id
>    - corollary: only tracing systems using the exact same format of
>    trace-id can be compatible under this spec
>
> similar, but not quite. Both systems must be able to "export and import"
this trace format to be compatible. For example, I had a chat with
DynaTrace and this is how they plan to interop with B3, for example
treating B3 as the compatibility header but preferring their own. In the
case theirs is lost, they can still continue the trace. There were similar
discussions in the past with microsoft, about using compat IDs at the edge
of the network, though sadly not documented here.

bogdandrutu (approved):

README.md
codefromthecrypt: merging this as it represents historical context of the spec. We will probably need to update this at some point when headers are finished definition, ex if we are able to meet this goal or if our header becomes secondary (backup header). We can't know yet if we can reach this goal basically, and for which vendors, but it is good to have a goal

Issue #16, opened 2017-09-22
Remove "hexadecimal" restriction


yurishkuro:
This has been discussed in the original PR (starting from https://github.com/TraceContext/tracecontext-spec/pull/1#issuecomment-313010383) but got lost/ignored. In the current form the spec excludes tracing systems that may be using differently formatted strings for trace and/or span IDs.

bogdandrutu:
Couple of things:
* I would say base16 not "hexadecimal". "hexadecimal" means you encode a number.
* You need to respect HTTP restrictions so you have to encode the "string" or "byte array" or however you call it. So we have to define what is that encoding because otherwise you put a restriction on the traceId to be HTTP compatible.

yurishkuro:
> because otherwise you put a restriction on the traceId to be HTTP compatible.

well, if we're talking about HTTP header spec, isn't that implied?

> I would say base16 not "hexadecimal". "hexadecimal" means you encode a number.

https://www.google.com/search?q=define+hexadecimal says "relating to or using a system of numerical notation that has 16 rather than 10 as its base."

The point is, the current spec exclude tracing systems that might choose a different encoding or length of the trace ID. Adrian had an argument about storage implication, which is valid in the sense that yes there are implications, but it's not impossible to resolve. In the best case the two systems use "compatible" trace-ids and can use them directly. In the worse case the receiving instrumentation will record the incoming trace-id as a correlation id.

codefromthecrypt:
Hi, Yuri.

What format does jaeger require that is not hexidecimal? What would happen
if it were not? How would you handle it in jaeger?

It is important we dont add technical debt in the process of doing this
work. Every constraint removed comes at a cost. It would be nice to know
you thought this through, especially as it can help others think this
through.

wu-sheng:
+1 about removing the base16 and length things.  Skywalking is using three longs for ID(trace segment id and trace id). So length changes for every generated id. If we follow the spec, we must cost more CPU and memory for the ID. This is not good.

Right now, skywalking can only consider to add a new HEAD for this spec if some people want this interop feature. Put two heads, one for skywalking, one for TraceContext Spec. It can work, but it is the best solution. :)

But still, I hope the spec can make the skywalking supporting easier.

wu-sheng:
@adriancole Right now, skywalking is considering put the ids(traceId and spanId) from outer tracer into tags of the EntrySpan.

codefromthecrypt:
>
[...]

PR #19, opened 2017-09-24
Variable-length IDs

bhs:
When trace-ids or span-ids are merely opaque strings (i.e., they're not used for arithmetic or substring manipulation, just for uniqueness), I don't see why this standard needs to specify their length. As such, I'm proposing that they become variable-length.

(Note that I would also welcome generic baggage support in this `tracecontext` format, but that's an orthogonal issue and thus not part of this PR)

Curious what other people think. (@bogdandrutu encouraged a pull request to discuss proposals like this)


codefromthecrypt:
I'm not sure how the number one could ever be a legitimate trace ID. Isn't this a bit too loose?

codefromthecrypt:
I guess it could be a legitimate trace ID once :P

bhs:
> I'm not sure how the number one could ever be a legitimate trace ID. Isn't this a bit too loose?

Yes, I included that just to emphasize the boundaries of the format... agreed that it's hard to imagine any uses outside of tests. The thinking here is just to be general when there's no good reason not to be. `/shrug`

codefromthecrypt:
"there's no good reason not to be" hmm I bet you can think of some reasons to have constraints, right? There must be some sort of impact

SergeyKanzhelev:
@bhs is the motivation here to support longer (or shorter) ids or make them an opaque strings? 

If first - for longer strings proposal should enable more scenarios. I've been talking to the team in MS that wants `span-id` be 128bit (basically a GUID) for their internal purposes. From an example I see on [aws](http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html) it looks they use longer span-id as well. 

Shorter strings may cause a problem as mistakenly sent GUID may be accepted as a valid string and the first part will be treated as a trace-id. Also uniqueness may become a problem as @adriancole mentioned.


If the proposal to support any string - I'd ask why hex vs base64 charset? Allowing trace-id be an opaque string may enable base64-encoded values. This may be beneficial if you want shorter string with the same 128bit uniqueness.


@bhs what is the scenario you trying to support?

SergeyKanzhelev:
oh, second qq already in discussion: https://github.com/TraceContext/tracecontext-spec/issues/16

bogdandrutu:
Couple of questions/comments here:
1) Who needs longer trace-ids?
2) What is the limit of the trace-ids (probably we should have a limited size)?
3) Have you tried to benchmark this encoding? Please consider to do so, I think it is important for products with high qps to have a cheap encoding/decoding (like envoy).

bhs:
@SergeyKanzhelev 

> is the motivation here to support longer (or shorter) ids or make them an opaque strings?

Good question. Much more the latter than the former, and I only kept the `base16` thing because I wanted to literally minimize the diff in this proposal. I do not have strong feelings about it (either way). If it was literally up to me (clearly not), I would vote for arbitrary string data, `base64`-encoded or otherwise.

@bogdandrutu 

> Who needs longer trace-ids?

... or shorter. I was actually thinking about 8-byte trace-ids, not 32-byte trace-ids when I proposed this. Though IIRC the AWS folks used 24-byte trace-ids? (Trying to remember what Abhishek said IRL in Berlin – I could be wrong for sure)

My thinking here is definitely more that the ids should literally be used for identification (and not much more) in v1 of this spec.

> What is the limit of the trace-ids (probably we should have a limited size)?

I'm happy to include sanity-checks on either (lower/upper) end if it makes this go through. Since we're dealing with HTTP headers, the upper bound is annoying inasmuch as it's inconsistent across implementations. What I would like to avoid is a requirement that all tracecontext-spec impls use precisely the same bitwidth for ids.

> Have you tried to benchmark this encoding? Please consider to do so, I think it is important for products with high qps to have a cheap encoding/decoding (like envoy).

I have not tried, though given that we're only dealing with this when there's also a bunch of I/O / syscalls, it's really hard to believe that parsing a variable-length string (with no additional syscalls) is  going to make a difference in a real-world benchmark.

If we can agree about the encoding stuff, I'm happy to create a test demonstrating the difference (and the RPS numbers needed to observe a performance impact, if any). If that's a "discussion requirement" for spec changes like this, let me know.

bogdandrutu:
@bhs 
> Good question. Much more the latter than the former, and I only kept the base16 thing because I wanted to literally minimize the diff in this proposal. I do not have strong feelings about it (either way). If it was literally up to me (clearly not), I would vote for arbitrary string data, base64-encoded or otherwise.

I thing this was discussed, and if I remember you kind of confirmed that this is better for production engineer (SREs). Happy to use base64 (a bit slower than base16 for small strings like this but probably not significantly). I would keep base16 for the moment and we can benchmark the encoding for 8/16/24 byte arrays and decide.

> I'm happy to include sanity-checks on either (lower/upper) end if it makes this go through. Since we're dealing with HTTP headers, the upper bound is annoying inasmuch as it's inconsistent across implementations. What I would like to avoid is a requirement that all tracecontext-spec impls use precisely the same bitwidth for ids.

Do you think a limited size of 128 byte arrays is not enough for each of the ids?

> I have not tried, though given that we're only dealing with this when there's also a bunch of I/O / syscalls, it's really hard to believe that parsing a variable-length string (with no additional syscalls) is going to make a difference in a real-world benchmark.

There are multiple possible options that we have here if we want to support variable sizes:
* Rely on a delimiter "-" to determine the size. 
* Include the size in the header base16(size)base16(trace_id) that may improve performance.
* Don't support variable sizes and pick a reasonable size 24 and always force 24 bytes. (against this PR, but I am adding as an option).

bhs:
@bogdandrutu

> > Good question. Much more the latter than the former, and I only kept the base16 thing because I wanted to literally minimize the diff in this proposal. I do not have strong feelings about it (either way). If it was literally up to me (clearly not), I would vote for arbitrary string data, base64-encoded or otherwise.

> I thing this was discussed, and if I remember you kind of confirmed that this is better for production engineer (SREs). Happy to use base64 (a bit slower than base16 for small strings like this but probably not significantly). I would keep base16 for the moment and we can benchmark the encoding for 8/16/24 byte arrays and decide.

I think I was arguing that there be a human-friendly delimiter (`-` as we have at `master` is 👍 on that front, and compatible / non-overlapping with both `base16` and `base64` encodings). I only bring up the base64 thing since it's a safe and non-fancy (if admittedly suboptimal) way of encoding arbitrary string data per @SergeyKanzhelev's suggestion.

> > I'm happy to include sanity-checks on either (lower/upper) end if it makes this go through. Since we're dealing with HTTP headers, the upper bound is annoying inasmuch as it's inconsistent across implementations. What I would like to avoid is a requirement that all tracecontext-spec impls use precisely the same bitwidth for ids.

> Do you think a limited size of 128 byte arrays is not enough for each of the ids?

I think 128 **bytes** is more than enough if you want to set that as a max... or, at least, I have little sympathy for anyone using a 129-byte identifier. :)

> > I have not tried, though given that we're only dealing with this when there's also a bunch of I/O / syscalls, it's really hard to believe that parsing a variable-length string (with no additional syscalls) is going to make a difference in a real-world benchmark.

> There are multiple possible options that we have here if we want to support variable sizes:
> - Rely on a delimiter "-" to determine the size.
> - Include the size in the header base16(size)base16(trace_id) that may improve performance.
> - Don't support variable sizes and pick a reasonable size 24 and always force 24 bytes. (against this PR, but I am adding as an option).

I don't feel super strongly about this, but I am running under the *assumption* that none of this matters at all in terms of real-world performance, and thus we should choose the thing that's easiest for human beings. If we can agree about the rest of the changes here I'm again glad to write some simple benchmarks to measure overhead relative to an HTTP request.

cce:
:+1: for longer trace IDs, we use 20-byte trace IDs in TraceView.

mtwo:
@cce are you embedding metadata into the ID, or is it just super long?

SergeyKanzhelev:
What's the status of this?

wu-sheng:
> What's the status of this?

@SergeyKanzhelev I am not sure too. Skywalking IDs are Variable-length.

bhs:
@SergeyKanzhelev I am waiting for someone to either LGTM or tell me what needs to change :)

It seems like this would be stalled behind many of the open issues in this repo... many stimulating discussions, but lots of questions in the air about various things.

SergeyKanzhelev (approved):
LGTM

wu-sheng:
@bhs Variable-length is good for skywalking, but we still need an extra valid char: `.` . After all, id is very different for different implementations.

SergeyKanzhelev:
@wu-sheng what is this separator for? In some MS teams we are using hierarchical identifiers. The idea was to keep `span-id` as it is and attach hierarchy as a tail section of the `trace-context`.

bhs:
@wu-sheng @SergeyKanzhelev I am fine with substantially broadening the set of allowable characters (I might even be enthusiastic)... we could also reuse (or, ugh, define) some sort of simple escaping mechanism to effectively represent any ascii char. Curious about thoughts on this front.

yurishkuro:
As I mentioned in https://github.com/TraceContext/tracecontext-spec/issues/16#issuecomment-332070303, I do not believe that the allowed alphabet is particularly important, what's important is that we simply _cannot assign any semantic meaning to the blob_ that represents a trace ID (whether it's hexadecimal or richer alphabet), since the receiving tracer has no guarantee where that blob has come from - it could've come from a tracer that generated the blob completely randomly but the resulting representation just happened to look like it's a "hierarchical ID" or whichever semantics the receiving tracer might want to infer.

mtwo:
I referenced this in a proposed plan for v0.1.0 in #31 as I _think_ this is the only big issue blocking us from a first release. Please let me know on that issue if you think this is a good or bad process.

Regarding fixed vs. variable-length IDs, I don't have a strong opinion, and from reading the replies above it looks like most people are leaning towards variable length (?). I'm curious to hear what @jkowall, @abhiksingh, and @CodingFabian think.

CodingFabian:
I am all for treating it as opaque. It would allow to preserve the value without meaning. But more importantly it helps us agreeing on it. In the and its a (hopefully) unique blob of bits. 
On the topic of encoding, I would like to point out that we transmit our ids as numeric values on many non http formats for efficiency. Its interesting to see how hard it is to use a 64bit long for that. 
We probably would stick to 8+8byte. 

Now the big question for me is: what to do with the trace ids we are discussing here. When they are vendor specific, we cannot preserve them. So at maximum we could use the as reference to other segments of a trace. 
Declaring a specific header as opaque would solve only one issue: mangling it. 

If we would specify an encoding and value range, that would help processing the id, but still neither the tracer, nor the end user would know where to retrieve the spans from

codefromthecrypt:
We should know the tradeoff when we do something and opening the range..
* seems to support some longer ids (ex Solarwinds, SkyWalking and MS hierarchical form)
* interferes with interop of tracing systems that can currently interop "without looking" (dapper-based systems with 64 or 128-bit trace IDs).
* doesn't alone enable multi-part IDs, ones that have IDs coupling across multiple fields, ex. AWS (timestamp+random although you can concat these)

In so far as it loosens the data that can be passed around, it doesn't solve the problems it creates. For example, before dapper-like systems could use the format directly. Now, we are saying anything goes, which means sequential or path-based data gets crammed into the field with no semantic clarification of what to do.. and with not enough data to say what to do next (next being when to devise another id).

Long story short, we are removing the ability for dapper-based systems to interop without prior knowledge (those systems can and do interop today, without the change proposed here). We also don't mention how practically these "open ids" can be shared. My guess is that the goal is to pass data and we'll just figure out later what to do with it?

codefromthecrypt:
For the moment, I assume this pull request is only focused on size and that any movement if any from base16 will happen on #16 as opened by @yurishkuro a while back.

Here are some concrete suggestions for text to add, should we decide to go with no minimum or maximum length of trace ID. If folks like @bogdandrutu @CodingFabian @jkowall @abhiksingh @christian-schwarzbauer are ok with this, I am. Please bear in mind, many are busy and scheduled time to be on-site Nov 1,2 to discuss this.

Anyway.. here are some notes I think should go into the spec or some usage related section if we remove all length constraints.

* Some systems may send padded data and others not. It is not guaranteed the case that the trace ID `00aabcd` is the same as `aabcd`.
* This specification allows unbounded trace IDs. For example, some systems may not be able to store a 4KiB identifier. The next hop might restart a trace they cannot store.

The latter point is possible even with the existing system. We don't have an issue to track how to handle an incoming ID that you cannot process. Most systems restart traces if they cannot continue a trace for any reason. However, we might add an issue to explore other options, such as moving the "original ID" to a correlation header.

CodingFabian:
unless all vendors will agree on an ID format (which I consider unlikely) the interop issue always remains.
if the goal of the spec is to enable interop (e.g. bring together a complete trace from two sources) then more work needs to be done (e.g. explaining from which api the trace can be retrieved given the incoming headers)
Just preserving headers a tracer cannot understand is neither feasible nor helpful.
the likely outcome is that in the current proposal interop is not possible: Tracers will see an incoming Id which doesn't match the format of the tracer. Because it cannot be handled in data-structures of the tracer it will be dropped.

Variable length specifies this as accepted solution. Which could be good to get the spec started off, but I doubt will be actually useful.

SergeyKanzhelev:
During workshop everybody was OK with the fixed length. We can revisit it later again.

wu-sheng:
@SergeyKanzhelev I will still need to argue about our variable length trace id and parent span id,  and non base64 ID.
flying to beijing, today. keep in touch.

wu-sheng:
I will start  a new issue .

Issue #20, opened 2017-09-25
Companion header for name-values pairs


SergeyKanzhelev:
Let's define the tags header that will contain name-value pairs that suppose to be propagated by tracing systems.

1. Call it `Correlation-Context`
2. Should be human readable
3. Make it appende-able (comma-separated) https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
4. If we need to specify the type of the value (int, bool) we can use the same approach as cookies - `name=value;type=int`

yurishkuro:
should there be some encoding if name/value are composed of unsafe characters for HTTP header value (or contain `= , ;`) ?

also, comma-separated is equivalent to repeating the header multiple times iirc, but does it put an implicit restriction on the length? I.e. if I have two long values, putting them into separate (identically named) headers will pass the proxies, but if one of them decides to reformat as comma-separated list, then it might exceed the max header length

SergeyKanzhelev:
making header append-able makes it easy to add another property without parsing existing header. For the proxy that appended headers - I think it may be just a limitation on size for now.

For encoding - maybe support urlencode?

bogdandrutu:
Couple of high level comments about this:
* What are the requirement/high level specs:
    * Do we want to have supported types for values?
    * Do we want to have limits on sizes?
    * Any restriction on the key/values characters?
* What is the encoding we use here?
    * base16/base64/other?

Let's define a bit what we try to achieve. It will be great if for example we decide on few types that we support, a set of size limits, etc. but I am fine if we cannot get an agreement.

SergeyKanzhelev:
Since we have general agreement - let's continue discussion in the PR https://github.com/TraceContext/tracecontext-spec/pull/21

bogdandrutu:
Closes as duplicate of this https://github.com/TraceContext/tracecontext-spec/issues/13?

Issue #13, opened 2017-09-16
Standardize format for "simple" baggages (a.k.a "tags")


bogdandrutu:
**Background:**
* OpenTracing call them "baggages" - they are a simpler version of [this paper] (http://pivottracing.io/mace15pivot.pdf)
* OpenCensus call them "tags"

**Open Questions:**
* 1) Use a different header than "Trace-Context" or extend that format?
* 2) What types do we support: key -> string? value -> (string, int64, bool) ?
* 3) Size restrictions: Per key or value? Per entire set of baggages/tags? None?

yurishkuro:
* +1 for different header(s). I actually liked, conceptually, the idea in Microsoft spec that only unique request ID (aka span ID) needs a dedicated header, all other values like traceID and sampling flags can be just part of the baggage. But it makes the basic tracing case more verbose.
* what types to support - given that this is a spec for a text protocol, I don't think the discussion of types is relevant, because it only manifests at the API level for accessing the values. From this spec pov, baggage keys and values are just strings.
* I would leave size restrictions out of the spec, at least in the first iterations - the implementations can apply their own rules to that.

Other open questions:
  * do we combine all baggage items into one header or split them?
    * Microsoft spec combines them iirc, which implicitly puts extra restrictions on the size
    * in Jaeger we split each baggage item into individual header, but I don't quite like the encoding scheme we're using where the baggage key is part of the header name, which puts restrictions on the character set available for the keys
    * one possible encoding is:
      * `Baggage-Size: <count of items>`
      * `Baggage0: <key>=<value>`
      * ...
      * `Baggage<n-1>: <key>=<value>`
  * are the implementations allowed to do drop the baggage, e.g. if it gets too large?
    * can all if it be dropped? (imo yes)
    * can some of it be dropped? (imo yes)
  * what is the encoding of the keys and values?
    * e.g. `\n` is not legal in the header value, `=` above is a reserved char, `,` is a reserved char in HTTP spec, so _some_ encoding is certainly required
  * we might need to define some approach to name spacing of baggage keys, to avoid different implementations overriding each other's metadata

yurishkuro:
btw, I am obviously biased towards "baggage" because of OpenTracing, but I think "tags" is way too generic and ambiguous term, while "baggage" is unusual enough to not be confused with anything else.

CodingFabian:
[...]

PR #21, opened 2017-09-25
first version of correlation context header

SergeyKanzhelev:
For the #20

### DIFF
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -1,6 +1,6 @@
 # Correlation Context HTTP Header Format
 
-A correlation context header is used to pass the name-value context properties (also known as tags) for the trace. This is a companion header for the `Trace-Context`. The values are intended for control of logging systems and should be passed along to any child requests. 
+A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Context`. The values should be passed along to any child requests. Note that uniqueness of the key within the `Correlation-Context` is not guaranteed. Context received from upstream service may be altered before passing it along.
 
 # Format
 
@@ -14,13 +14,22 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 `name1=value1[;type=(number|bool|string)], name2=value2[;type=(number|bool|string)]`
 
-Overall Correlation-Context length MUST NOT exceed `1024` bytes. Key and value length should stay well under the combined limit of `1024` bytes. 
+**Limits:**
+1. Maximum number of name-value pairs: `180`.
+2. Maximum number of bytes per a single name-value pair: `4096`.
+3. Maximum total length of all name-value pairs: `8192`.
+
+## Name format
+
+Url encoded string. Spaces should be trimmed from beginning and the end of the name. Names are case sensitive.
 
 ## Supported value types
 
+All spaces should be trimmed from the beginning and the end of the value.
+
 ### String
 
-Default value type. Represents a url encoded string value
+Default value type. Represents a url encoded string value. Value is case sensitive.
 
 **Examples**:
 
@@ -32,18 +41,18 @@ component=Front%3Dend
 
 ### Boolean
 
-Binary flag. Supported values `true` and `false`
+Binary flag. Supported values `1` for true and `0` for false.
 
 **Examples**:
 
 ```
-IsAuthenticated=true;type=bool
-IsAuthenticated=false;type=bool
+IsAuthenticated=1;type=bool
+IsAuthenticated=0;type=bool
 ```
 
 ### Number
 
-Numeric value in invariant culture (`.` as a decimal separator).
+Numeric value as described in IEEE 754-2008 binary64 (double precision) numbers [IEEE754](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).
 
 **Examples**:
 
@@ -60,5 +69,5 @@ Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true;
 
 ```
 Correlation-Context: component=Frontend
-Correlation-Context: flightName=DF:28, ExposurePercentage=33.33;type=number
+Correlation-Context: flight%3DName=DF:28, ExposurePercentage=33.33;type=number
 ```
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -10,9 +10,9 @@ A correlation context header is used to pass the name-value context properties f
 
 Multiple correlation context headers are allowed. Values can be combined in a single header according to the [rfc](https://tools.ietf.org/html/rfc7230#page-24).
 
-## Field value
+## Header value
 
-`name1=value1[;type=(number|bool|string)], name2=value2[;type=(number|bool|string)]`
+`name1=value1,name2=value2`
 
 **Limits:**
 1. Maximum number of name-value pairs: `180`.
@@ -25,51 +25,26 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 Url encoded string. Spaces should be trimmed from beginning and the end of the name. Names are case sensitive.
 
-## Supported value types
+## Value format
 
 All spaces should be trimmed from the beginning and the end of the value.
 
-### String
-
-Default value type. Represents a url encoded string value. Value is case sensitive.
+Value represents a url encoded string. Value is case sensitive.
 
 **Examples**:
 
 ```
 component=Frontend
-component=Frontend;type=string
 component=Front%3Dend
-```
-
-### Boolean
-
-Binary flag. Supported values `1` for true and `0` for false.
-
-**Examples**:
-
-```
-IsAuthenticated=1;type=bool
-IsAuthenticated=0;type=bool
-```
-
-### Number
-
-Numeric value as described in IEEE 754-2008 binary64 (double precision) numbers [IEEE754](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).
-
-**Examples**:
-
-```
-ExposurePercentage=33.33;type=number
-Step=10;type=number
 ```
 
 # Examples of HTTP headers
 
 ```
-Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true;type=bool
+Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true
 ```
 
 ```
 Correlation-Context: component=Frontend
-Correlation-Context: flight%3DName=DF:28, ExposurePercentage=33.33;type=number
+Correlation-Context: flight%3DName=DF:28, ExposurePercentage=33.33
 ```
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -8,7 +8,7 @@ A correlation context header is used to pass the name-value context properties f
 
 `Correlation-Context`
 
-Multiple correlation context headers are allowed. Values can be combined in a single header according to [rfc](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).
+Multiple correlation context headers are allowed. Values can be combined in a single header according to the [rfc](https://tools.ietf.org/html/rfc7230#page-24).
 
 ## Field value
 
@@ -18,6 +18,8 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 1. Maximum number of name-value pairs: `180`.
 2. Maximum number of bytes per a single name-value pair: `4096`.
 3. Maximum total length of all name-value pairs: `8192`.
+
+*See [CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md#Limits] for rationale of these limits.*
 
 ## Name format
 
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -1,0 +1,64 @@
+# Correlation Context HTTP Header Format
+
+A correlation context header is used to pass the name-value context properties (also known as tags) for the trace. This is a companion header for the `Trace-Context`. The values are intended for control of logging systems and should be passed along to any child requests. 
+
+# Format
+
+## Header name
+
+`Correlation-Context`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to [rfc](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).
+
+## Field value
+
+`name1=value1[;type=(number|bool|string)], name2=value2[;type=(number|bool|string)]`
+
+Overall Correlation-Context length MUST NOT exceed `1024` bytes. Key and value length should stay well under the combined limit of `1024` bytes. 
+
+## Supported value types
+
+### String
+
+Default value type. Represents a url encoded string value
+
+**Examples**:
+
+```
+component=Frontend
+component=Frontend;type=string
+component=Front%3Dend
+```
+
+### Boolean
+
+Binary flag. Supported values `true` and `false`
+
+**Examples**:
+
+```
+IsAuthenticated=true;type=bool
+IsAuthenticated=false;type=bool
+```
+
+### Number
+
+Numeric value in invariant culture (`.` as a decimal separator).
+
+**Examples**:
+
+```
+ExposurePercentage=33.33;type=number
+Step=10;type=number
+```
+
+# Examples of HTTP headers
+
+```
+Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true;type=bool
+```
+
+```
+Correlation-Context: component=Frontend
+Correlation-Context: flightName=DF:28, ExposurePercentage=33.33;type=number
+```
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -2,13 +2,15 @@
 
 A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Context`. The values should be passed along to any child requests. Note that uniqueness of the key within the `Correlation-Context` is not guaranteed. Context received from upstream service may be altered before passing it along.
 
+*See [rationale document](CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md) for details of decisions made for this format.*
+
 # Format
 
 ## Header name
 
 `Correlation-Context`
 
-Multiple correlation context headers are allowed. Values can be combined in a single header according to the [rfc](https://tools.ietf.org/html/rfc7230#page-24).
+Multiple correlation context headers are allowed. Values can be combined in a single header according to the [RFC 7230](https://tools.ietf.org/html/rfc7230#page-24).
 
 ## Header value
 
@@ -19,8 +21,6 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 2. Maximum number of bytes per a single name-value pair: `4096`.
 3. Maximum total length of all name-value pairs: `8192`.
 
-*See [CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md#Limits] for rationale of these limits.*
-
 ## Name format
 
 Url encoded string. Spaces should be trimmed from beginning and the end of the name. Names are case sensitive.
@@ -29,12 +29,9 @@ Url encoded string. Spaces should be trimmed from beginning and the end of the n
 
 All spaces should be trimmed from the beginning and the end of the value. Value ends with the special character `;`, separator `,` or end of string. Value represents a url encoded string and case sensitive. 
 
-**Examples**:
+## Properties
 
-```
-component=Frontend
-component=Front%3Dend
-```
+Properties are expected to be in a format of keys & key-value pairs `;` delimited list `;k1=v1;k2;k3=v3`. All unknown property names and name-value pairs should be skipped.
 
 # Examples of HTTP headers
 
@@ -44,5 +41,5 @@ Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true
 
 ```
 Correlation-Context: component=Frontend
-Correlation-Context: flight%3DName=DF:28, ExposurePercentage=33.33
+Correlation-Context: flight%3DName=DF28, ExposurePercentage=33.33
 ```
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -35,11 +35,22 @@ Properties are expected to be in a format of keys & key-value pairs `;` delimite
 
 # Examples of HTTP headers
 
+Single header: 
+
 ```
-Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true
+Correlation-Context: component=Frontend,flightName=DF:28,IsAuthenticated=true
 ```
+
+Context might be split into multiple headers:
 
 ```
 Correlation-Context: component=Frontend
-Correlation-Context: flight%3DName=DF28, ExposurePercentage=33.33
+Correlation-Context: flight%3DName=DF28,ExposurePercentage=33.33
+```
+
+Values and names might begin and end with spaces:
+
+```
+Correlation-Context: component =   Frontend
+Correlation-Context: flight%3DName = DF28, ExposurePercentage = 33.33
 ```
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
@@ -12,7 +12,7 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 ## Header value
 
-`name1=value1,name2=value2`
+`name1=value1[;properties1],name2=value2[;properties2]`
 
 **Limits:**
 1. Maximum number of name-value pairs: `180`.
@@ -27,9 +27,7 @@ Url encoded string. Spaces should be trimmed from beginning and the end of the n
 
 ## Value format
 
-All spaces should be trimmed from the beginning and the end of the value.
-
-Value represents a url encoded string. Value is case sensitive.
+All spaces should be trimmed from the beginning and the end of the value. Value ends with the special character `;`, separator `,` or end of string. Value represents a url encoded string and case sensitive. 
 
 **Examples**:
 
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
@@ -16,6 +16,12 @@ Correlation context header is used to propagate properties not defined in `Trace
 
 Header should be human readable and editable. Thus spaces are allowed before and after the comma separator.
 
+## Case sensitivity of names
+
+There are few considerations why the names should be case sensitive:
+- some keys may be a url query string parameters which are case sensitive
+- forcing lower case will decrease readability of the names written in camel case
+
 ## Strings encoding
 
 Url encoding is low-overhead way to encode unicode characters for non-latin characters in the values. Url encoding keeps a single words in latin unchanged and easy readable.
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
@@ -10,7 +10,6 @@ Correlation context header is used to propagate properties not defined in `Trace
 
 - Should be human readable. Cryptic header will hide the fact of potential information disclosure.
 - Should be appende-able (comma-separated) https://tools.ietf.org/html/rfc7230#page-24 so nodes can add context properties without parsing an existing headers.
-- Some systems specify type of value. Most commonly the same set as JSON allows - boolean and numeric. Should allow to express this.
 - It is expected that the typical name will be a single word in latin and the value will be a short string in latin or a derivative of an url.
 
 ## Trimming of spaces
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
@@ -12,6 +12,23 @@ Correlation context header is used to propagate properties not defined in `Trace
 - Should be appende-able (comma-separated) https://tools.ietf.org/html/rfc7230#page-24 so nodes can add context properties without parsing an existing headers.
 - It is expected that the typical name will be a single word in latin and the value will be a short string in latin or a derivative of an url.
 
+## Why a single header?
+
+Another option is to use prefixed headers, such as `trace-context-X` where `X` is a propagated field name. This could reduce the size of the data, particularly in http/2 where header compression can apply.
+
+Generally speaking `Correlation-Context` header may be split into multiple headers and compression may be at the same ballpark as repeating values will be converted into a single value in HPAC's dynamic collection. That's said there were no profiling made to make this decision.
+
+The approach with multiple headers has the following problems:
+- Name values limitation is much more pressing when the context name is used a part of a header name.
+- The comma-separated format similar to the proposed still needs to be supported in every individual header. This makes parsing harder.
+- Single header is easier to configure for tracing by many app servers.
+
+## Why not Vary-style?
+
+The [Vary](https://tools.ietf.org/html/rfc7231#section-7.1.4) approach is another alternative, which could be used to accomplish the same. For example, `Correlation-Context: x-b3-parentid;ttl=1` could tell the propagation to look at and forward the parent ID header, but only to the next hop. This has an advantage of http header compression (hpack) and also weave-in with legacy tracing headers.
+
+Vary approach may be implemented as a new "header reference" value type `ref`. `Correlation-Context: x-b3-parentid;type=ref;ttl=1` if proven needed.
+
 ## Trimming of spaces
 
 Header should be human readable and editable. Thus spaces are allowed before and after the comma separator.
--- a/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
+++ b/CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
@@ -1,0 +1,33 @@
+# Rationale for Correlation Context HTTP Header Format
+
+This document provides rationale for the decisions made for the `Correlation-Context` header format.
+
+## Purpose of a header
+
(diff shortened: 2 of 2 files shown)


CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,64 @@
+# Correlation Context HTTP Header Format
+
+A correlation context header is used to pass the name-value context properties (also known as tags) for the trace. This is a companion header for the `Trace-Context`. The values are intended for control of logging systems and should be passed along to any child requests.
```
yurishkuro: Rm "also known as tags"

yurishkuro: Rm reference to logging systems, it's unnecessary restriction

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,64 @@
...
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to [rfc](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).
+
+## Field value
+
+`name1=value1[;type=(number|bool|string)], name2=value2[;type=(number|bool|string)]`
+
+Overall Correlation-Context length MUST NOT exceed `1024` bytes. Key and value length should stay well under the combined limit of `1024` bytes. 
+
+## Supported value types
```
yurishkuro: Missing description of keys, such as 
- case sensitive
- urlencoded?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,64 @@
...
+
+`Correlation-Context`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to [rfc](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).
+
+## Field value
+
+`name1=value1[;type=(number|bool|string)], name2=value2[;type=(number|bool|string)]`
+
+Overall Correlation-Context length MUST NOT exceed `1024` bytes. Key and value length should stay well under the combined limit of `1024` bytes.
```
yurishkuro: Does the combined limit apply to a single header or across all correlation headers combined? I'd prefer just one.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+component=Frontend
+component=Frontend;type=string
+component=Front%3Dend
+```
+
+### Boolean
+
+Binary flag. Supported values `true` and `false`
```
bogdandrutu: If we talk about save space, easy to save here by using 1 character to represent this.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,64 @@
...
+
+**Examples**:
+
+```
+IsAuthenticated=true;type=bool
+IsAuthenticated=false;type=bool
+```
+
+### Number
```
bogdandrutu: Do we need to support doubles or int64 are enough? If not can you give me some examples where?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,64 @@
+# Correlation Context HTTP Header Format
+
+A correlation context header is used to pass the name-value context properties (also known as tags) for the trace. This is a companion header for the `Trace-Context`. The values are intended for control of logging systems and should be passed along to any child requests. 
+
+# Format
+
+## Header name
+
+`Correlation-Context`
```
bogdandrutu, line 11: I would argue a bit more about the name, I like the simplicity of the format.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
yurishkuro: Should it say something about whether nodes are allowed to drop these headers? If they are, should there be some recommendation which ones to drop, e.g. with longest value first?

yurishkuro (commented):
Overall lgtm

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
bogdandrutu: If it is just one can we have a lower limit for one? Let's say something like:
* keys max size X bytes
* string values max size X bytes
* numbers have to be also size limited if we use the current format definition.
* Maybe a total max headers?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Do you want to argues as part of this PR or as part of #13?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: like it

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: it's JSON-like number. May be both. Do you want to split into int and float?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: fixed

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: proposed new limits. The idea of proposed limits: 
- url should fit easily as a value
- overall size should not be driven by max number multiply by max size as it will become too big
- max number of properties matches max number of cookies typically supported by browser for a single domain (generally browsers supports less, just took Chrome's number for a reference)

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: added

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
...
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to [rfc](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).
+
+## Field value
+
+`name1=value1[;type=(number|bool|string)], name2=value2[;type=(number|bool|string)]`
+
+**Limits:**
+1. Maximum number of name-value pairs: `180`.
```
yurishkuro: wait, why not 42? :-)

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
...
+2. Maximum number of bytes per a single name-value pair: `4096`.
```
yurishkuro: what is this based on? Maybe include a reference

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+ExposurePercentage=33.33;type=number
+Step=10;type=number
+```
+
+# Examples of HTTP headers
+
+```
+Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true;type=bool
```
codefromthecrypt: seems there's an intentional space after the commas here? if intended, maybe add a delimiter section

codefromthecrypt (commented):
knowing thrashing has happened in the past, I think a sidecar rational doc should be present. ex correlation-format_rationale.md

In this rationale, we should mention tradeoffs, for example, this will subvert http/2 (hpack) compression almost always. Presumably we are doing this as we don't expect folks to pass a header pattern or we prefer not to (in order to pre-ordain a fixed count of headers to make code easier ex map-based access). Presumably, URL-encoding is to ensure there's no interference with comma or equals? An alternative is form encoding, for example. Just to vocalize (well capture in text) so that others don't have to answer this in long form every time this format is questioned.

I agree w/ @yurishkuro comments too, for example, justification of constraints should show here, too.

Please don't view the above as an argument against this, I just want rationale because I usually end up having to explain things that aren't documented. We can do it up front and save us all time later.

For example, this format might change between now and november as folks who aren't online today become online.. I don't want future issues to become 50 comments of rehashing.. make sense?

codefromthecrypt (approved):
on a trust basis that rationale ends up here :)

cc @basvanbeek I think you mentioned something like this in the past

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
+# Correlation Context HTTP Header Format
+
+A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Context`. The values should be passed along to any child requests. Note that uniqueness of the key within the `Correlation-Context` is not guaranteed. Context received from upstream service may be altered before passing it along.
```
codefromthecrypt, line 3: use cases can go here, or at least one example. For example, one can pass vendor-specific data here which helps resolve a trace identifier against an existing format

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
...
+
+### String
+
+Default value type. Represents a url encoded string value. Value is case sensitive.
+
+**Examples**:
+
+```
+component=Frontend
+component=Frontend;type=string
```
codefromthecrypt: this will explode into more bytes a lot. I like that type defaults to string to reduce that. unless you are expecting to add more attributes, possibly better a keyword approach ex priority=0.1;number

codefromthecrypt: or even cost=42;n, worthless=0;b

yurishkuro: or support strings only. HTTP is a text format so the value is going to be read as a string anyway, and the application that's interested in interpreting a particular baggage key should already know which type it should be.

codefromthecrypt:
Starting (possibly ending) just strings works for me. Easier to start and
starting is important.

I know census tags are typed, so maybe folks will have opinions about this,
but could well still circle around. Will defer to others on this. Could be
fair play as another issue anyway.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: lost in outdated comments...

proposed new limits. The idea of proposed limits:

url should fit easily as a value
overall size should not be driven by max number multiply by max size as it will become too big
max number of properties matches max number of cookies typically supported by browser for a single domain (generally browsers supports less, just took Chrome's number for a reference)

I'll add correlation-format_rationale.md as @adriancole suggests

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: it was intentional to demonstrate that the name and value should be trimmed

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
...
+
+A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Context`. The values should be passed along to any child requests. Note that uniqueness of the key within the `Correlation-Context` is not guaranteed. Context received from upstream service may be altered before passing it along.
+
+# Format
+
+## Header name
+
+`Correlation-Context`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to [rfc](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).
```
codefromthecrypt: ps rfc2616 is dead (ty @mnot) https://www.mnot.net/blog/2014/06/07/rfc2616_is_dead

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,73 @@
+# Correlation Context HTTP Header Format
```
codefromthecrypt, line 1: fyi @louiscryan this is that thing which can carry extra stuff (separate from the trace-context)

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
wu-sheng: Prefer String only.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Thanks, will change the link

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Thanks, will change the link

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
wu-sheng: The values are asked to do URL encoding, I think the space is unnecessary.

wu-sheng (commented):
- The length limit is strange. Why choose those values?
- Hope only support String for easier usage.
- Space should not be used.

In general, I like this Correlation Context. @SergeyKanzhelev :+1:

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Next thing I was thinking people may ask is the ttl. For context properties like "component name". Another it PII (personal identifiable information). PII flag may help to identify when to pass it along and when not to. Like pass it to the dependent services with the this host name, but not to others.

Alternative to `;type=` may be some sort of prefix for a value. The only problem with prefix as a custom escape scheme and ultimately a conflict with the human readability requirement.

SergeyKanzhelev: I'm open to ideas and ok to strip out the `type`

SergeyKanzhelev:
@adriancole for the types of values - it is generally for census. Also some internal teams in MS also uses typed properties collections. I can strip out types if there would be any controversy.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
wu-sheng: Does any tracer process other tracer's `Correlation-Context` HEAD? Do we have any use cases? Or are we going to define some common keys? @SergeyKanzhelev @adriancole

codefromthecrypt:
>
> Next thing I was thinking people may ask is the ttl. For context
> properties like "component name". Another it PII (personal identifiable
> information). PII flag may help to identify when to pass it along and when
> not to. Like pass it to the dependent services with the this host name, but
> not to others.
>
TTL is a very good one, also for vendor-scoped data (ex if ttl=1 the next
hop is not same vendor, can toss it, else refresh)

maybe we reserve the character ';' ?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
SergeyKanzhelev: @wu-sheng I was thinking defining the well-known names later as a separate discussion. Same way open tracing defines specific tags

wu-sheng: @SergeyKanzhelev Let's discuss that later. In my experiences, the contents in `Correlation-Context` are really much implementation related. Hard to say common keys are useful or not.

SergeyKanzhelev:
I'll create separate issues for:
- `type` - type of the value
- `ttl` - number of hops to live
- `pii` - level of access to data
- discussion for the max total length limit
- discussion for well-known context names catalog

once this PR merged. Unless there will be more critical comments

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+# Examples of HTTP headers
+
+```
+Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true
+```
+
+```
+Correlation-Context: component=Frontend
```
codefromthecrypt: good example. often ppl forget about headers being a multimap

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,32 @@
...
+
+## General considerations
+
+- Should be human readable. Cryptic header will hide the fact of potential information disclosure.
+- Should be appende-able (comma-separated) https://tools.ietf.org/html/rfc7230#page-24 so nodes can add context properties without parsing an existing headers.
+- It is expected that the typical name will be a single word in latin and the value will be a short string in latin or a derivative of an url.
+
+## Trimming of spaces
+
+Header should be human readable and editable. Thus spaces are allowed before and after the comma separator.
```
codefromthecrypt: nit: isn't a space before a comma a bit strange? should we suggest that?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,32 @@
...
+## Limits
+
+The idea behind limits is to provide trace vendors common safeguards so the content of the `Correlation-Context` header can be stored with the request. Thus the limits are defined on the number of keys, max pair length and the total size. The last limit is the most important in many scenarios as it allows to plan for the data storage limits.
+
+Another consideration was that cookies may be an alternative way to store and pass the correlation context fields. So the limits should make context name-value pairs fit the typical cookie limit.
+
+- *Maximum number of name-value pairs* - this limit was taken as a number of cookies allowed by Chrome.
+- *Maximum number of bytes per a single name-value pair* - the limit allows to store URL as a value with some extra details as a single context name-value pair. It is also a typical cookie size limitation.
+- *Maximum total length of all name-value pairs* - TODO: LOOKING FOR SUGGESTIONS HERE
+
```
codefromthecrypt, line 55: tradeoffs:

Another option is to use prefixed headers, such as trace-context-X where X is a propagated field. This could reduce the size of the data, particularly in http/2 where header compression can apply. However XXyour_reason_hereXX

codefromthecrypt (commented):
small notes. lookin good

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
codefromthecrypt: another option to add to rational section is http vary style

## Why not Vary-style?
The [Vary](https://tools.ietf.org/html/rfc7231#section-7.1.4) approach is another alternative, which could be used to accomplish the same. For example, `Correlation-Context: x-b3-parentid;ttl=1` could tell the propagation to look at and forward the parent ID header, but only to the next hop. This has an advantage of http header compression (hpack) and also weave-in with legacy tracing headers. However **YOUR_REASON_HERE**

bogdandrutu (approved):
I would like to start also an issue to revisit the sizes. Good work.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+**Limits:**
+1. Maximum number of name-value pairs: `180`.
+2. Maximum number of bytes per a single name-value pair: `4096`.
+3. Maximum total length of all name-value pairs: `8192`.
+
+*See [CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md#Limits] for rationale of these limits.*
+
+## Name format
+
+Url encoded string. Spaces should be trimmed from beginning and the end of the name. Names are case sensitive.
```
jcchavezs: I have a concern about the `case sensitive` as some frameworks out there `normalize` headers making them lowercase. What about require a lowercase so we avoid these situations?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,32 @@
...
+## Limits
+
+The idea behind limits is to provide trace vendors common safeguards so the content of the `Correlation-Context` header can be stored with the request. Thus the limits are defined on the number of keys, max pair length and the total size. The last limit is the most important in many scenarios as it allows to plan for the data storage limits.
+
+Another consideration was that cookies may be an alternative way to store and pass the correlation context fields. So the limits should make context name-value pairs fit the typical cookie limit.
+
+- *Maximum number of name-value pairs* - this limit was taken as a number of cookies allowed by Chrome.
+- *Maximum number of bytes per a single name-value pair* - the limit allows to store URL as a value with some extra details as a single context name-value pair. It is also a typical cookie size limitation.
+- *Maximum total length of all name-value pairs* - TODO: LOOKING FOR SUGGESTIONS HERE
```
jcchavezs: Is the sqs queue max message size a number we want to consider for the decision here?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
bogdandrutu: I don't think that is the case @jcchavezs. Headers name may be normalized but I am pretty sure the values cannot be normalized as lowercase because that will completely change the value in case of base64 encoded strings for example.

jcchavezs: Maybe I am missing something but aren't this going to be non-encoded names?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
bogdandrutu: I tried to give you an example why that is not possible to happen.

SergeyKanzhelev:
Extra comment that requires the mention in rationale and discussion issue is case sensitivity of names. Is there anything else I need to change before merging?

codefromthecrypt:
At least for this round, I suggest we merge then take follow-ups for
tuning. Sergey has done a good job on revisions, but lets not wear him out
;)

bogdandrutu:
Please merge.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
yurishkuro: > Another option is to use prefixed headers, such as trace-context-X

I'd rather not. Jaeger uses that, and it puts restriction on which keys people want to use for baggage, even when the transport is NOT http but some binary protocol that can perfectly preserve any unicode keys.

SergeyKanzhelev:
@bogdandrutu I'm not a contributor. Unless you've been asking @adriancole

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
yurishkuro: both names and values are on the right-hand side of the header, I agree with @bogdandrutu

yurishkuro:
please don't merge until it's available for review for a couple of days after the last change

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
jcchavezs: I see, then forget about my doubt.

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Purpose of a header
+
+Correlation context header is used to propagate properties not defined in `Trace-Context`. There are two common use cases. First is to define a context on trace initiation. Such context will have customer's identity, high-level operation name and other properties like a flight name. Second use case is to pass the caller's name to the next component. This name-value pair will be overridden by every service to it's own name.
+
+## General considerations
+
+- Should be human readable. Cryptic header will hide the fact of potential information disclosure.
+- Should be appende-able (comma-separated) https://tools.ietf.org/html/rfc7230#page-24 so nodes can add context properties without parsing an existing headers.
+- It is expected that the typical name will be a single word in latin and the value will be a short string in latin or a derivative of an url.
```
yurishkuro, line 13: -1. Counter example: `auth-token={long-base64-string}`

SergeyKanzhelev: url encode will not inflate base64 value too much unless it has a lot of `/` and `+`. It will loose readability in favor of consistent encoding pattern. Do you see this as a major issue?

SergeyKanzhelev:
If this PR is waiting, I'll create issues to start discussion: 

type - type of the value https://github.com/TraceContext/tracecontext-spec/issues/22
ttl - number of hops to live https://github.com/TraceContext/tracecontext-spec/issues/23
pii - level of access to data https://github.com/TraceContext/tracecontext-spec/issues/24
discussion for the max total length limit https://github.com/TraceContext/tracecontext-spec/issues/25
discussion for well-known context names catalog https://github.com/TraceContext/tracecontext-spec/issues/26

SergeyKanzhelev:
@yurishkuro - I started timer =) I'll try to divert all comments to specific issues unless there are critical feedback that will block the merge

codefromthecrypt:
There was a comment on the suggestions for rationale. The "prefix" and
"vary" approaches were mentioned as something we could add to rationale as
why we only consider this option.

That said one comment against prefix , by yuri, was that prefix doesnt work
well outside http. This is an http format, and so we havent added scope for
this format to also be not http. I dont think adding this rationale is
great unless we expand the scope to non-http. We should think of a
different reason or say why this is also not http.

codefromthecrypt:
Pushing auth-tokens in trace data is one of the things security folks and
other engineers have cringed at the idea of eventhough some sites like uber
might do this. Let's please not document this example.. choose something
else less polarizing and less likely to raise antibodies.

SergeyKanzhelev:
Added rationale for `Vary` and multiple headers approach.

codefromthecrypt:
fyi I'm still :shipit:

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+
+A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Context`. The values should be passed along to any child requests. Note that uniqueness of the key within the `Correlation-Context` is not guaranteed. Context received from upstream service may be altered before passing it along.
+
+# Format
+
+## Header name
+
+`Correlation-Context`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to the [rfc](https://tools.ietf.org/html/rfc7230#page-24).
```
yurishkuro: nit: `s/rfc/RFC 7230/`

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+## Header value
+
+`name1=value1[;properties1],name2=value2[;properties2]`
+
+**Limits:**
+1. Maximum number of name-value pairs: `180`.
+2. Maximum number of bytes per a single name-value pair: `4096`.
+3. Maximum total length of all name-value pairs: `8192`.
+
+*See [CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md#Limits] for rationale of these limits.*
```
yurishkuro: should this be turned to a link?

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Name format
+
+Url encoded string. Spaces should be trimmed from beginning and the end of the name. Names are case sensitive.
+
+## Value format
+
+All spaces should be trimmed from the beginning and the end of the value. Value ends with the special character `;`, separator `,` or end of string. Value represents a url encoded string and case sensitive. 
+
+**Examples**:
```
yurishkuro: this section seems redundant, as it's followed by larger Examples section

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Examples of HTTP headers
+
+```
+Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true
+```
+
+```
+Correlation-Context: component=Frontend
+Correlation-Context: flight%3DName=DF:28, ExposurePercentage=33.33
```
yurishkuro: invalid example? The colon in `DF:28` would be url-encoded

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Header name
+
+`Correlation-Context`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to the [rfc](https://tools.ietf.org/html/rfc7230#page-24).
+
+## Header value
+
+`name1=value1[;properties1],name2=value2[;properties2]`
```
yurishkuro: I'd add a note that that `properties` are expected to be in the format `;k1=v1;k2=v2`, but the exact spec will be added later

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Strings encoding
+
+Url encoding is low-overhead way to encode unicode characters for non-latin characters in the values. Url encoding keeps a single words in latin unchanged and easy readable.
+
+## Limits
+
+The idea behind limits is to provide trace vendors common safeguards so the content of the `Correlation-Context` header can be stored with the request. Thus the limits are defined on the number of keys, max pair length and the total size. The last limit is the most important in many scenarios as it allows to plan for the data storage limits.
+
+Another consideration was that cookies may be an alternative way to store and pass the correlation context fields. So the limits should make context name-value pairs fit the typical cookie limit.
```
yurishkuro: > cookies may be an alternative way

that's misleading, and most likely incorrect iirc how cookie semantics are defined. But drawing a parallel with passing custom data via headers as cookies makes sense. I suggest rephrasing:

Another consideration was that HTTP cookies provide a similar way to pass custom data via HTTP headers. So the limits should make the correlation context name-value pairs fit the typical cookie limits.

yurishkuro (approved):
minor clean-ups, otherwise lgtm

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
codefromthecrypt: ps I consider my comment addressed as the rationale section now exists

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Examples of HTTP headers
+
+```
+Correlation-Context: component=Frontend, flightName=DF:28, IsAuthenticated=true
+```
+
+```
+Correlation-Context: component=Frontend
+Correlation-Context: flight%3DName=DF28, ExposurePercentage=33.33
```
wu-sheng: Is the whitespace between key-value-pair optional?

wu-sheng: The implementation should remember to trim, after split by `,`

wu-sheng (approved):
LGTM. I have only a question to be sure.

bogdandrutu:
@SergeyKanzhelev please fix last comments from @wu-sheng and we are good to merge on this. We have this LGTM with only minor comments for the last 2-3 days. @yurishkuro is that ok?

yurishkuro:
+1

SergeyKanzhelev:
I made examples more explicit regarding spaces

wu-sheng (approved):
Thanks. LGTM

CORRELATION_CONTEXT_HTTP_HEADER_FORMAT_RATIONALE.md
mnot: What does "Vary-style" even mean here? Can someone please just specify some [ABNF](https://tools.ietf.org/html/rfc5234) for the syntax being proposed?

codefromthecrypt:
>
> What does "Vary-style" even mean here?
>
> was from here, but agree it isn't detailed well in the rationale
https://github.com/TraceContext/tracecontext-spec/pull/21#discussion_r140977140

What I meant here is that for example the Vary header is metadata about
caching which contains a value of a list of other
headers.

Ex. today you have folks passing headers without an industry standard, such
as B3 <https://github.com/openzipkin/b3-propagation>

ex.
x-b3-traceid: 13871f00ea84241febf33e1a81dc6f71
x-b3-parentspanid: bd7a977555f6b982
x-b3-spanid: be2d01e33cc78d97
x-b3-sampled: 1

There might be some signal loss involved with porting to the current
'trace-context' header:

trace-context: 00-13871f00ea84241febf33e1a81dc6f71-be2d01e33cc78d97-01

In this case, the parentspanid header isn't conveyed. Rather than embedding
a field with the same purpose (potentially conflicting on field names), we
could have a header to push the names of other required headers

trace-context: 00-13871f00ea84241febf33e1a81dc6f71-be2d01e33cc78d97-01
correlation-context: x-b3-parentspanid;ttl=1
x-b3-traceid: 13871f00ea84241febf33e1a81dc6f71
x-b3-parentspanid: bd7a977555f6b982
x-b3-spanid: be2d01e33cc78d97
x-b3-sampled: 1

In the above case, a client can push both the old and the standard header
format, indicating to the recipient that there's a correlated header which
shouldn't be dropped via an intermediary's header forwarding policy.

For example, correlation-context: x-b3-parentspanid;ttl=1 tells a proxy to
pass the x-b3-parentspanid even if it drops other data, but that it is only
valid to the next hop (hop defined as a hop that participates in tracing).

I didn't see any interest in this proposal, surely it has lots of flaws,
but yeah that's what I meant by Vary approach (clearly not a good name :P)
make sense?

codefromthecrypt:
and before I raise antibodies, it wouldn't be required to pass two
types of headers, I was just calling out the brown-field use case this
supports, which is that you don't know if downstream supports the
standard header or not, and you don't want to break the tracing you
actually support

codefromthecrypt:
sorry to spam, but to further explain.. it is implicitly the case that
correlation-context means correlated with trace-context. The "formerly
known as Vary style" approach could be completely independent, ex to
pass trace-context, you have to add it to the correlation-context:

correlation-context: trace-context;ttl=1,x-b3-parentspanid;ttl=1

For example, if you are a proxy, you could read the
correlation-context to see which minimal headers the caller wants it
to push, even vendor headers, even headers unrelated to tracing (like
billing code). Obviously a proxy could choose to ignore this advice.

PR #27, opened 2017-09-28
Remove the HTTP description.

wu-sheng:
We have sections about Binary Format and HTTP Format. So this spec is not only about HTTP, right? 

I adjust the goal.


codefromthecrypt (approved):
thanks. this just removes a few words which were invalid, so seems safe!

PR #29, opened 2017-09-29
Remove some meaningless words about goal.

wu-sheng:
The interop heads are between tracers, right. The sentence in goal makes meaningless to me. Please correct me if I misunderstand this. @adriancole @bogdandrutu 

First of all, TraceContext Spec HTTP HEADS influence anything except the load balance or gateway has security strategy, right? Even they know this spec, they must ask to permit this HEADS with manual config. Security is the highest priority for these system, right.

And for Binary, it is hard to say. It depends on the RPC framework usages. But in general, it is same as HTTP HEADS, if they support something similar, e.g. head, attachment, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 Specification for TraceContext propagation format.
 
 ## Goal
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
I think this is fine, as we elaborate things later in the file now.

PR #30, opened 2017-10-04
Refactor the repository structure.

bogdandrutu:
Moved trace-context and correlation-context files into different directories. No text changes except that long lines are split between multiple lines.

/cc @SergeyKanzhelev @yurishkuro


wu-sheng (commented):
Sorry for the delay. In the National Day Vacation :)

IMHO, so far, we are still discussing propagation through HTTP. I prefer to move all HEADER and CORRELATION files into `trace_context/http` dictionary. Because I think, in binary format, there should be an whole new data structure, in `trace_context/binary` dictionary.

codefromthecrypt:
Sorry for the delay. In the National Day Vacation :)

Glad you took some time off! Hope was fun

IMHO, so far, we are still discussing propagation through HTTP. I prefer to
move all HEADER and CORRELATION files into trace_context/http dictionary.
Because I think, in binary format, there should be an whole new data
structure, in trace_context/binary dictionary.

Enoding is different sure, but the fields would be the same right?
Otherwise might be misleading to call a format the same name if you cannot
convert between representations of it. I was thinking we can describe the
fields generically and then how to map them to http or binary separately.
Are you thinking similarly?

I wouldnt expect too many files to need a separate directory for each
format readme of the same struct. Suppose question I have is whether you
are organizing file directories or anticipate incompatible structs..

wu-sheng:
> I was thinking we can describe the
fields generically and then how to map them to http or binary separately.
Are you thinking similarly?

+1. I agree. If we don't have so many files, we can keep them in the root.

mtwo (approved):

mtwo:
I'm guessing that we'll want to give baggage (if it's included) its own directory, much like how we're differentiating between trace and correlation context?

codefromthecrypt:
I think we are good here.. we can leave binary and text format for trace-context in the root (of the tracecontext directory). thx for the feedback!

codefromthecrypt:
>
> I'm guessing that we'll want to give baggage (if it's included) its own
> directory, much like how we're differentiating between trace and
> correlation context?
>
if there's a separate mechanism for baggage vs correlation context, SG.
Since folks might squint these the same, let's chat about it during the
workshop in a couple weeks (even if we progress before then)

Issue #41, opened 2017-11-04
Define purpose of correlation context


AloisReitbauer:
Currently the definition and example of what correlation context is about are misleading. We should define that this is for tracing tool usage only. We should adjust the examples to reflect vendor specific extension like hierarchic correlation.

Happy to take the first run at this one.

wu-sheng:
> Currently the definition and example of what correlation context is about are misleading.

Mislead to where? I am a little confused. This whole repository is about tracing context propagation. You can see that in our goal: https://github.com/TraceContext/tracecontext-spec#goal

codefromthecrypt:
>
> Mislead to where? I am a little confused. This whole repository is about
> tracing context propagation. You can see that in our goal:
> https://github.com/TraceContext/tracecontext-spec#goal
>
> Hi, Sheng. I will export a copy of the workshop notes to you in PDF. Main
thing is that the correlation-context is not adoptable by a number of
parties due to it being unbounded and other reasons. One way we discussed
to make trace-context in general adoptable by major players is
re-associating the correlation-context to essentially the carrier of
extended *required* data about the trace context. For example, this can
hold extra fields needed by the originator of the trace. This is different
than its former scope which was arbitrary data.

https://docs.google.com/document/d/1uZpf7-X1cojSMhghYGP4O5DVVHtISrDq-HlwBKaDjDg/edit

AloisReitbauer:
I was referring to the correlation context. We should not have an example with username etc. in there. Makes sense?

yurishkuro:
maybe it should not be called a general purpose `Correlation-Context` then, but mention tracing, e.g. `Trace-Correlation`

SergeyKanzhelev:
I agree with @yurishkuro - `correlation-context` better suites the baggage. Maybe `Trace-Context-Externded` or `Trace-Context-All`? 

Alternatively use `Trace-Context` for name/value pairs and `Trace-Id` for "essential" fields extracted for performance.

Any suggestions on thread?

CC: @bogdandrutu explicitly

codefromthecrypt:
what? someone dared to suggest a naming convention?! :) I like
trace-context-ext (eventhough it stutters)

bogdandrutu:
👍  "trace-context-ext"

wu-sheng:
`trace-context-ext` is good for me.

SergeyKanzhelev:
Is PR looks ok?

wu-sheng:
[...]

PR #42, opened 2017-11-07
trace-context-ext header

SergeyKanzhelev:
#41

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,13 @@ together.
 
 ## Headers
 
-TODO: explain why two headers here.
+Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these headers were promoted to a separate header. This header has fixed length and defined sequence of fields.
 
-* See [TraceContext Header](trace_context/README.md)
-* See [CorrelationContext Header](correlation_context/README.md)
+Libraries and platforms SHOULD propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace.
+
+* See [Trace-Context Header](trace_context/README.md)
+* See [Trace-Context-Ext Header](trace_context_ext/README.md)
+* See [Correlation-Context Header](correlation_context/README.md)
 
 ## Reference Implementations
 TODO: add link here
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ together.
 
 ## Headers
 
-Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these headers were promoted to a separate header. This header has fixed length and defined sequence of fields.
+Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
 
 Libraries and platforms SHOULD propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace.
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ together.
 
 Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
 
-Libraries and platforms SHOULD propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace.
+Libraries and platforms MUST propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace. 
+
+TODO: Add details on behavior when one of the headers cannot be parsed
 
 * See [Trace-Context Header](trace_context/README.md)
 * See [Trace-Context-Ext Header](trace_context_ext/README.md)
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ together.
 
 Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
 
-Libraries and platforms MUST propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace. 
+Libraries and platforms MUST propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace. Libraries and platforms MAY propagate this header.
 
 TODO: Add details on behavior when one of the headers cannot be parsed
 
--- a/trace_context_ext/HTTP_HEADER_FORMAT.md
+++ b/trace_context_ext/HTTP_HEADER_FORMAT.md
@@ -40,13 +40,13 @@ Properties are expected to be in a format of keys & key-value pairs `;` delimite
 Single header: 
 
 ```
-Trace-Context-Ext: epoch=123234
+Trace-Context-Ext: parent_application_id = 123
 ```
 
 Context might be split into multiple headers:
 
 ```
-Trace-Context-Ext: ms_h_id=1.2.2
-Trace-Context-Ext: epoch=123234
+Trace-Context-Ext: parent_application_id = 123
+Trace-Context-Ext: trace_roads = App1%7cApp2%7cApp
 ```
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -8,10 +8,13 @@ together.
 
 ## Headers
 
-TODO: explain why two headers here.
+Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these headers were promoted to a separate header. This header has fixed length and defined sequence of fields.
```
yurishkuro: s/few of these headers/few of these pairs/ (or fields)

trace_context_ext/HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,52 @@
...
+```
+Trace-Context-Ext: epoch=123234
+```
+
+Context might be split into multiple headers:
+
+```
+Trace-Context-Ext: ms_h_id=1.2.2
+Trace-Context-Ext: epoch=123234
+```
```
wu-sheng: I think the examples should be more specific. e.g.

```
Trace-Context-Ext: parent_application_id = 123
Trace-Context-Ext: trace_roads = App1%7cApp2%7cApp3
```

wu-sheng (approved):
LGTM. Only give a suggestion.

README.md
```diff
@@ -8,10 +8,13 @@ together.
 
 ## Headers
 
-TODO: explain why two headers here.
+Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
 
-* See [TraceContext Header](trace_context/README.md)
-* See [CorrelationContext Header](correlation_context/README.md)
+Libraries and platforms SHOULD propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace.
```
codefromthecrypt: SHOULD to MUST I think? Otherwise we are specifying something one can meet without being compatible with anything, right?

trace_context_ext/HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,52 @@
+# Trace Context Extended HTTP Header Format
+
+A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Context`. The values should be passed along to any child requests. Note that uniqueness of the key within the `Correlation-Context` is not guaranteed. Context received from upstream service may be altered before passing it along.
```
codefromthecrypt: typo s/Correlation-Context/Trace-Context-Ext

trace_context_ext/HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,52 @@
...
+`Trace-Context-Ext`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to the [RFC 7230](https://tools.ietf.org/html/rfc7230#page-24).
+
+## Header value
+
+`name1[=value1[;properties1]],name2[=value2[;properties2]]`
+
+**Limits:**
+Maximum length of a combined header should be less than 300 Kb.
```
codefromthecrypt: Appreciate the strawman, but this sounds rediculously high for what we'd like to be mandatory... is this just a copy/paste?

Let's use http/2 for example. The max frame size is 16KiB, and there is some small overhead. If we can specify a limit which is under the frame size, it will prevent continuations. https://http2.github.io/http2-spec/#FrameHeader

Simultaneously, a lower value makes the data itself easier to store on trace restart, specially on messaging transports, which are often limited well below 300KiB per message.

Finally, 300KiB is orders of magnitude larger than some production payloads, for adoptability, we should be careful to not disincentivize people from using the spec due to its heft.

trace_context_ext/HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,27 @@
+#  Trace Context Extended HTTP Header Format Rationale
```
codefromthecrypt, line 1: hints we may eventually need to extract a definition of the comma separatey format as it would mostly be reused for trace-context-ext and correlation-context so far.

trace_context_ext/README.md
```diff
@@ -0,0 +1,12 @@
+# Trace-Context-Ext Header
```
codefromthecrypt: this looks good in general.. nit I might title this Trace-Context-Ext *Field* as header is transport specific

trace_context_ext/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Trace context is a set of common and vendor specific properties identifying the trace. `Trace-Context` header is designed for fast parsing - it has a predefined structure and fixed length fields. `Trace-Context-Ext` header carries the extended properties in a form of collection of name/value pairs.
+
+`Trace-Context-Ext` header defined separately from `Correlation-Context`. It designed to carry ONLY properties defined by tracing library, not user-defined properties. This way cloud vendors and libraries may guarantee it's value transmission. Transmission of a larger baggage header is not guaranteed.  
+
+## HTTP Format
+The HTTP format is defined [here](HTTP_HEADER_FORMAT.md) and the rationale is defined
```
codefromthecrypt: nit HTTP header format

codefromthecrypt (commented):
Made some comments which are likely mostly copy/paste related. thanks for kicking this off!

README.md
codefromthecrypt: I think the trickiness is in the form here.. ex A library must propagate an incoming trace-context with any trace-context-ext. If not, it must restart or abandon the trace. (ack this may take time to elaborate, but in such a case adding a TODO to this doc with a link is also fine).

trace_context_ext/HTTP_HEADER_FORMAT.md
adriancole: switched my comment above to me a bit less zipkin-specific. making it more about http limits

SergeyKanzhelev: I meant bytes =))) thinking more  - 300 bytes may be too little. Maybe 2Kb limit (as typical URL limit). Assuming URL limit allows to hold enough details typically to address the resource?

codefromthecrypt:
>
> I meant bytes =))) thinking more - 300 bytes may be too little. Maybe 2Kb
> limit (as typical URL limit). Assuming URL limit allows to hold enough
> details typically to address the resource?
>
whew! I think we should try to explicitly get a nod from folks about this
limit, and use a concrete vs a theoretical constraint. 2KiB is still quite
a lot of room for abuse (ex using this as a form of propagation storage),
but anyway far better than 300KiB :) Gut feeling is that this can be well
under a kilobyte (inclusive of the non-extended part). @CodingFabian
thoughts?

wu-sheng:
@adriancole 2kb is a lot for head. Especially allow multi headers, so the limit is for each one, I suppose.

bogdandrutu:
@adriancole @SergeyKanzhelev @wu-sheng I personally prefer 512B (or lease) total is possible.

I think for languages like C++ in order to avoid a malloc every time pre-allocating 512B is not the end of the world.

Even if we go with higher value I think it should be for the total size (including all multi headers).

CodingFabian:
Thoughts?
yes, I think that is reasonable, as it talks about the combined header, not individual kv pairs.
I do not think it makes sense to send that data, but its not too large. Is there maybe http RFC that mentiones similar header lengths we could reference?

SergeyKanzhelev:
@wu-sheng no, the limit is for combined header. 512 looks OK for me. I'll update PR

SergeyKanzhelev:
@bogdandrutu @adriancole would you please merge this PR?

README.md
```diff
@@ -8,10 +8,15 @@ together.
 
 ## Headers
 
-TODO: explain why two headers here.
+Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
 
-* See [TraceContext Header](trace_context/README.md)
-* See [CorrelationContext Header](correlation_context/README.md)
+Libraries and platforms MUST propagate `Trace-Context` and `Trace-Context-Ext` headers to guarantee that trace will not be broken. `Correlation-Context` header is a companion header representing user-defined baggage associated with the trace.
```
bogdandrutu: Add a MAY propagate for the correlation-context.

trace_context_ext/HTTP_HEADER_FORMAT.md
```diff
@@ -0,0 +1,52 @@
...
+`Trace-Context-Ext`
+
+Multiple correlation context headers are allowed. Values can be combined in a single header according to the [RFC 7230](https://tools.ietf.org/html/rfc7230#page-24).
+
+## Header value
+
+`name1[=value1[;properties1]],name2[=value2[;properties2]]`
+
+**Limits:**
+Maximum length of a combined header MUST be less than 512 characters.
```
bogdandrutu: a character has an encoding: for example UTF16, so a lot of work to determine the size in characters. Better to say 512 bytes.

SergeyKanzhelev:
Sure. Done

codefromthecrypt (approved):

codefromthecrypt:
Thanks for working together on this folks, notably sergey for tracking requests. If you have any changes you feel need discussion, please open an issue

PR #45, opened 2017-12-22
add comment that repository was moved

SergeyKanzhelev:


wu-sheng (approved):

tylerbenson:
It's a shame we couldn't migrate the repo on Github's side like @wu-sheng did with Skywalking.... Was that option already considered and decided against for some reason?

wu-sheng:
@tylerbenson The tranditional ASF community people just accept the way. :) Maybe w3c people are even more tranditional? ;P

SergeyKanzhelev:
@tylerbenson this would be ideal. I just sent another e-mail. Now to W3C support.

SergeyKanzhelev:
Not relevant any more

PR #49, opened 2018-01-17
split one report into two

SergeyKanzhelev:
Following yesterday meeting splitting one report into two separate - one for `Trace-Context` and `Trace-Context-Ex` and another for `Correlation-Context`. Motivation for a split is 
1. Allow `Correlation-Context` be implemented independently from trace context and serve wider purpose. 
2. Highlight the significant difference of handling these headers. Trace context is mandatory to pass, correlation context is optional


codefromthecrypt (approved):
thanks for doing this

Issue #50, opened 2018-01-17
Define behavior for response header context propagation


SergeyKanzhelev:
There are scenarios when services need to return correlation information in response. Scenarios include:

- Sampling flag (+ sampling score) for delegated sampling
- Tenant ID/identity of the service so caller knows where to query telemetry from

Spec need to define which headers SDK may expect and service should use for http response. Also behavior for service mash and proxies needs to be defined so headers would not be lost.

codefromthecrypt:
one thing interesting about this is whether or not downstream need to be
considered for the data propagated back. The two you mentioned could be
resolved locally, ex not depending on the result of any potential
downstream calls.

The complexity of sending back sampling flag, trace ID (if for example none
was provisioned), tenant ID is not hard, but some of these data imply
either taking roles others are doing independently not (ex tenant id is
often a separate header and unrelated to tracing).

I guess you mention different types of data because you are asking if
broadly speaking, trace-context and correlation-context are to be
propagated back, such that they can be used for other middleware concerns
such as knowing the tenant ID. Is that right?

SergeyKanzhelev:
Yes, that was the main point. We need information that is required to build a trace to be returned back in response in certain scenarios. First scenario is for `sampled` flag that is part of `Trace-Context` header and second is about *solution specific* tenant id that is part of `Trace-Context-Ex`.

And we need to make sure those headers will be preserved and passed further in response for proxies and balancers. But we cannot demand it from all services. 

Also - does it ever make sense to return `trace-id` in response headers? Or we can limit spec to only allow `Trace-Context-Ex` header to be returned.

codefromthecrypt:
>
> Yes, that was the main point. We need information that is required to
> build a trace to be returned back in response in certain scenarios. First
> scenario is for sampled flag that is part of Trace-Context header and
> second is about *solution specific* tenant id that is part of
> Trace-Context-Ex.
>
thx for clearing it up

> And we need to make sure those headers will be preserved and passed
> further in response for proxies and balancers. But we cannot demand it from
> all services.
>
ok so the implementor can't blindly overwrite or depend on local state. it
has to be prepared to consider and parse an incoming trace context

> Also - does it ever make sense to return trace-id in response headers?
>
was trying to figure this out. For example, there are middleware
discussions of passing trace ID back which occurred in the past
https://github.com/openzipkin/b3-propagation/issues/4

> Or we can limit spec to only allow Trace-Context-Ex header to be returned.
>
[...]

PR #51, opened 2018-01-18
Define response headers behavior - old

SergeyKanzhelev:
Initial set of behavior definition. Mostly to address #50


abstract/trace-context-overview.md
```diff
@@ -1,7 +1,11 @@
 # Overview
 
-Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
+Trace context represented by set of name/value properties describing identity of every http request. Few of these properties are promoted to a separate header. It also allows to align different tracing frameworks on semantic of these properties. This is also done as a performance optimization measure - header has fixed length and defined sequence of fields. It is easy to parse and mutate this header.
```
codefromthecrypt: Stab at a different overview:

The trace context identifies an http request and its position in a potentially distributed operation. Its properties exist in a fixed-length generic header, `Trace-Context`, and possibly extend into a vendor-specific `Trace-Context-Ext` header. The `Trace-Context` defines a sequence of fields of fixed length all vendors must process, such as a trace ID. If a vendor cannot process `Trace-Context-Ext`, it must restart the trace. The following elaborates format details and propagation semantics implied by this standard.

abstract/trace-context-overview.md
```diff
@@ -1,7 +1,13 @@
...
 
-**TODO:** Add details on behavior when one of the headers cannot be parsed
+Libraries and platforms SHOULD strive to minimize the number and size of trace context properties. Shortening size of properties allows libraries implementing different scheme of trace correlation to append properties needed for that library. So different vendors may co-exist in the different components of the same distributed trace.
+
+Headers `Trace-Context` and `Trace-Context-Ext` can be used to pass http request identity information as the request and response headers. Libraries and platforms MAY mutate these headers according to the semantics described in corresponding sections. Libraries and platforms MUST propagate `Trace-Context` and `Trace-Context-Ext` request headers to all outgoing calls. Propagation of the header guarantees that trace correlation is not broken. Libraries and platforms SHOULD propagate `Trace-Context` and `Trace-Context-Ext` response headers.
```
yurishkuro: "propagate response headers" sounds wrong. The typical behavior of a server would be to send back the same trace context as received in the incoming request, not propagate something that came from downstream service,

abstract/trace-context-overview.md
```diff
@@ -1,7 +1,13 @@
...
+
+Headers `Trace-Context` and `Trace-Context-Ext` may appear on request and response independently. Missing `Trace-Context-Ext` header means that there are no additional properties defined in the request. Missing `Trace-Context` header means that library or platform has not started the trace. However has some tracing related information that identifies the request. See corresponding section for the examples of properties that can be used in `Trace-Context-Ext` header.
```
yurishkuro: "However has some tracing related information" - what did you mean?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -1,7 +1,7 @@
 # Trace Context HTTP Header Format
 
-A trace context header is used to pass trace context information across systems for a HTTP 
-request. Our goal is to share this with the community so that various tracing and diagnostics 
+A trace context header is used to pass trace context information across systems for an HTTP
```
yurishkuro: I would avoid providing justifications here to remove redundancy. The purpose of the header is already explained in the common section. Enough to say that this part defines the format of the header specifically for HTTP protocol.

abstract/trace-context-overview.md
SergeyKanzhelev: Clarified. I think

bogdandrutu:
Can you please resync?

AloisReitbauer:
Overall, we need to work on a more precise definition of a processing model anyways. I propose to commit the current state with a todo line on what needs to be still discussed. 

Here is an example of a processing model: https://www.w3.org/TR/beacon/#sec-processing-model

AloisReitbauer:
What should we do with this PR?

PR #59, opened 2018-02-07
renamed headers

SergeyKanzhelev:
Simple renaming as a first step towards https://github.com/w3c/distributed-tracing/issues/57 This will minimize conflicts of editing specs forward


abstract/trace-context-overview.md
```diff
@@ -2,6 +2,6 @@
 
 Trace context represented by set of name/value pairs describing identity of every http request. As a performance optimization measure few of these pairs were promoted to a separate header. This header has fixed length and defined sequence of fields.
```
yurishkuro, line 3: is the file name still correct?

SergeyKanzhelev: it is an overview of both headers. Two of them makes a trace context.

trace_state/HTTP_HEADER_FORMAT.md
```diff
@@ -23,15 +23,15 @@ Maximum length of a combined header MUST be less than 512 bytes.
 
 Url encoded string. Spaces MUST be trimmed from beginning and the end of the name. Names are case sensitive. Vendor libraries encouraged to use namespaced names of the properties to avoid conflicts.
 
-Names `id`, `span-id`, `trace-id`, `sampled` are reserved. These properties are defined in `Trace-Context` header.
+Names `id`, `span-id`, `trace-id`, `sampled` are reserved. These properties are defined in `Trace-Parent` header.
```
wu-sheng, line 26: Most likely, we don't **reserve** these ids for `Trace-Parent`, because only the vendor-segment (just give a name) will be trusted by implementors.

trace_state/HTTP_HEADER_FORMAT.md
```diff
@@ -1,14 +1,14 @@
...
+A correlation context header is used to pass the name-value context properties for the trace. This is a companion header for the `Trace-Parent`. The values MUST be passed along to any child requests. Note that uniqueness of the key within the `Trace-State` is not guaranteed. Context received from upstream service may be altered before passing it along.
 
 *See [rationale document](HTTP_HEADER_FORMAT_RATIONALE.md) for details of decisions made for this format.*
 
 # Format
 
 ## Header name
 
-`Trace-Context-Ext`
+`Trace-State`
```
wu-sheng, line 11: `Trace-State` should make a statement about what are inside. Such as: vendor-name=vendor-context-segment . And make an example for it. Even we have not discussed about vendor-context-segment .

trace_state/HTTP_HEADER_FORMAT.md
wu-sheng: I think most of the commercial vendors will keep their trace-id, span-id in this part too, just like DynaTrace and AWS suggest.

wu-sheng (commented):
Some suggestions about make `Trace-State` more clear.

trace_state/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: yes. I still think we need to avoid using these names in `Trace-State` to avoid confusions.

trace_state/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Let's make it in separate PR. I want to merge renaming to minimize conflicts. File renamings makes merges nasty

SergeyKanzhelev:
@wu-sheng @adriancole I need a sign off to merge this one.

wu-sheng (approved):
@SergeyKanzhelev Sure. Let's use a separated pr to make `trace-state` more clear.

trace_state/HTTP_HEADER_FORMAT.md
wu-sheng: @SergeyKanzhelev I argue about these names, just because we can't sure the vendors are going to use key=value in vendor-segments, or whether should we suggest to do so. That is based on encoding parts of `trace-state`.

codefromthecrypt:
I think this part is separate issue. Lets merge the rename and keep
improving

Shipit!

wu-sheng:
@SergeyKanzhelev @adriancole Merged.

PR #63, opened 2018-02-12
Fixes examples for Trace-State

cwe1ss:
Text says there should be no spaces. And I think the `;` in the other example shouldn’t be there either.

### DIFF
--- a/trace_state/HTTP_HEADER_FORMAT.md
+++ b/trace_state/HTTP_HEADER_FORMAT.md
@@ -21,7 +21,7 @@ Maximum length of a combined header MUST be less than 512 bytes.
 
 ## Name format
 
-Url encoded string. Spaces MUST be trimmed from beginning and the end of the name. Names are case sensitive. Vendor libraries encouraged to use namespaced names of the properties to avoid conflicts.
+Url encoded string. Spaces MUST be trimmed from the beginning and the end of the name. Names are case sensitive. Vendor libraries encouraged to use namespaced names of the properties to avoid conflicts.
 
 Names `id`, `span-id`, `trace-id`, `sampled` are reserved. These properties are defined in `Trace-Parent` header.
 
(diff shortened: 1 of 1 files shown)


wu-sheng:
@cwe1ss Based on the last workshop, `Trace-State` is not just a new name instead of old `Trace-Context-Ext`. We should redefine it. Since English is not my primary language, I am waiting @SergeyKanzhelev or @adriancole or others to do so. 

I am not saying the spaces issue is irrelevant, just for clear.

cwe1ss:
No worries! Feel free to close this issue, if you don’t need it.

trace_state/HTTP_HEADER_FORMAT.md
```diff
@@ -33,20 +33,20 @@ All spaces MUST be trimmed from the beginning and the end of the value. Value en
...
-Properties are expected to be in a format of keys & key-value pairs `;` delimited list `;k1=v1;k2;k3=v3`. All unknown property names and name-value pairs MUST be preserved. 
+Properties are expected to be in a format of keys & key-value pairs `;` delimited list `k1=v1;k2;k3=v3`. All unknown property names and name-value pairs MUST be preserved. 
 
 # Examples of HTTP headers
 
 Single header: 
 
 ```
-Trace-State: parent_application_id = 123
+Trace-State: parent_application_id=123
```
SergeyKanzhelev, line 43: spaces were intentional to demo that you CAN do spaces here. Do you think we should have two examples - one specifically to demonstrate spaces trimming?

SergeyKanzhelev:
The meaning of the phrase is "if spaces are there - feel free to trim them while reading". Can you suggest how to re-phrase it?

cwe1ss:
On this page it says *"Spaces MUST be trimmed from beginning and the end of the [name|value]"*. If I understand correctly, this affects the producers of headers, right?

I guess that consumers/parsers would still look for spaces and trim them but given the previous definition, I wouldn't put spaces into the examples.

But again, feel free to close this if you are going to rewrite it anyway.

codefromthecrypt:
Trace-State might well include the option to allow (sad) things like space
between equals. Curious though.. is it pretty normal to the degree we want
to optimize for putting spaces after equals? I havent seen this but if so I
suppose we will want to keep this. Also citing other headers that do space
after equals trimming might help prevent others from scratching their heads.

Regardless our examples should clearly promote sane things and loudy
mention if we also handle not sane things. Otherwise we might accidentally
encourage crappy values

cwe1ss:
Looking at other http headers it seems to be common to not have spaces around the `=` but it seems to be common to have a space between key:value pairs.

* [Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) (which probably is the most similar one)
* [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)
* [X-Forwarede-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)

e.g. `Cookie: name=value; name2=value2; name3=value3`

So maybe following the `Cookie` format might be a good way? (For `Trace-State` AND `Correlation-Context` to be consistent)

SergeyKanzhelev:
@cwe1ss I opened this PR to clarify spaces behavior: https://github.com/w3c/distributed-tracing/pull/66

SergeyKanzhelev:
closing this PR. Feel free to re-open if needed

PR #86, opened 2018-03-15
mutating of traceparent

### LINKED ISSUE #79: Should there be a recommendation on how trace IDs are generated?
This was raised in Jaeger [here](https://github.com/jaegertracing/jaeger-client-java/issues/318#issuecomment-371843638), but it seems like an interesting aspect if we expect different tracing systems to share the same trace ID.

Is [RFC 4122](https://tools.ietf.org/html/rfc4122.html) way of generating UUIDs good enough for tracing purposes? Or is it better to use fully random IDs?

SergeyKanzhelev:
List of allowed mutations to the `traceparent`. No attempt to resolve #79.

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -108,6 +108,21 @@ make a decision to trace and the decision might be deferred). When `trace-option
 the default value for this bit is `0`
 * The behavior of other bits is currently undefined.
 
+# Mutating the header 
+
+Library or platform receiving `traceparent` request header MUST send it to outgoing http requests. It MAY mutate the value of this header before passing to outgoing http requests.  
+
+Library or platform receiving `traceparent` response header MAY send it as a response header to originating http request. Proxies, load balancers, and other platforms with 1:1 mapping of request and response SHOULD return this `traceparent` header to originating http request. `traceparent` response header MAY also be mutated.
+
+Here is the list of allowed mutations: 
+
+1. **Update `span-id`**. The value of property `span-id` can be regenerated. 
+2. **Mark trace for sampling**. The value of `sampled` flag of `trace-options` may be set to `1` if it had value `0` before. `span-id` MUST be regenerated with the `sampled` flag update. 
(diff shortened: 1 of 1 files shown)


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -108,6 +108,21 @@ make a decision to trace and the decision might be deferred). When `trace-option
...
 
+# Mutating the header 
+
+Library or platform receiving `traceparent` request header MUST send it to outgoing http requests. It MAY mutate the value of this header before passing to outgoing http requests.  
+
+Library or platform receiving `traceparent` response header MAY send it as a response header to originating http request. Proxies, load balancers, and other platforms with 1:1 mapping of request and response SHOULD return this `traceparent` header to originating http request. `traceparent` response header MAY also be mutated.
+
+Here is the list of allowed mutations: 
+
+1. **Update `span-id`**. The value of property `span-id` can be regenerated.
```
yurishkuro, line 119: What does this mean if done for the response?

SergeyKanzhelev: there are different cases. This mutation basically says you can receive the response header from somebody, but you should not preserve the `span-id` when returning `traceparent` to your caller. You can replace it with your own `span-id`.

There are also some async scenarios possible. I think we have to standardize valid mutations, not exact semantic for every scenario.

AloisReitbauer:
Ii would really like to get started on a processing model for how trace state is handled.  I am ok to merge this one to have a consolidates state, but please mark as work in progress.

SergeyKanzhelev:
How can I mark it as WIP? Have a comment in the doc?

SergeyKanzhelev:
BTW, I started improving demo app to show how mutations work:
https://cdn.rawgit.com/w3c/distributed-tracing/sergkanz/uiForModifications/try-it.html

The idea is to have different mutation buttons. Each with it's own behavior. And there may be many steps so you can see how headers change in multiple hops.

yurishkuro:
Similar to #88, the suggestion on the call yesterday was to close this PR and move the proposed wording to an issue for further discussion, because it's premature to start documenting behavior and formats before we even agreed if this is something that should be part of the standard. E.g. many tracers and/or instrumentations may not even support context propagation in the response.

AloisReitbauer:
@SergeyKanzhelev fine with this?

Issue #81, opened 2018-03-12
Describe headers using ABNF format


codefromthecrypt:
from @AloisReitbauer 

Honestly, I find it weird to describe the header format in plain English. We should use EBNF syntax to describe it. As an example look at the grammar for the HTTP Cookie header

...
cookie-date = *delimiter date-token-list delimiter
date-token-list = date-token ( 1delimiter date-token )
date-token = 1non-delimiter

delimiter = %x09 / %x20-2F / %x3B-40 / %x5B-60 / %x7B-7E
non-delimiter = %x00-08 / %x0A-1F / DIGIT / ":" / ALPHA / %x7F-FF
non-digit = %x00-2F / %x3A-FF

day-of-month = 1*2DIGIT ( non-digit *OCTET )
month = ( "jan" / "feb" / "mar" / "apr" /
"may" / "jun" / "jul" / "aug" /
"sep" / "oct" / "nov" / "dec" ) OCTET
year = 24DIGIT ( non-digit *OCTET )
time = hms-time ( non-digit OCTET )
hms-time = time-field ":" time-field ":" time-field
time-field = 12DIGIT

AloisReitbauer:
Talked to Mark Nottingham today. He pointed me to http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html for header formats.

AloisReitbauer:
Renaming issue to use "structured headers"

AloisReitbauer:
I think the structured headers can be most easily used for Trace-State. We need to decide on which format we want to use:

* [Dictionary](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#dictionary) - Which are basically unique name/value pairs. Example: _Example-DictHeader: foo=1.23, en="Applepie", da=*w4ZibGV0w6ZydGUK*_
* [Lists](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#list) - Which is basciall an array of items _Example-StrListHeader: "foo", "bar", "It was the best of times."_
* [Parameterized Lists](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#param). Which are a combination of both approaches. (Honestly, I find them confusing)

I think for our purpose dictionary is the best approach. 

I am, however, not sure how to represent Trace-Parent

SergeyKanzhelev:
I added https://github.com/w3c/distributed-tracing/pull/96 for traceparent. 

[...]

Issue #80, opened 2018-03-12
Position zero in tracestate identifies the traceparent


codefromthecrypt:
I believe the left-most position of the `tracestate` should identify the calling trace (the one that wrote `traceparent`). There were arguments against doing this at one workshop, but it makes a lot of sense if we just make this a requirement. That's why I've done this in https://github.com/w3c/distributed-tracing/pull/72

In doing so, we remove the ability to split `tracestate` across multiple header values (as perhaps they might not be put back in order?) I think this is ok. For example, we don't define behavior of multiple `traceparent` entries, so even special casing multiple `tracestate` entries is a bit of a strange optimization especially with a length constraint of 512 bytes.

yurishkuro:
Can we list the arguments for and against that?

codefromthecrypt:
Note to whoever wants to respond to this (even if future self) this was
discussed at the last tracing workshop so notes might be there and/or
video. Possibly not.

christian-schwarzbauer:
I will start with arguments for this:
* when trying to correlate the spans, you can instantly decide if the parent span was created by your tracing system or another system
  * if yes ->  you should be able to look it up in your system
  * if not -> in this case you can skip searching in your system or you can ask the other tracing system, if possible
* when receiving such a `tracestate`, you can also immediately tell if the parent was created by your system, in case you need different handling of something in your tracing code

IMHO this behavior should make it *a lot* easier to correlate spans of traces that have been created by multiple tracing systems.

yurishkuro:
@christian-schwarzbauer couldn't the same be achieved without position-based approach? E.g.
  * receiver looks for its own vendor key in `tracestate` and parses span id
  * it then parses `traceparent` and compares span id
  * only if span IDs match it can infer that the previous hop was traced with the same vendor

SergeyKanzhelev:
few disadvantages of positioning are:
- not every trace library wants to know where it sends telemetry to. Sometimes it will be better to semantically recognize certain `tracestate` keys and read/update them if it knows what to do with them
- `traceparent` was created to optimize identities that standard "recommend" for distributed tracing correlation. So basic interoperability should be on that level without the need to always parse `tracestate`
- unnatural requirement for http. Http treats headers with comma-separated values as concatenate-able

codefromthecrypt:
Sergey your first point is less about positioning and more about opacity
right? Isnt arbitrary keys in trace state something we discussed as not
viable in the past (due to others clobbering eachothers values etc).
[...]

PR #107, opened 2018-04-13
ABNF definition for tracestate

SergeyKanzhelev:
#81

### DIFF
--- a/correlation_context/HTTP_HEADER_FORMAT.md
+++ b/correlation_context/HTTP_HEADER_FORMAT.md
@@ -14,26 +14,14 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 ## Header value
 
+`tracestate` is a Structured header in accordance to [Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html). Its value MUST be a Parameterised List ([Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html), Section 4.3).
+
 `name1=value1[;properties1],name2=value2[;properties2]`
 
 **Limits:**
 1. Maximum number of name-value pairs: `180`.
 2. Maximum number of bytes per a single name-value pair: `4096`.
 3. Maximum total length of all name-value pairs: `8192`.
-
-## Name format
-
-Url encoded string. Spaces are allows before and after the name. Header with the trimmed name and with spaces before and after name MUST be considered identical.
-
-## Value format
-
-Value starts after equal sign and ends with the special character `;`, separator `,` or end of string. Value represents a url encoded string and case sensitive. Spaces are allowed in the beginning and the end of the value. Value with spaces before and after MUST be considered identical to the trimmed value. 
-
-## Properties
-
-Properties are expected to be in a format of keys & key-value pairs `;` delimited list `;k1=v1;k2;k3=v3`. Some properties may be known to the library or platform processing the header. Such properties may effect how library or platform processes corresponding name-value pair. Properties unknown to the library or platform MUST be preserved if name and/or value wasn't modified by the library or platform.
-
-Spaces are allowed between properties and before and after equal sign. Properties with spaces MUST be considered identical to properties with all spaces trimmed.
 
 # Examples of HTTP headers
 
--- a/correlation_context/HTTP_HEADER_FORMAT.md
+++ b/correlation_context/HTTP_HEADER_FORMAT.md
@@ -14,6 +14,8 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 ## Header value
 
+TODO: This header may need it's own ABNF description. Examples do not match the RFC definition.
+
 `Correlation-Context` is a Structured header in accordance to [Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html). Its value MUST be a Parameterised List ([Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html), Section 4.3).
 
 `name1=value1[;properties1],name2=value2[;properties2]`
--- a/correlation_context/HTTP_HEADER_FORMAT.md
+++ b/correlation_context/HTTP_HEADER_FORMAT.md
@@ -14,16 +14,26 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 ## Header value
 
-TODO: This header may need it's own ABNF description. Examples do not match the RFC definition.
-
-`Correlation-Context` is a Structured header in accordance to [Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html). Its value MUST be a Parameterised List ([Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html), Section 4.3).
-
 `name1=value1[;properties1],name2=value2[;properties2]`
 
 **Limits:**
 1. Maximum number of name-value pairs: `180`.
 2. Maximum number of bytes per a single name-value pair: `4096`.
 3. Maximum total length of all name-value pairs: `8192`.
+
+## Name format
+
+Url encoded string. Spaces are allows before and after the name. Header with the trimmed name and with spaces before and after name MUST be considered identical.
+
+## Value format
+
+Value starts after equal sign and ends with the special character `;`, separator `,` or end of string. Value represents a url encoded string and case sensitive. Spaces are allowed in the beginning and the end of the value. Value with spaces before and after MUST be considered identical to the trimmed value. 
+
+## Properties
+
+Properties are expected to be in a format of keys & key-value pairs `;` delimited list `;k1=v1;k2;k3=v3`. Some properties may be known to the library or platform processing the header. Such properties may effect how library or platform processes corresponding name-value pair. Properties unknown to the library or platform MUST be preserved if name and/or value wasn't modified by the library or platform.
+
+Spaces are allowed between properties and before and after equal sign. Properties with spaces MUST be considered identical to properties with all spaces trimmed.
 
 # Examples of HTTP headers
 
--- a/correlation_context/HTTP_HEADER_FORMAT.md
+++ b/correlation_context/HTTP_HEADER_FORMAT.md
@@ -14,7 +14,7 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 ## Header value
 
-`tracestate` is a Structured header in accordance to [Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html). Its value MUST be a Parameterised List ([Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html), Section 4.3).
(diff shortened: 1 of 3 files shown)


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -132,16 +132,22 @@ base16(<TraceOptions>) = 00  // not-sampled
...
 ## Header name
 
 `tracestate`
 
 ## Header value
 
-`vendorName1=opaqueValue1,vendorName2=opaqueValue2`
+`tracestate` is a Structured header in accordance to [Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html). Its value MUST be a dictionary ([Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html), Section 4.1).
+
+Maximum length of a combined header MUST be less than 512 bytes. If the maximum length of a combined header is more than 512 bytes it SHOULD be ignored.
```
yurishkuro: how did this 512 number come up? Why 512? HTTP Cookies are 4kb

SergeyKanzhelev: https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT_RATIONALE.md#size-limits

Idea was that the size should be small enough so it will fit into messaging systems metadata for most systems. Like queues and such. It was a main concern against using the correlation context

yurishkuro: I suggest adding the reason to the spec.

SergeyKanzhelev: Added comment into the field definition. Please take a look

codefromthecrypt: That rationale is vague and distracting. Can you put a concrete place where 512 would be a limitation? even one or two is better than something that a lot of people find distractingly small.

AloisReitbauer: Should the header be ignored or truncated?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -132,16 +132,22 @@ base16(<TraceOptions>) = 00  // not-sampled
...
 
-The value a concatenation of trace graph name-state pairs. Only one entry per
-name is allowed because the entry represents that last position in the trace.
+Example: `vendorName1=opaqueValue1,vendorName2=opaqueValue2`
+
+The value a concatenation of trace graph key-value pairs. Only one entry per
+key is allowed because the entry represents that last position in the trace.
```
AloisReitbauer: This is already defined by using a dictionary

SergeyKanzhelev: you commenting to outdated text

SergeyKanzhelev: I think you missed discussion in #81

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Truncation can only be formalized with the ordering (issue #80). I intentionally make a language `SHOULD` to allow for advanced implementations.

Also see mention of it in #110. I mention that for privacy concerns you may whitelist keys. But this is highly discouraged

correlation_context/HTTP_HEADER_FORMAT.md
```diff
@@ -14,27 +14,15 @@ Multiple correlation context headers are allowed. Values can be combined in a si
 
 ## Header value
 
+`tracestate` is a Structured header in accordance to [Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html). Its value MUST be a Parameterised List ([Draft of Structured Headers for HTTP](http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html), Section 4.3).
```
nicmunroe: Should this be `Correlation-Context`, not `tracestate`?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -145,16 +145,43 @@ base16(<TraceOptions>) = 00  // not-sampled
...
+The `tracestate` HTTP header field conveys information about request position in multiple distributed tracing graphs.
+
 ## Header name
 
 `tracestate`
 
 ## Header value
 
-`vendorName1=opaqueValue1,vendorName2=opaqueValue2`
+This section uses the Augmented Backus-Naur Form (ABNF) notation of [RFC5234](https://tools.ietf.org/html/rfc5234), including the DIGIT rule from that document. It also includes the OWS rule from [RFC7230](https://www.rfc-editor.org/info/rfc7230).
```
nicmunroe: Might be good to add some quick in-line layman definitions for `DIGIT` and `OWS` - they're not particularly complicated and would save some people not steeped in specs (like me) from having to go digging around in some lengthy RFCs. 

At the very least more focused links that get the reader closer to the thing you're pointing out, like [the DIGIT rule in appendix B.1 for RFC 5234](https://tools.ietf.org/html/rfc5234#appendix-B.1) and [the OWS rule from RFC 7230 Section 3.2.3](https://tools.ietf.org/html/rfc7230#section-3.2.3) (the RFC7230 link you currently have doesn't take you directly to the spec, let alone a section in the spec - I recommend this tools.ietf.org/html link instead even if you don't deep link into the relevant section).

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -145,16 +145,43 @@ base16(<TraceOptions>) = 00  // not-sampled
...
-The value a concatenation of trace graph name-state pairs. Only one entry per
-name is allowed because the entry represents that last position in the trace.
+```
+key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+lcalpha    = %x61-7A ; a-z
+```
+
+Note that identifiers can only contain lowercase letters.
+
+Valus is opaque string up to 256 characters printable ASCII [RFC0020](https://www.rfc-editor.org/info/rfc20) characters (i.e., the range 0x20 to 0x7E) except comma `,` and `=`. Note that this also excludes tabs, newlines, carriage returns, etc.
```
nicmunroe: 👍 for the layman-readable definitions immediately surrounding the RFC-style definitions. Nicely done. I think we could use one of these layman definitions for `dictionary` and `dict-member` above. It appears to be the only piece of the puzzle missing a layman definition.

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -145,16 +145,43 @@ base16(<TraceOptions>) = 00  // not-sampled
...
+Identifiers are short (up to 256 characters) textual identifiers.
 
-The value a concatenation of trace graph name-state pairs. Only one entry per
-name is allowed because the entry represents that last position in the trace.
+```
+key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+lcalpha    = %x61-7A ; a-z
+```
+
+Note that identifiers can only contain lowercase letters.
```
nicmunroe: Recommend changing this line to: 

Note that keys must begin with a lowercase letter, and can only contain lowercase letters `a-z`, digits `0-9`, underscores `_`, dashes `-`, asterisks `*`, and forward slashes `/`.

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -145,16 +145,43 @@ base16(<TraceOptions>) = 00  // not-sampled
...
+
 ## Header name
 
 `tracestate`
 
 ## Header value
 
-`vendorName1=opaqueValue1,vendorName2=opaqueValue2`
+This section uses the Augmented Backus-Naur Form (ABNF) notation of [RFC5234](https://tools.ietf.org/html/rfc5234), including the DIGIT rule from that document. It also includes the OWS rule from [RFC7230](https://www.rfc-editor.org/info/rfc7230).
+
```
nicmunroe: I'd suggest a layman definition here before the `dictionary` RFC-style definition, maybe something like: 

>The `tracestate` header value is a `dictionary` as defined below. The `dictionary` is a series of `dict-members` separated by commas `,`, and a `dict-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `dict-member`s are ignored. There can be a maximum of 128 `dict-member`s in a `dictionary`.
>
>A simple example of a `dictionary` with two `dict-member`s might look like: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`

I know you have that example later, but it might be good to have it up front to give people some context as they're reading through the RFC-style definitions.

SergeyKanzhelev:
@nicmunroe great review! I think I addressed all the feedback.

nicmunroe:
@SergeyKanzhelev I had one last recommendation that I took too long to write out before you addressed my comments. :) I guess I should have done an actual review and submitted it all at once rather than individual comments. Sorry!

Overall I'm pretty happy with how this reads between the combination of layman definitions and RFC-style definitions that are still fairly simple.

I'm personally punting on the 512 bytes issue for now, but don't take my silence on that particular issue as consent necessarily - I'd like to see it addressed as mentioned by others. The specific 512 number does feel a little arbitrary without an example or two of systems that have a 512 limit.

nicmunroe:
Based on the discussion here: https://github.com/w3c/distributed-tracing/issues/81#issuecomment-386137859

Seems like we could slightly adjust the definition of `tracestate` to something like: `tracestate = #(dict-member)`, getting rid of the `dictionary` definition, and it solves the multiple-same-named-header issues and implies that ordering is important all at the same time? (Maybe rename `dict-member` to `ts-member` or something more relevant after getting rid of `dictionary`).

It's possible I'm misreading, and I have no idea if the `#(...)` syntax from [RFC 7230 Section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2) is valid ABNF syntax, but if this works then the definition of `tracestate` is even more succinct, and we can add some explanatory text that links to https://tools.ietf.org/html/rfc7230#section-3.2.2 and explains that multiple-same-named `tracestate` headers implies a comma separated list where ordering matters.

nicmunroe:
If we decide to go this route of explicitly supporting multi-same-named-headers to solve some of our definition and ordering issues we should consider adding some `SHOULD` language to encourage implementations to combine multi-`tracestate`-headers into a single concatenated header when doing further HTTP propagation. This would maybe help with some of the accidental breakage we might see in the wild, e.g. calling `headers.get("tracestate")` instead of `headers.getAll("tracestate")`. A good spec-implementing citizen could then concatenate into a single header in the hopes of things "just working" anyway even if downstream does the wrong thing.

If we want to explicitly *not* support multi-same-named headers (something that [RFC 7230 Section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2) implies we can do) then we probably would need to pick a different delimiter than comma. (And then we'd still probably have to define what to do if you receive a multi-header anyway since it will undoubtedly happen, like with cookies - see the commentary in appendix A.2.3. of [this doc](https://arxiv.org/pdf/cs/0105018.pdf).)

mnot:
This shouldn't reference or mention Structured Headers if it's not going to use it.

nicmunroe:
Agreed. If we end up not going with structured headers then this PR name should be changed (again).

SergeyKanzhelev:
I'm not sure what was the conclusion during workshop. Since we have ordering in specification already I adjusted the ABNF definition to use `list` instead of `dictionary`. Also it seems like everybody OK with proposed format as oppose to use of structured header.

Can you please let me know if this was discussed and whether there are still concerns with this particular PR.

AloisReitbauer (approved):

Issue #93, opened 2018-04-02
Define behaviour on how Trace-ID are to be modified. 


AloisReitbauer:

SergeyKanzhelev:
closed via #108

PR #108, opened 2018-04-13
mutating of trace state

SergeyKanzhelev:
#93

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -189,11 +189,13 @@ tracestate: rojo=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01,congo=l
 
 ## Base mutations
 
-Library or platform receiving `traceparent` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests.
+Library or platform receiving `traceparent` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests. 
+
+If the value of `traceparent` field wasn't changed before propagation - `tracestate` should also be propagated unmodified. Unmodified headers propagation is typically implemented in a pass-thru services like proxies. This behavior may also be implemented in a services which currently does not collect distributed tracing information.
 
 Here is the list of allowed mutations:
-1. **Update `span-id`**. The value of property `span-id` can be regenerated.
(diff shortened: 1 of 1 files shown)


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -184,3 +184,16 @@ Multiple tracing systems (with different formatting):
 ```
 tracestate: rojo=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01,congo=lZWRzIHRoNhcm5hbCBwbGVhc3VyZS4=
 ```
+
+# Mutating the traceparent field
+
+## Base mutations
+
+Library or platform receiving `traceparent` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests.
```
yurishkuro: Why MAY mutate instead of SHOULD?

codefromthecrypt:
>
> +
> +Library or platform receiving `traceparent` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests.
>
>
guess this is for pass-through propagation? If so, this example should be
noted I guess.

SergeyKanzhelev:
Yes, MAY was for the pass-thru. I was thinking to make it more radical and suggest that if `traceparent` was not mutated - `tracestate` shouldn't be either. One exception may be when proxy wants to tall something about itself. Like some identifier of a proxy.

Anybody knows real life scenarios when pass-thru for `traceparent` doesn't mean pass-thru for the `tracestate`?

SergeyKanzhelev:
Posted an update. @adriancole @yurishkuro appreciate feedback.

yurishkuro (commented):
I prefer to separate the players into pass-through only and participating. The former must not mutate. The later must mutate the parent.

The reason MAY bothers me is that it leaves the standard vague. A node can be screwing the overall trace and still be compliant.

SergeyKanzhelev:
@yurishkuro I added this paragraph. Is it what you think needed?

> If the value of `traceparent` field wasn't changed before propagation - `tracestate` MUST NOT be modified as well. Unmodified headers propagation is typically implemented in a pass-thru services like proxies. This behavior may also be implemented in a services which currently does not collect distributed tracing information.

SergeyKanzhelev:
@adriancole I'd appreciate sign off as I can only merge after one contributor approval.

bogdandrutu (approved):

bogdandrutu:
@SergeyKanzhelev done 👍

codefromthecrypt (approved):

Issue #55, opened 2018-02-01
Add language about abuse and risks for context and sampling flag entering/leaving organizational boundaries via public APIs, outbound calls


cce:
As discussed just now at the Distributed Tracing Workshop, it would be good to have some language in the spec around Trace-Context, Trace-Context-Ext, and Correlation-Context headers that come from callers outside your organization, or are unintentionally propagated out to HTTP servers outside of your organization's administrative control.

For example, if you have enabled tracing on a service with a public API and naïvely continue any trace with the sampling flag set using the trace ID provided, a malicious attacker (or simply any third party using the HTTP headers) could overwhelm your application with tracing overhead, forge trace ID collisions that make your monitoring data unusable, or run up your tracing bill with your SaaS tracing vendor.

Similarly, if you automatically propagate sensitive baggage values in your Correlation-Context to all HTTP requests your services make, you may be revealing sensitive information to external HTTP APIs services outside your organization's boundary. Or your vendor/implementor may reveal something about your organization in the Trace-Context-Ext header.

AWS X-Ray's docs have [a little blurb](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader) about this:
> Tracing Header Security
> A tracing header can originate from the X-Ray SDK, an AWS service, or the client request. Your application can remove X-Amzn-Trace-Id from incoming requests to avoid issues caused by users adding trace IDs or sampling decisions to their requests.

cce:
^^ @adriancole @SergeyKanzhelev @bogdandrutu As discussed in the previous session, please chime in if you have additional concerns you think should be mentioned.

mtwo:
To add on, this is a huge deal companies that provide widely used APIs (Amazon, Microsoft, Google come to mind, but it's a general problem). We've had a lot of discussions about this with respect to OpenCensus, and it's why concepts like links exist in OpenCensus, but it's worth calling out here as well.

AloisReitbauer:
In most cases a tracing system will not know that an outbound call is leaving a security boundary. This is why I consider correlation-context a potential security risk. 

For incoming headers there MUST be a way for a trace to ignore sampling decisions etc. We need to define how this is handled.

SergeyKanzhelev:
Closed via #157

PR #110, opened 2018-04-17
privacy considerations

SergeyKanzhelev:
Issues cleaning continued: adding some privacy notes for #55. I'll tag it that issue with `correlation-context` tag once this PR is complete. So it will not be in the main query


wu-sheng (commented):
> Platforms and tracing systems MUST NOT include any personal identifieable information into `tracestate` header.

I think OpenTracing  should be noticed, the `baggage` concepts will have some conflicts with the Privacy rules, if some supported tracers support both this spec and OT.

FYI @tedsuo

SergeyKanzhelev:
>> Platforms and tracing systems MUST NOT include any personal identifieable information into `tracestate` header.

>I think OpenTracing  should be noticed, the `baggage` concepts will have some conflicts with the Privacy rules, if some supported tracers support both this spec and OT.

I added privacy note to make sure any service may store and propagate `tracestate` without much privacy concerns. So the idea is ONLY store non-identifiable information there like application ID, sampling and such. For baggage `Correlation-Context` header should be used.

trace_context/compliance.md
```diff
@@ -0,0 +1,19 @@
+# Privacy
+
+Requirements to propagate headers to downstream services opens a potential privacy concerns. The only way to remove this concern is to inspect and remove values from the fields values before allowing the platform or tracing system to execute code that potentially can propagate these headers. All mutations should, however, conform to the list of mutations defined in this specification.
+
+## Privacy of traceparent field
```
bogdandrutu: Not necessary a privacy concern, more like an usability concern is when the service initiating calls use the same trace-parent every time then this cause the service owner to not be able to distinguish between different calls. this my be solved by restarting the trace on the service side if the client is untrusted.

SergeyKanzhelev: Yes. I'd start security section for this. This doc purpose was to ensure the `traceparent` and `tracestate` can be stored and transmitted without implications for privacy and things like GDPR.

CodingFabian:
👍 from my side. It would be a good idea to explicitly mention that there must be no storage restrictions on the tracestate value.

SergeyKanzhelev:
@CodingFabian I made it more explicit

tedsuo:
@wu-sheng I think baggage is a separate issue, related to `correlation-context`, not `tracestate`.

LGTM, except fot the bit about mutating the tracestate. 

If the suggestion is that `systems MUST NOT include any personal identifieable information into tracestate header`, then the follow up suggestion that `systems extremely sensitive for personal information exposure MAY implement selective removal of values corresponded to the unknown keys` doesn't really make sense. 

"Extremely sensitive systems" should be following the main privacy directive, and not be putting "unkown keys" into their `tracestate` to begin with. (I presume the suggestion here is **not** to mess with _other systems_ tracestate, but with previous instances of *your* tracestate). 

Simply saying "don't put any personally identifiable information in tracestate" should be enough for now, the nuance muddles the message IMHO.

mtwo (approved):

SergeyKanzhelev:
@tedsuo I realize there are some re-iteration of the same. I intentionally mentioned it as there are services who will want to have an extra level of protection and will want to blacklist. Not talking theoretically =)

The language can be cleaned up though and I'd propose we clean it with upcoming editors clean up. 

Thank you for feedback. If you still believe the language needs to change factually - please file an issue. Haven't meant to disregard your feedback here

Issue #109, opened 2018-04-16
Explain/remove 512 limit of tracestate in spec


SergeyKanzhelev:
We have explanation of 512 limit in rationale file, but not in spec. Suggestion was [made](https://github.com/w3c/distributed-tracing/pull/107) to put it into spec.

SergeyKanzhelev:
We explained in PR why we need limit. We still haven't explained why the limit is 512

SergeyKanzhelev:
So far asking different providers - everybody wants smaller limit, but have not upper-bound physical limitation that will prevent saving and propagation of 512 characters.

So we need to put rationaly of why we need 512. If you implemented tracestate - can you post example/size you need?

SergeyKanzhelev:
For example, B3 encoding will require 72 characters `tracestate: b3=80f198ee56343ba864fe8b2a57d3eff7-05e3ac9a4f6e3b90-1-e457b5a2e4d86bd1` (see: https://github.com/openzipkin/b3-propagation/issues/21).

yurishkuro:
I feel that any limit number is going to be largely arbitrary. We can instead clearly outline procedure for trimming the state and not specify a hard limit, or use an external one like limit for cookies.

SergeyKanzhelev:
One thing I think we can avoid by defining some hard limit is abuse of the header to propagate user-defined baggage. If you know that it will be cut anyway - you'd look for solution like `Correlation-Context` from the beginning instead of frustration when context will not be propagated,

I like that idea though. Definitely good idea to detail the trimming procedure

yurishkuro:
I just think the exact limit can be very application specific, e.g. an application that exchanges many small messages will be more sensitive to the header size than one that carries 20Kb JSON requests. So the exact limit might be configurable per application, but the spec's hard limit should be more generous, like the cookie size limit of 4093.

Let the market forces pressure vendors to keep the size to a minimum.

discostu105:
Here is how the Dynatrace tracestate will _likely_ (typically) look like:

`fw5-29a-3039@dt=1;FEC354CD6;1;2;727c` (36 bytes)

However, the elements are of variable length, so they could be longer. The theoretical maximum would be like this:

`fw5-ffffffff-ffffffff@dt=ffffffff;ffffffffffffffff;ffffffff;ffffffff;ffffffff` (77 bytes)

SergeyKanzhelev:
Closing in favor of https://github.com/w3c/trace-context/issues/239

PR #111, opened 2018-04-17
explain aggressive limit

SergeyKanzhelev:
#109

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -155,9 +155,11 @@ Rather, the entry would be rewritten to only include the most recent position:
 `congo=congosSecondPosition,rojo=rojosFirstPosition`
 
 **Limits:**
-Maximum length of a combined header MUST be less than 512 bytes.
+There might be multiple `tracestate` headers in a single request according to [RFC 7230 section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2). Maximum length of a combined header MUST be less than 512 characters. This length include commas required to separate list items. But SHOULD NOT include optional white space (OWA) characters.
 
-`tracestate` field contains essential information for requests correlation. Platforms and tracing systems MUST propagate this header. Compliance with specification will require storing of `tracestate` as part of request payload or associated metadata. Allowing the long field values can make compliance to the specification impossible. Thus the aggressive limit of 512 bytes was chosen.
+`tracestate` field contains essential information for requests correlation. Platforms and tracing systems MUST propagate this header. Compliance with specification will require storing of `tracestate` as part of request payload or associated metadata. Allowing the long field values can make compliance to the specification impossible. Thus the aggressive limit of 512 characters was chosen.
+
+If the `tracestate` value size is bigger than 512 characters, the tracer CAN decide to forward the `tracestate`. When propagating `tracestate` with the excessive length - the assumption SHOULD be that the callee will drop this header.
 
 ## Name format
 
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -155,7 +155,9 @@ Rather, the entry would be rewritten to only include the most recent position:
 `congo=congosSecondPosition,rojo=rojosFirstPosition`
 
 **Limits:**
-Maximum length of a combined header MUST be less than 512 bytes. 
+Maximum length of a combined header MUST be less than 512 bytes.
+
+`tracestate` field contains essential information for requests correlation. Platforms and tracing systems MUST propagate this header. Compliance with specification will require storing of `tracestate` as part of request payload or associated metadata. Allowing the long field values can make compliance to the specification impossible. Thus the aggressive limit of 512 bytes was chosen.
 
 ## Name format
 


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -155,7 +155,9 @@ Rather, the entry would be rewritten to only include the most recent position:
 `congo=congosSecondPosition,rojo=rojosFirstPosition`
 
 **Limits:**
-Maximum length of a combined header MUST be less than 512 bytes. 
+Maximum length of a combined header MUST be less than 512 bytes.
+
+`tracestate` field contains essential information for requests correlation. Platforms and tracing systems MUST propagate this header. Compliance with specification will require storing of `tracestate` as part of request payload or associated metadata. Allowing the long field values can make compliance to the specification impossible. Thus the aggressive limit of 512 bytes was chosen.
```
codefromthecrypt: I still think a more concrete example of why 512 vs 2048 or any other fixed val

codefromthecrypt: @codingfabian @abhiksingh can you help explain this? I think you have a lot of experience with systems, maybe can help shed some light. Right now, feels just like a couple people saying things again and again.. we need better than this.

In http most servers are subject to a total limit of header size, not individual. Other parts of the spec later will overrun completely towards the same budget. We can assume that this size is indeed only about this part and not directly about headers in general.

I think guidance should clearly be for vendors to use the least space practical. However, where we put a hard limit, such a line is important, as it affects the viability of multiple vendors at all. IOTW, it shouldn't be cripplingly small, neither should it arbitrarily and early punt towards continuations (like a second header with the same name) as that just moves the problem in ways that only work in systems with multi-value headers.

So anyway, any concrete examples to help pin down what I still feel is a bit arbitrary?

SergeyKanzhelev: The history of 512 was that it was a number nobody wanted to go below. I heard another suggestion is to haver total of `tracestate` and `traceparent` be under 1K.

CC: @bogdandrutu who has opinion and example with some queue service and metadata.

AloisReitbauer: I would rephrase this part to be a bit more open. I suggest we define this as:

Tracestate must be forwared by all participating tracing if it i is less or equal than 512 bytes. If the tracestate is bigger than this, the tracer can decide to forward the trace state or drop it.

Note: I know we talked about truncation. However, since there is no final decision on how this would work, the current working draft should state that you either forward or drop.

CodingFabian: 👍 on the wording by @AloisReitbauer

tedsuo:
I believe the goal of 512k was to ensure that 3 hops (the maximum expected in most real world scenarios) would not cause this header to be dropped or truncated. If that is the motivation, it should be listed here. Currently this explains that an overflow could occur, but not why the limit is 512k vs something else.

If such an explanation is better put in the Rationale document instead, ignore this.

mtwo (approved):

codefromthecrypt:
512 bytes you meant :P

agree it is still arbitrary and would like to see concretely why not 1024 or something larger. This is the very best time to answer this question rather than people years later looking for the answer and never finding it. If we can't answer the question we shouldn't have this constraint.

jkowall:
I know we will overrun this size... I'm not sure what the number is but 1024 seems small to me. 512 is a big issue.

yurishkuro:
This is easily half a dozen times I've seen this question asked: why 512? Where did it come from? Did someone just randomly picked it?

SergeyKanzhelev:
@adriancole even if there will be no meaningful explanation we need to pick a limit. One of the purpose of this limit is to make sure that `tracestate` is not abused for a baggage. It is critical for services to propagate it and ideally store with produced telemetry without implication of cost or reliability attacks. So it's better if nobody assumes "no limit" from the first iteration. People years later will think we did it wrong anyway =).

(not advocating for complete random here so let's discuss if there are implications of `512` to be a limit).

@jkowall what data do you plan to store in `tracestate` to overrun the limit? Note, `tracestate` is intended for information crucial for positioning of the span in a distributed call graph. Baggage is propagated via separate header.

@yurishkuro - yes, it was mostly random pick. Mostly it was coming from discussion that `tracestate` when implemented by cloud services - should be stored and propagated thru metadata. Ideally free of charge and not affecting performance.

codefromthecrypt:
"People years later will think we did it wrong anyway =)." lol true

SergeyKanzhelev:
BTW, I realize this PR doen't close the corresponding issue. It merely explains why value needs to have an aggressive limit. Not the reasoning for `512`.

jkowall:
@SergeyKanzhelev I'm just saying out header can get quite large, and may overrun this number. The data in the header is crucial for us to place the span in context of the application, business transaction, and other data we require. Some of that could be moved to other headers but then it may not be propagated making it useless for downstream correlation.

wu-sheng:
> @SergeyKanzhelev I'm just saying out header can get quite large, and may overrun this number. The data in the header is crucial for us to place the span in context of the application, business transaction, and other data we require. Some of that could be moved to other headers but then it may not be propagated making it useless for downstream correlation.

Saw and known some header of AppD, those must be useful for backend analysis. SkyWalking depends on those headers too. Just not so long.

I think @SergeyKanzhelev hope to move those headers to the third type of header. But as same as SkyWalking, I guess the backends need those long headers. Those are not optional. :)

SergeyKanzhelev:
@wu-sheng I understand you are talking about hierarchical identifier, correct? Do you have an acceptable size limit in mind to satisfy the requirement of keeping `tracestate` small to guarantee propagation by cloud vendors?

wu-sheng:
>  I understand you are talking about hierarchical identifier, correct? 

In AppD, more like hierarchical path, or trace path. At least last time I read the headers.  Of course, @jkowall is the authority :)

I have no good suggestions... sorry. I am just aware a lot of commercial APMs put a lot of info in headers.

codefromthecrypt:
here's a point of reference. Amazon will restart its trace ID if the header
gets to 7KiB

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html

PR #114, opened 2018-04-19
mutate of tracestate

SergeyKanzhelev:
#80


codefromthecrypt:
looks good. there might be some ways to clean up what is different between
the 1st and 3rd points but as you can imagine I am happy to see this change.

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -189,13 +189,24 @@ tracestate: rojo=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01,congo=l
...
+
+# Mutating the tracestate field
+
+Library or platform receiving `tracestate` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests. The main concept of `tracestate` mutations is that order of unmodified key-value pairs should be preserved. Modified keys should be moved to the beginning of the list.
+
+Here is the list of allowed mutations:
+
+1. **Update key value**. The value of any key can be updated. Modified key SHOULD be moved to the beginning of the list. This is the most common mutation.
+2. **Delete the key-value pair**. Any key-value pair MAY be deleted. It is highly discouraged to delete keys that wasn't generated by the same tracing system or platform. Deletion of unknown key-value pairs will break correlation in other system.
+3. **Make key-value pair first**. Tracing system MAY move any key-value pair to the first position. This mutation will indicate to the downstream which tracing system or platform sent this request.
```
codefromthecrypt: Why would you move something and not update it?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -189,13 +189,24 @@ tracestate: rojo=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01,congo=l
...
\ No newline at end of file
+Libraries and platforms MUST NOT make any other mutations to the `traceparent` header.
+
+# Mutating the tracestate field
+
+Library or platform receiving `tracestate` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests. The main concept of `tracestate` mutations is that order of unmodified key-value pairs should be preserved. Modified keys should be moved to the beginning of the list.
+
+Here is the list of allowed mutations:
+
+1. **Update key value**. The value of any key can be updated. Modified key SHOULD be moved to the beginning of the list. This is the most common mutation.
```
codefromthecrypt: SHOULD seems loose as it means we can't rely on people following behaviors like this

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: Enables tracing systems that want to be generic and do not want to duplicate entire `traceparent` inside their `key`.

SergeyKanzhelev:
@adriancole I changed the language. Let's see if this will affect any implementations.

I'd appreciate sign off if there are no more comments.

codefromthecrypt:
>
> +3. **Make key-value pair first**. Tracing system MAY move any key-value pair to the first position. This mutation will indicate to the downstream which tracing system or platform sent this request.
>
> Enables tracing systems that want to be generic and do not want to
> duplicate entire traceparent inside their key.
>
how would that work? if generic you are updating your span ID, so modifying
it when you move it.

In that case how is it different than 1?

codefromthecrypt:
who is implementing this?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -189,13 +189,25 @@ tracestate: rojo=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01,congo=l
...
+
+# Mutating the tracestate field
+
+Library or platform receiving `tracestate` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests. The main concept of `tracestate` mutations is that order of unmodified key-value pairs SHOULD be preserved. Modified keys SHOULD be moved to the beginning of the list.
+
+Single request may be processed by multiple tracing systems in a single component. So any number of allowed mutations may happen in a single component. Here is the list of allowed mutations:
+
+1. **Update key value**. The value of any key can be updated. Modified key MUST be moved to the beginning of the list. This is the most common mutation.
+2. **Delete the key-value pair**. Any key-value pair MAY be deleted. It is highly discouraged to delete keys that wasn't generated by the same tracing system or platform. Deletion of unknown key-value pairs will break correlation in other system.
+3. **Make key-value pair first**. Tracing system MAY move any key-value pair to the first position. This mutation will indicate to the downstream which tracing system or platform sent this request.
```
codefromthecrypt: based on implementation I would suggest this:

1. Insert trace state entry <- you are starting your trace in beginning position, regardless of if there are other state entries or not
2. Update trace state entry < you are resuming a trace, by modifying your state, and moving it to position zero again
2. Delete trace state entry <- no idea why we document this? This is more like truncate

FYI: Please do not rush this to merge. I am pretty sure no-one is implementing this besides tyler and me so listen instead of rush language that conflicts with how things work

SergeyKanzhelev: > Delete trace state entry <- no idea why we document this? This is more like truncate

Mostly for [privacy](https://github.com/w3c/distributed-tracing/pull/110). And to enable potential truncation. I'm calling it explicitly so one will not be out of compliance by implementing the truncation or privacy sweep.

For other items - are you suggesting change of name for mutations? Or your feedback that you do not see the reason for unmodified mutation?

codefromthecrypt:
>
> Mostly for privacy <https://github.com/w3c/distributed-tracing/pull/110>.
> And to enable potential truncation. I'm calling it explicitly so one will
> not be out of compliance by implementing the truncation or privacy sweep.
>
> ok so here maybe change text for privacy as I don't understand how you
would know other trace state you don't own are a privacy risk. not to say
others don't understand this just if it is mostly for privacy, say that in
the text?

For other items - are you suggesting change of name for mutations? Or your
> feedback that you do not see the reason for unmodified mutation?
>
mostly I don't understand why you would reorganize entries you don't
modify. that is the main thing. in doing so it destroys the ordering
assumption doesn't it? This is the part that regardless of how clearly
worded seems most suspect as I believe when participating in a trace we
already say you modify things.

The section titles mentioned for example resume your trace by moving it
front can be special cases later if we define an entry can be without a
value. the overall logic would be held in tact.

tldr I think the modifications to headings.. three primitives of create
state entry, modify (both imply moving to first position) and delete.

Basically I can't currently see any reason to have arbitrary reorder as
something supported. Maybe you can highlight why.. if you prefer just
remove this feature and make it a separate PR as this is the sticky part
for me.

>

SergeyKanzhelev:
Updated. I'm ok to take unmodified mutation out for now.

codefromthecrypt:
as far as I can tell, the primary purpose of this wording is to facilitate
one hop making multiple simultaneous state changes. it does that while
complicating normal state change where only one state entry is changed by a
party (as opposed to multiple as suggested here presumably to facilitate MS
and N formats)

I don't think it is good to describe this advanced case at the same time as
describing normal single trace on one hop. for example feedback on
complexity from nic had us remove even special casing where we removed
special casing a trace parent which was the same format as trace state.

All that said, personally I don't want to argue about it and it saddens me
to know others either dont find this too complex or don't care enough to
say it out loud.

Regardless, I would encourage decoupling these but really want someone
implementing who isn't me to make such a suggestion so that when I am not
around things around simplicity can remain priority.

yurishkuro:
I am still not convinced that mutating multiple keys is a valid use case. I would like to have a separate issue where that use case is presented with concrete, unambiguous, and actually practical examples, like
  * tracestate: k1=blob1, k2=blob2
  * where blob1=v1/v2/v3
  * where blob2=w1/w2/w3
  * and some node X in the graph MUST mutate both blob1 and blob2 by replacing vN or wN simultaneously and it absolutely cannot work othewwise
  * and instead of using abstract v/w names the example uses concrete field names with explained semantics

If such example cannot be provided, we can use much simpler mutation rules for tracestate. I would also suggest, in that case, to add a rule that if a node is adding a new key to tracestate, it can _only add a single key_ that represents its own state as opaque blob.

codefromthecrypt:
So, I think @yurishkuro and I are on the same page, being conservative about the features.

What if we re-purpose this PR with more strict definitions as "trace mutations" in essence could cover that.

In other words, go with the 3 mutations allowed (insert new, mutate existing, delete to truncate), strictly defined to cover only one mutation per change in traceparent.

This would in essence reverse document the examples already in the text. If you would like help with that, I can provide some text. The other part about multi-mutation per change in traceparent could be elaborated in a different issue and changes composed *over* text about how to deal with single-state vs conflated in the same change.

sg?

SergeyKanzhelev:
I'm fine elaborating multiple mutations in a separate issue. Removed the note about multiple mutations

yurishkuro:
> The main concept of `tracestate` mutations is that order of unmodified key-value pairs SHOULD be preserved. Modified keys SHOULD be moved to the beginning of the list.

Maybe we need to wait to resolve the new issue you booked (#117). The above wording doesn't make things clearer because it still allows mutating multiple keys, so the second sentence is actually fairly meaningless, because if more than one key is moved to the head, then their relative order is undefined, which means it breaks the main assumption that a node can rely on `entry[0]` having a significance.

SergeyKanzhelev:
hm... this `SHOULD` is left over from when there were a mutation with the unmodified key move. That we agreed to remove for now. 

Multiple mutations will make significance of `entry[0]` to be less. But you know that first few items were definitely modified recently.

I changed the language anyway for now to keep allowed mutations and overview of mutations in sync. Good catch.

codefromthecrypt (approved):

PR #125, opened 2018-05-15
Fix example of how to parse the trace options

felixbarny:
I could totally be wrong, but I think there is a bug in the example code how to extract the traced flag from the trace options bit field, as `1 << 1` is `00000010` and not `00000001`.


bogdandrutu (approved):
I would wait for others to comment, but this seems like a bug to me.

SergeyKanzhelev (approved):

PR #131, opened 2018-07-10
fix ABNF

reyang:
1. fix off-by-one
2. avoid ambiguous grammar - SP (+U0020) could be treated as trailing space of list-member OR an OWS before COMMA
3. change the wording to cover combined header size = 512 case


reyang:
@SergeyKanzhelev, I will create a few more pull requests for the spec improvement, this is the first one.

SergeyKanzhelev (approved):

PR #134, opened 2018-07-17
Reduce the total number of elements in the tracestate list.

bogdandrutu:
Based on a comment in https://github.com/census-instrumentation/opencensus-java/pull/1300 the total number of elements in the tracestate is too high. Here is the rational:
* 512 (total size) / 128 = 4 (characters per entry).
* Out of this 4 characters 2 are delimiters (= and ,) for each entry (except the first or the last one which has only 1).
* On average an entry has only 2 characters 1 for key and 1 for value. Based on the restriction of the key characters it is impossible to have 128 different keys of length 1.
* Decision to limit the number of entries to 32 is based on the total size, so every entry has on average 14 characters (16 - 2 delimiters).

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -167,13 +167,13 @@ This section uses the Augmented Backus-Naur Form (ABNF) notation of [RFC5234](ht
 
 The `OWS` rule defines an optional whitespace. It is used where zero or more whitespace characters might appear. When it is preferred to improve readability - a sender SHOULD generate the optional whitespace as a single space; otherwise, a sender SHOULD NOT generate optional whitespace. See details in corresponding RFC.
 
-The `tracestate` field value is a `list` as defined below. The `list` is a series of `list-members` separated by commas `,`, and a `list-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `list-member`s are ignored. There can be a maximum of 128 `list-member`s in a `list`.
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
Can you put this description in rationale doc, please? Those calculations are based on `512` limit which is also wasn't explained yet. But it is fine

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -167,7 +167,7 @@ This section uses the Augmented Backus-Naur Form (ABNF) notation of [RFC5234](ht
 
 The `OWS` rule defines an optional whitespace. It is used where zero or more whitespace characters might appear. When it is preferred to improve readability - a sender SHOULD generate the optional whitespace as a single space; otherwise, a sender SHOULD NOT generate optional whitespace. See details in corresponding RFC.
 
-The `tracestate` field value is a `list` as defined below. The `list` is a series of `list-members` separated by commas `,`, and a `list-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `list-member`s are ignored. There can be a maximum of 128 `list-member`s in a `list`.
+The `tracestate` field value is a `list` as defined below. The `list` is a series of `list-members` separated by commas `,`, and a `list-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `list-member`s are ignored. There can be a maximum of 32 `list-member`s in a `list`.
```
SergeyKanzhelev, line 170: Also update BNF rule below: `list  = list-member 0*127( OWS "," OWS list-member )` to `list  = list-member 0*31( OWS "," OWS list-member )`

bogdandrutu: Done.

bogdandrutu:
Done.

SergeyKanzhelev (approved):

codefromthecrypt:
sgtm

PR #135, opened 2018-07-23
fix ABNF

reyang:


SergeyKanzhelev (approved):

codefromthecrypt:
trying to understand why this is allowing 255 spaces before a non-space.

#188 has a lot of talk about impact of presumably having a trailing empty member. I saw no code references, just talk about test cases invisible to open source, yet used to justify that change. Regardless, this literally permits 255 empty spaces than a non-space. If what's said in 188 is true, and that libraries arbitrarily add white (please cite exact libraries and versions).. what does this help?

I think it just makes sure space is only allowed for left-padding a non-space. If correct, the description should be updated. If incorrect, I guess I am misreading the abnf and would appreciate knowing how.

dyladan:
I read the ABNF the same way you do and I share your confusion. With no PR description, just "fix ABNF" as a commit message, and a single approval with no comments it is tough to divine the reasoning here.

PR #139, opened 2018-08-06
Add instana as known implementation

SergeyKanzhelev:


AloisReitbauer (commented):
We should also add the other vendors who have an implementations

SergeyKanzhelev:
@AloisReitbauer as long as there will be implementation in main branch we should definitely do it. I'd suggest to start the list.

bripkens:
The implementation doesn't contain any Instana specifics. IMHO we should have two lists:

1. Libraries that are useful when implementing TraceContext support (we all need a parser, so why not share it?). This is intended for implementers.
2. Tools / Vendors supporting TraceContext. This might come in handy for users.

SergeyKanzhelev:
Decision was to only add implementation that conform to test suite

Issue #54, opened 2018-01-23
Define behavior when library do not understand the version number


SergeyKanzhelev:
The behavior may differ between not being able to parse header and unknown version number. Spec should make some considerations for the versions revision in future.

SergeyKanzhelev:
I was thinking about this one. I feel we can make a good use of `tracestate` here. So if one sees unknown version prefix - behavior may be the following:

- Pass-thru the `traceparent`
- Start a new trace. Let's say it's `00`. Place it into `tracestate` like:

Incoming:

```
traceparent: 02-abcdefg
tracestate: vendor1=opaqueValue1
```

Outgoing (from the system that only knows about `00`):

```
traceparent: 02-abcdefg                                        //unchanged
tracestate: v00=00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01, vendor2=opaqueValue2, vendor1=opaqueValue1
```

This will require:
1. Reserver `vXX` keys in `tracestate`
2. Recommendation: define the size limit as a sum of `tracestate` and `traceparent`. So size-critical systems will be able to deal with higher version of a header if it will be longer. It will just eat up some `tracestate` space
3. Parsing will be a bit more involved for the cases with the unknown version of `traceparent`. Now you'd need to look up `v00` header in case it was already started before.

SergeyKanzhelev:
@yurishkuro since you seems to be around - do you have an opinion here? Does this approach looks reasonable for you?

yurishkuro:
Sorry, I don't have an opinion about this. I'd like to see first the whole thing working even with a single version and understand all the edge cases before prescribing behavior for mismatched versions.

On the first glance, adding `v00=` seems unnecessary, since the next hop will received v0 `traceparent` and its own state may or may not be found in the tracestate.

I think downgrading the version is reasonable and should be allowed, while upgrading the version should not be allowed. Aside from that I don't know if we need any other prescriptions.

SergeyKanzhelev:
>I think downgrading the version is reasonable and should be allowed, while upgrading the version should not be allowed.

Can you elaborate on this? My thinking was quite opposite. If downgrading will be the default behavior than it will be very hard to move to the next version. Any component that is stubborn to upgrade will break features of the next version.

yurishkuro:
[...]

PR #141, opened 2018-08-06
versioning proposal during the implemetors workshop

SergeyKanzhelev:
#54


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -147,7 +150,22 @@ base16(<SpanId>) = 00f067aa0ba902b7
...
+## Versioning
+
+Implementation is opinionated about future version of specification. We hope to make format of `traceparent` additive.
+
+Implementation should follow a following rules when parsing headers with unexpected format:
+
+1. Pass thru services should not analyze version. Pass thru service needs to expect that headers may have bigger limits in future.
+2. When version prefix cannot be parsed (it's not 2 hex characters followed by dash), implementation should restart trace.
+3. If higher version is detected - implementation should try to parse.
+  a. If size of header shorter - implementation should not parse it.
```
yurishkuro: shorter than what, a combined length of 2+32+16+2+(separators)?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -147,7 +150,22 @@ base16(<SpanId>) = 00f067aa0ba902b7
...
+  b. Try parse `trace-id`: from dash 32 characters. Make sure followed by dash.
+  c. Try parse `span-id`: from the second dash 16 characters. Make sure followed by dash.
+  d. Try parse sampling bit of `flags`:  2 characters from third dash. Following with either end of string or a dash.
+  If all three values were parsed successfully - implementation should use them. After parsing - implementation MUST NOT assume anything about other parts of `traceparent` and MUST send version `00` to downstream services.
```
yurishkuro: why 00? Maybe "send the version that application understands"? What about the potential future fields after the flags - keep them or drop them? (if sending earlier version then have to drop them)

codefromthecrypt:
any references here? who proposed it who agreed or didn't. I think the version was proposed before so a summary of that (why it was removed last time.. iirc @mnot said something like "you usually have only one go at a spec like this" or something.

I do think grpc propagation should be cited as that includes version in exact same way I think.

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: I put a note that it is for the current specification

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: yes. Made it explicit

SergeyKanzhelev:
@adriancole we are not throwing away [our shot](https://www.youtube.com/watch?v=Ic7NqP_YGlg) =)

I've put more details in rationale. Basically notes from the meeting. Sorry for incomplete PR. Please comment if rationale section is not clear on thought process (or you think it's faulty).

codefromthecrypt:
I don't like that this is squishy, for example, you can't even assume a constrained character set anymore. I'm not sure why for example we are permitting non-hex additions to what is supposed to be the section for export to other consumers. Remember one of the use cases here is for logging, when we start doing things like this it amplifies the concern not just to tracer code but also observers such as correlation functions. By allowing things to change mid-stream, it will make that code harder to write for speculative reasons.

For example, we are pushing for this extensibility and I can't seem to find the use case which mandate it? So, maybe we are focusing on how to extend it with a bit less on why and which ways?

Anyway there must be some other folks to consider on topics like this. For example, only yuri is site owner, others are service providers. Can you enumerate the use cases this solves for you? Can you mention sites who currently have custom stuff and how this helps them? Or why are we revisiting versioning again? (again not the *how* but they why)

SergeyKanzhelev:
Spec doesn’t say what to do when you see unknown version. It just says the current version is 00. So this is not a re visiting. It’s just clarification. Out of three behaviors third one helps the best to move versions forward easy. 

So the use case is defining behavior we can rely on when future us talking about the next version of spec. 

Btw, we still require hex. That note is saying to future us to change position of dashes if we want parsing to fail. Simply forcing non hex character will not necessarily break parsing as we expect some implementations to take substrings and not check for every character for perf reasons.

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -147,7 +150,23 @@ base16(<SpanId>) = 00f067aa0ba902b7
...
+
+1. Pass thru services should not analyze version. Pass thru service needs to expect that headers may have bigger size limits in future and only disallow prohibitively large headers.
+2. When version prefix cannot be parsed (it's not 2 hex characters followed by dash (`-`)), implementation should restart the trace.
+3. If higher version is detected - implementation should try to parse.
+  a. If size of header shorter than 55 characters - implementation should not parse header and should restart the trace.
+  b. Try parse `trace-id`: from the first dash - next 32 characters. Make sure they followed by dash.
+  c. Try parse `span-id`: from the second dash at 35th position - 16 characters. Make sure followed by dash.
+  d. Try parse sampling bit of `flags`:  2 characters from third dash. Following with either end of string or a dash.
+  If all three values were parsed successfully - implementation should use them. After parsing - implementation MUST NOT assume anything about other parts of `traceparent` and MUST send the version `00` (in the current version of specification) to downstream services.
```
codefromthecrypt: if I understand correctly, this should send the highest version it understands to downstream services, correct? If we  word like this we don't need to write out `00` correct and it will stay true for version `01` as well version `ae`

trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -61,3 +63,21 @@ Lowercase names has a few benefits:
...
+Versioning options are:
+
+1. Pass thru unknown headers
+2. Re-start trace when you see unknown header
+3. Try to parse trace following some rules when you see unknown header
+
+One variation is whether original or new header you cannot recognize is preserved in `tracestate`.
+
+- Option 1 is least favorable as it makes one bad header break entire distributed trace.
+- Option 2 is better. It's easy, doesn't restrict future version in any way and re-started trace should be understood by new systems. So only one "connection" is lost. And the lost connection issue can be solved by storing original header in `tracestate`. Drawbacks are also obvious. First, single old component always breaks traces. Second, it's harder to transition without customer dissat of broken traces.
```
codefromthecrypt, line 78: s/dissat/dissatisfaction/

trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -61,3 +63,21 @@ Lowercase names has a few benefits:
...
+3. Try to parse trace following some rules when you see unknown header
+
+One variation is whether original or new header you cannot recognize is preserved in `tracestate`.
+
+- Option 1 is least favorable as it makes one bad header break entire distributed trace.
+- Option 2 is better. It's easy, doesn't restrict future version in any way and re-started trace should be understood by new systems. So only one "connection" is lost. And the lost connection issue can be solved by storing original header in `tracestate`. Drawbacks are also obvious. First, single old component always breaks traces. Second, it's harder to transition without customer dissat of broken traces. 
+
+  Storing original value also has negative effects. Valid `traceparent` is 55 characters (out of 512 allowed for `tracestate`). And "bad" headers could be much longer pushing valuable `tracestate` pairs out. Also this requirement increases the chance of abuse. When bad actor will start sending header with the version `99` that is only understood by that actor. And the fact that every system passes thru original value allows this actor to build complete solution based on this header.
```
codefromthecrypt, line 80: sorry maybe I missed something.. are you implying the data in 512 <= traceparent.length + tracestate.length? Ex in the variable length traceparent (which versioning implies how it is defined now) traceparent grows at the expense of tracestate?

trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -61,3 +63,21 @@ Lowercase names has a few benefits:
...
+- Option 3 with the fallback to option 2 seems to allow the easiest transition between versions by forcing a lot of restrictions on the future. Initial proposal was to try to parse individual parts like `trace-id` than `span-id`. Assuming `span-id` size or format may change without changing `trace-id`. However majority sees potential for abuse here. So we suggest to force future versions be additive to the current format. And if parsing fails at any stage - simply restart the trace.
+
+Another consideration - whether to force validation of hex numbers as part of a spec. When parsing unknown versions we may or may not require hex validation in addition to size checks and checks of dashes (`-`) in proper location. Since we anticipate many implementation of loose check for `trace-id` and `span-id` - future version of standard should assume current implementations didn't validate whether values are hex. So future version of spec cannot force incompatible version by non-hex character on the same places where hex characters are required now. Only by length of parts.
```
codefromthecrypt: The discussion about validation seems to have got me distracted. Can it be simplified to focus less on character set (which distracted me as I didn't expect us to change that anyway), and more on the feature people rely on (offset-based parsing).

This spec supports offset-based parsing of data. For example, some implementations look for where hyphens should be (delimiting the trace and span ID) as a quick check. To continue supporting this, future versions cannot change the order of the first 55 characters. It is worth repeating that future versions will also not change the allowed characters either (lower hex or hyphen).

AloisReitbauer: We discussed validation. Agreed on the comment that we need to find a way to ensure the headers can be parsed and not just everything can be sent. 

We will avoid sending "junk" as this would be truncated at the next hop

AloisReitbauer (approved):

Issue #122, opened 2018-05-03
Do we need a sampling bit in trace-parent?


AloisReitbauer:
**Issues**
* Most APM vendors will ignore the sampling bit.
* Sampling will drive costs with cloud provider.
* Semantics are not clearly defined. What should you do? Can you change it?
* How to handle system that do not upfront sampling?
* We cannot *force* third parties to trace a request. We can only *recommend* it. 
* Forcing downstream tracing might result in a DDoS attack.

**The problem to be solved**
Systems **cannot** be forced to trace a request. However, as different systems use different sampling rates we want to sync on important requests to increase the probability to sample the same trace and thus improve the probability for an end-to-end trace.

**Proposal**
Have a tracing bit be a _hint_ from the parent. The flag is supposed to have multiple values ordered in the following priority.

* Traced - Last hop of request is traced. (1)
* Undecided - Tracing data collection decision is deferred.  (2)
* Not traced - Last hop of request is not traced. (0)

Reasons why traces can be undecided:
* Injecting IDs in the first hop but not collecting a trace and not making a sampling decision
* A load balancer waiting for a request to return and deciding on whether to keep a trace.

We decided to use these values as they map with current implementations. 

**Beyond**
We also need a way to define that a certain request was requested to be traced (e.g. by the user for debugging). This should be a separate flag as it relates to the entire trace and not just to the last hop. This field also **must not** be modified. The proposal is to call the flag _requested-trace_. Examples are:

* user requested to trace a request from e.g. Chrome Dev Tools
* a synthetic test users want to collect end-to-end information for

SergeyKanzhelev:
is it intentional that `Undecided` is `2`. With the `0` in a `Traced` bit? Not `3` with `1` in Traced bit?

tylerbenson:
By calling it a sampling *bit*, that implies to me this is a bitfield, so I assume 3 would portray an odd notion of both sampled and undecided.

PR #142, opened 2018-08-07
Change traced flag semantics and add recorded flag

erabug:
This PR proposes slightly changing the semantics of the `traced` (also called `sampled`) flag in trace-options, and adding a new flag.

* `traced?` becomes `requested?` and describes whether someone explicitly requested this trace be captured
  * This value is either true or false
  * Once true, this value cannot be downgraded to false
* `recorded` describes whether your parent recorded its tracing information out-of-band
  * This value is either no or maybe
  * This value can change from node to node, it just describes what happened in the last hop

**Background**

There are three use cases I've heard described:

1. As an end-user, I would like to see an end-to-end trace for a particular request. I would like to mark a request for tracing and have that decision propagated and respected by all downstream callers.

_(This is pretty hard to do, since you can't actually force anything downstream to do something. Although also called "force" or "debug" trace, this is more like a "pretty-please" request or hint.)_

2. As an end-user, I would like to see my request traced through multiple implementations/vendors. If a tracer sees that its caller was from a different implementation, it can record that information (trace ID, span ID, and implementor name) so I can go find more information from that implementor. If the upstream caller did not record any additional information (such as timing), it saves me a lookup.

_(Because the hint cannot actually be trusted, i.e. you can't trust that ALL the upstream nodes actually recorded information, just understanding what your parent did is helpful.)_

3. As an end-user, I would like traces to include proxies and load balancers.

_(However, at the time the request hits the proxy or load balancer, you may not know whether or not you should sample. You want to defer the decision.)_

**Proposal**

These use cases can be solved by two different signals: (1) did someone explicitly request that I record this? (2) did my parent propagate its tracing information out of band? These signals can be sent separately as two trace-option flags.

| option | recorded? | requested? | recording probability | situation                                                          |
| ------ | --------  | ---------- | --------------------- | ------------------------------------------------------------------ |
| 00     | no        | false      | low                   | I definitely dropped the data and no one asked for it              |
| 01     | no        | true       | medium                | I definitely dropped the data but someone asked for it             |
| 10     | maybe     | false      | medium                | Maybe I recorded this but no one asked for it yet (maybe deferred) |
| 11     | maybe     | true       | high                  | Maybe I recorded this and someone asked for it                     |

I believe these changes address the concerns of PR #122:

>"sampling states"
>* Undecided - Tracing data collection decision is deferred. (2)
>* Traced - Last hop of request may be recorded. (1)
>* Not traced - Last hop of request is not recorded. (0)

The `recorded?` flag covers 0 and 1 and `requested?` covers 2.

>"force/debug"
>* user requested to trace a request from e.g. Chrome Dev Tools
>* a synthetic test users want to collect end-to-end information

These are covered by `requested?`.

> "deferred"
>* Injecting IDs in the first hop but not collecting a trace and not making a sampling decision
>* A load balancer waiting for a request to return and deciding on whether to keep a trace.

The two deferred cases are covered by `10`.

At @SergeyKanzhelev's suggestion, I added the table to the spec. While it describes all existing flags, those are both specific to "sampling" and may not scale well if we add more unrelated flags in the future. (On a related note, @tedsuo suggested renaming whole trace-options section to sampling-options until the point where we have unrelated flags.)

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -118,12 +118,15 @@ static final byte FLAG_TRACED = 1; // 00000001
 boolean traced = (traceOptions & FLAG_TRACED) == FLAG_TRACED
 ```
 
-#### Traced Flag (00000001)
+#### Requested Flag (00000001)
 When set, the least significant bit recommends the request should be traced. A caller who
 defers a tracing decision leaves this flag unset.
 
+#### Recorded Flag (00000010)
+When set, the least significant bit documents that the caller may have recorded trace data. A caller who does not record trace data out-of-band leaves this flag unset.
+
 #### Other Flags
-The behavior of other flags, such as (00000010) are undefined.
+The behavior of other flags, such as (00000100) are undefined.
 
 ## Examples of HTTP headers
 
@@ -258,8 +261,9 @@ If the value of `traceparent` field wasn't changed before propagation - `tracest
 Here is the list of allowed mutations:
 
 1. **Update `span-id`**. The value of property `span-id` can be regenerated. This is the most typical mutation and should be considered a default.
-2. **Mark trace for sampling**. The value of `sampled` flag of `trace-options` may be set to `1` if it had value `0` before. `span-id` MUST be regenerated with the `sampled` flag update. This mutation typically happens to mark the importance of a current distributed trace collection.
-3. **Restarting trace**. All properties - `trace-id`, `span-id`, `trace-options` are regenerated. This mutation is used in the services defined as a front gate into secure network and eliminates a potential denial of service attack surface. 
+2. **Request trace capture**. The value of `requested` flag of `trace-options` may be set to `1` if it had value `0` before. `span-id` MUST be regenerated with the `requested` flag update. This mutation typically happens to mark the importance of a current distributed trace collection.
+3. **Update `recorded`**. The value of `recorded` reflects the caller's recording behavior: either the trace data were dropped or they may have been recorded out-of-band. This mutation gives the downstream tracer information about the likelihood its parent's information was recorded.
+4. **Restarting trace**. All properties - `trace-id`, `span-id`, `trace-options` are regenerated. This mutation is used in the services defined as a front gate into secure network and eliminates a potential denial of service attack surface. 
 
 Libraries and platforms MUST NOT make any other mutations to the `traceparent` header.
 
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -118,6 +118,15 @@ static final byte FLAG_TRACED = 1; // 00000001
 boolean traced = (traceOptions & FLAG_TRACED) == FLAG_TRACED
 ```
 
+### Flag behavior
+
+| option       | recorded? | requested? | recording probability | situation                                                          |
+| ------------ | --------  | ---------- | --------------------- | ------------------------------------------------------------------ |
+| 00000000     | no        | false      | low                   | I definitely dropped the data and no one asked for it              |
+| 00000001     | no        | true       | medium                | I definitely dropped the data but someone asked for it             |
+| 00000010     | maybe     | false      | medium                | Maybe I recorded this but no one asked for it yet (maybe deferred) |
+| 00000011     | maybe     | true       | high                  | Maybe I recorded this and someone asked for it                     |
+
 #### Requested Flag (00000001)
 When set, the least significant bit recommends the request should be traced. A caller who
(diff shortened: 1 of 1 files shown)


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -118,12 +118,15 @@ static final byte FLAG_TRACED = 1; // 00000001
 boolean traced = (traceOptions & FLAG_TRACED) == FLAG_TRACED
 ```
 
-#### Traced Flag (00000001)
+#### Requested Flag (00000001)
 When set, the least significant bit recommends the request should be traced. A caller who
 defers a tracing decision leaves this flag unset.
```
discostu105: I would add something like this:

"An implementation might still decide on it's own if this flag is respected or not. Consider a scenario when capturing a trace induces cost. An upstream caller could set this flag to maliciously induce cost in a tracing system. It's recommended that an implementation is careful about this and only respects the flag if the upstream caller is trusted."

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -118,12 +118,15 @@ static final byte FLAG_TRACED = 1; // 00000001
...
 
+#### Recorded Flag (00000010)
+When set, the least significant bit documents that the caller may have recorded trace data. A caller who does not record trace data out-of-band leaves this flag unset.
```
discostu105: Should we be more clear about the "Definitively not recorded" and "Maybe recorded" states here? Otherwise discussion will spur again, when an implementation cannot definitively declare if it recorded a trace ("undecided").

SergeyKanzhelev (approved):

codefromthecrypt:
FWIW I don't like this for many reasons.

Firstly, it is it translated in a non-user friendly way. For example, say we did have two options like sampled and debug which exist today. The way this translated in is a way that forces people to think in bit flags. The documentation doesn't show how it will look if you did have both flags set except requested/not requested. Adding "recorded" will show how this will look very odd to people.
https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#flag-behavior

Secondly the mapping isn't semantically correct at least not in a way that matches existing stuff. For example, it is adding a trait "recorded" which is hard to propagate accurately while also muddying the "sampled" state in a way that conflates itself with "debug" or "force trace". For example, "requested" is more muddy than yes/no/abstain of something sampled it.. It doesn't naturally have any more information about user intent. "force trace" is very edge case on the other hand and more important. The whole thing seems like good intentions but missed some experience that is already in B3 and Amazon trace specifications yet muddied with this design. It is also missing real world situation where what is recorded is quite flexible with regards to sampling. The abstraction of traceparent if done well allows a simple compatible route for existing sites who have used similar mechanics for several years in various open source tools... more on this below.

Finally, I'm surprised folks like @felixbarny don't speak up and "just implement it" This is a bad sign that people just accept whatever eventhough it doesn't match in a reliable way. Lacking outreach of those directly impacted, lacking folks pushing back at all, and adding rush to merge when meetings happen doesn't lead way for a solid standard, just "another standard". Please next time reach out directly especially knowing that B3 and Amazon both do not follow this, and both represent a significant stake.

justindsmith:
I had a clarifying question around this as well. (This might be my ignorance of the previous specs - b3, etc - or the history of this spec.)

Does the `requested` flag represent:
1. A human user's request to explicitly have a specific trace be captured
2. A sampler's decision to have a specific trace be captured
3. Both 1 and 2

> (1) did someone explicitly request that I record this?

I guess I'm looking for a definition of who "someone" is.

Thanks!

codefromthecrypt:
In B3 and things based on it or derived from similar work, or previous
version of this (including grpc format and the original work on this
before w3c)...

sampled was not trying to explain who, rather what. For example it was
sampled upstream, or not, or no decision has been made yet.

debug/force trace in B3 and derived things is usually user
involvement, and is in a way emphasizing the sampled decision, but it
doesn't imply a user did this. It could be automation, for example. At
any rate it is edge case.

The "who" thing about the someone.. there's no prior practice I'm
aware of, at least in open source. My suggestion is to revert this
change as since this is a standard, we should do our best to leverage
what's out there... and we were, but this changed that.

SergeyKanzhelev:
`requested` was described the same way as sampled would be. Basically it is a decision of a system upstream on whether to keep this trace or not. 

`recorded` is about your parent's decision. Main idea is to avoid situation when vendor which wants to decide on sampling later had a way to tell about it downstream without always changing of `requested` flag.

There is no explicit `debug` flag in the spec. `requested` can be used for this purpose.

SergeyKanzhelev:
needs clarification in a doc, I'll send a PR tomorrow morning.

justindsmith:
Thanks! That helps!

Issue #137, opened 2018-07-27
Allow multiple tracestate entries for one vendor


discostu105:
__Use Case__

Multi-tenant tracing.

Currenty TraceContext elegantly solves the problem of different tracing system that interact with each other. Via _tracestate_, "foreign" vendor information is transported throughout the trace. This is great, but a system traced only by one vendor can have a similiar problem. If services are monitored by different tenants/customers of that vendor, the trace breaks (assuming the vendor is relying on _tracestate_ information).

__Concrete example__
A 3-tier trace where the first and the last tier are monitored by the same tenant (A), but the in-between tier is monitored by a different tenant (B), but both use the same APM vendor.

(A) -> (B) -> (A)
```
1. tracestate="vendor1=tenant:0xaaa;importantData:0xfff"
2. tracestate="vendor1=tenant:0xbbb;importantData:0xeee"
3. tracestate="vendor1=tenant:0xaaa;importantData:0xccc"  // trace for tenant A is broken
```

Since, with current design, a vendor can only use one key in _tracestate_, the trace would break.

__Possible Solutions__

 1. Vendor needs to encode multi-tenant support into a single _tracestate_ entry.
```
1. tracestate="vendor1=(tenant:0xaaa;importantData:0xfff)"
2. tracestate="vendor1=(tenant:0xbbb;importantData:0xeee)|(tenant:0xaaa;importantData:0xfff)"
3. tracestate="vendor1=(tenant:0xaaa;importantData:0xfff)|(tenant:0xbbb;importantData:0xeee)"
```

 2. Vendor is allowed to have multipe _tracestate_ enties with the same key.
```
1. tracestate="vendor1=tenant:0xaaa;importantData:0xfff"
2. tracestate="vendor1=tenant:0xbbb;importantData:0xeee, vendor1=tenant:0xaaa;importantData:0xfff"
3. tracestate="vendor1=tenant:0xaaa;importantData:0xfff, vendor1=tenant:0xbbb;importantData:0xeee"
```

 3. Vendor is allowed to encode dynamic info into the _tracestate_ key.
```
1. tracestate="vendor1-0xaaa=importantData:0xfff"
2. tracestate="vendor1-0xbbb=importantData:0xeee, vendor1-0xaaa=importantData:0xfff"
3. tracestate="vendor1-0xaaa=importantData:0xfff, vendor1-0xbbb=importantData:0xeee"
```

__Thoughts__

[...]

PR #153, opened 2018-08-07
Allow @ in tracestate key names for multi-tenant vendors

SergeyKanzhelev:
#137

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -225,10 +225,11 @@ Identifiers are short (up to 256 characters) textual identifiers.
 
 ```
 key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+key = lcalpha 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" ) "@" lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
 lcalpha    = %x61-7A ; a-z
 ```
 
-Note that identifiers MUST begin with a lowercase letter, and can only contain lowercase letters `a`-`z`, digits `0`-`9`, underscores `_`, dashes `-`, asterisks `*`, and forward slashes `/`.
+Note that identifiers MUST begin with a lowercase letter, and can only contain lowercase letters `a`-`z`, digits `0`-`9`, underscores `_`, dashes `-`, asterisks `*`, and forward slashes `/`. For multi-tenant vendors scenarios `@` sign can be used to prefix vendor name. Suggested use is to allow set tenant id in the beginning of key like `fw529a3039@dt` - `fw529a3039` is a tenant id and `@dt` is a vendor name. Searching for `@dt=` would be more robust for parsing (searching for all vendor's keys).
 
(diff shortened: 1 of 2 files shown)


trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -64,6 +64,11 @@ Lowercase names has a few benefits:
 Url encoding is low-overhead way to encode unicode characters for non-latin characters in the 
 values. Url encoding keeps a single words in latin unchanged and easy readable.
 
+## Vendor name in a key
+
+Sign `@` is allowed in a key for easy parsing of vendor name out of tracestate key. Idea is that with the registry of trace vendors one can easily understand the vendor name and how to parse it's trace state. Without `@` sign parsing will be more complicated. Also `@` sign has known semantics in
+addressing for protocols like ftp and e-mails.
```
discostu105, line 70: unintentional line-break?

discostu105, line 70: Also, not sure if "Also `@` sign has known semantics in addressing for protocols like ftp and e-mails" really adds a lot here. Would it hurt to remove this?

SergeyKanzhelev: it's markdown

SergeyKanzhelev: I think it's fine. Any way to explain it easier?

discostu105: I't ok, I'm also fine with keeping this.

AloisReitbauer (approved):

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -225,10 +225,11 @@ Identifiers are short (up to 256 characters) textual identifiers.
 
 ```
 key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+key = lcalpha 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" ) "@" lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
 lcalpha    = %x61-7A ; a-z
 ```
 
-Note that identifiers MUST begin with a lowercase letter, and can only contain lowercase letters `a`-`z`, digits `0`-`9`, underscores `_`, dashes `-`, asterisks `*`, and forward slashes `/`.
+Note that identifiers MUST begin with a lowercase letter, and can only contain lowercase letters `a`-`z`, digits `0`-`9`, underscores `_`, dashes `-`, asterisks `*`, and forward slashes `/`. For multi-tenant vendors scenarios `@` sign can be used to prefix vendor name. Suggested use is to allow set tenant id in the beginning of key like `fw529a3039@dt` - `fw529a3039` is a tenant id and `@dt` is a vendor name. Searching for `@dt=` would be more robust for parsing (searching for all vendor's keys).
```
codefromthecrypt, line 232: I think you all are way too quick to make decisions like this.

codefromthecrypt:
for posterity there was mention about no objections. in the doc it also
said that many people were only partially available monday. can folks that
agree we should document this practice please make yourselves known, or at
least say who was present when no objections were claimed?

it is good to make decisions but it should be transparent which
"implementors" will be using this and how. I am not saying this is a bad
call just didn't see folks outside DT talking about their use cases for
this.

PR #157, opened 2018-08-13
Language about security risks

SergeyKanzhelev:
#55


trace_context/security.md
```diff
@@ -0,0 +1,25 @@
...
+considered as sensitive. For example, `traceparent` may allow to correlate calls so identify from one call may be correlated
+to the data sent with another call. `tracestate` may imply the version of monitored software used by caller. This
+information can be used to author an attack to the service.
+
+Even though the risk is minimal, application owner should make sure to control the boundary of organization's administrative
+control. Tracing vendors and platforms should allow to control any unintentional propagation of distributed tracing context.
+
+## Denial of monitoring by malicious caller
+
+When distributed tracing was enabled on a service with a public API and naively continue any trace with
```
yurishkuro: s/was/is/

trace_context/security.md
```diff
@@ -0,0 +1,25 @@
+# Security
+
+There are two areas of security risk - information exposure to malicious dependency service and denial of monitoring
+by malicious caller. Both risks exist with modern approaches for distributed trace context propagation. Standard makes
+it more likely for services to get these attacks.
```
yurishkuro: why does it make it more likely?  The exact same risks exist today. Obscurity is not security.

trace_context/security.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Information exposure
+
+As mentioned in privacy section, information in `traceparent` and `tracestate` headers may carry information that can be
+considered as sensitive. For example, `traceparent` may allow to correlate calls so identify from one call may be correlated
+to the data sent with another call. `tracestate` may imply the version of monitored software used by caller. This
+information can be used to author an attack to the service.
+
+Even though the risk is minimal, application owner should make sure to control the boundary of organization's administrative
```
yurishkuro: > Even though the risk is minimal

is an opinion, why do we need to state it?

trace_context/security.md
```diff
@@ -0,0 +1,25 @@
...
+control. Tracing vendors and platforms should allow to control any unintentional propagation of distributed tracing context.
```
yurishkuro: "control ... administrative  control" - could you please restate this more clearly?

trace_context/security.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Denial of monitoring by malicious caller
+
+When distributed tracing was enabled on a service with a public API and naively continue any trace with
+the sampling flag set using the `trace-id` provided via `traceparent` field, a malicious attacker could overwhelm an
```
yurishkuro: "set using the `trace-id`" not necessary (makes it hard to read)

trace_context/security.md
yurishkuro: s/naively continue/naively continues/

trace_context/security.md
SergeyKanzhelev: The point here is that it is more likely to have generic attacks for standard headers than special attack for specialized `B2` or .NET headers. Like everybody tries to access wordpress admin portal to every domain without even checking. Do you think I can express this though clearer?

mtwo: Perhaps we could reword the last sentence to "Having a standard for context propagation provides attackers with a known way of triggering tracing within an application"

trace_context/security.md
mtwo: Agreed, though I don't think it's a big deal. We can remove "Even though the risk is minimal"

mtwo: Also, "application owner" should be changed to "application owners"

mtwo: Let's change the paragraph to "Application owners should either ensure that no proprietary or confidential information is stored into the tracestate, or they should ensure that tracestate isn't present in requests to external systems".

mtwo (changes requested):
I made some comments requesting minor changes, looks good otherwise

trace_context/security.md
```diff
@@ -0,0 +1,36 @@
...
+the sampling flag set, a malicious attacker could overwhelm an application with tracing overhead, forge trace ID collisions
+that make monitoring data unusable, or run up your tracing bill with your SaaS tracing vendor.
+
+Tracing vendors and platforms should account to these situations and make sure that checks and balances are in place to
+protect denial of monitoring by malicious or badly-authored callers.
+
+## Other risks
+
+Application owner need to make sure to test all code path leading to sending of `traceparent` and `tracestate` headers. For
+example, browsers single page applications it is typical to make cross-origin calls from the browser. If one of code path leads
```
yurishkuro: > browsers single page applications it is typical

something's off here

yurishkuro: > If one of code path leads

paths

mtwo:
I went through the security.md file and updated the grammar

mtwo (approved):

PR #158, opened 2018-08-14
added ordering rationale:

### LINKED ISSUE #123: Structure of TraceState
*Requirements*
* We need to know, whether we are the parent or not. 
* We need to understand how to cut off headers. 
* We need to know who our parents are for diagnostics purposes
* We can directly link tracing data information from other systems.
* Intermediary order helps to understand whether to query third parties. 

**Do we need ordering? (Issue #81)**

* Do we need **full** ordering or do we just now the last hop?
**  It might prove helpful, but is not needed.

* Does everybody need to have a trace-state entry; even if it is empty?
Yes, all tracers need to have a trace state. Otherwise you only generate IDs. 

* Do we only need to know the upstream tracer (Issue #80)?

**Are we using the dictionary format of IETF?**

**What do we do, when trace-state gets too long?**

**How do we ensure unique key names?**

* Proposal about using a vendor-directory (Issue #58)

SergeyKanzhelev:
I suggest to close #123 with the words about ordering in rationale document. #123 refers to other two issues which are tracked separately already

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
+++ b/trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
@@ -40,7 +40,12 @@ by the same tracing vendor. So information supplied by tracing system originated
 request will typically be less and less important deeper in distributed trace. Immediate
 caller's information on other hand typically is more valuable as it is more likely being
 monitored by the same tracing vendor. Thus it is logical to move immediate caller's
-information to the beginning of the `tracestate` list.
+information to the beginning of the `tracestate` list. So less important values will be
+pushed to the end of the list.
+
+This prioritization of `tracestate` values improves performance of querying value of
+tracestate - typically you only need a first pair. It also allows to meaningfully truncate
+`tracestate` when required instead of dropping the entire list of values.
 
 ### Size limits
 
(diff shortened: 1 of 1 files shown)


trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
```diff
@@ -31,6 +31,22 @@ interfere with identification of the tracing system responsible for an entry.
 - The typical name will be a single word in latin and the value will be a
 copy of the `traceparent` format or an opaque string.
 
+## Ordering of keys in `tracestate`
+
+Specification calls for ordering of values in tracestate. This requirements allows better interoperability between tracing vendors.
```
bogdandrutu, line 36: I think this md file uses different line size based on whoever updates it. Probably makes sense to format the entire file using some rules. Maybe in a different PR. But at least in this PR use consistent line sizes please.

bogdandrutu (approved):

trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
SergeyKanzhelev: I noticed this as well. I wonder if there is a setting common between vim and vscode that will control it. I'll configure this for vscode.

PR #162, opened 2018-08-23
Add clarification for #142

SergeyKanzhelev:
Add clarification for #142

... and couple cosmetic changes


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -144,12 +149,37 @@ boolean traced = (traceFlags & FLAG_TRACED) == FLAG_TRACED
 
 #### Requested Flag (00000001)
```
codefromthecrypt, line 150: I think this should be renamed to sampled as that's far more common usage

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -144,12 +149,37 @@ boolean traced = (traceFlags & FLAG_TRACED) == FLAG_TRACED
...
+
+Caller who defers a tracing decision SHOULD leaves this flag unset. Caller may
+communicate the priority of deferred decision (most likely will be collected)
+using `recorded` flag.
+
+Flag `recorded` can also be used to communicate the ask to collect information
+or notify that information was collected and available for querying from the
+caller service.
 
 #### Recorded Flag (00000010)
```
codefromthecrypt, line 178: I think this flag should be removed as it adds little value and results in a hex character "03" which is not intuitive either

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: "sampled" as a name has a confusion. If "sampled"  means "sampled in" or "sampled out"? The name `requested` eliminates this confusion.

`requested` creates another confusion. Is it `debug` or `sampled`. As specification does not define how sampling probability or sampling rate can be propagated - this confusion is easier to describe in text.

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: The problem this flag addresses is the ability for vendor to communicate priority of deferred decision. With the semantic "if in doubt - try to collect". See this:

> Caller who defers a tracing decision SHOULD leaves this flag unset. Caller may
communicate the priority of deferred decision (most likely will be collected)
using `recorded` flag.
> Flag `recorded` can also be used to communicate the ask to collect information
or notify that information was collected and available for querying from the
caller service. 

Basically it is a relieve valve to minimize the abuse of `requested` flag.

I understand the problem of `03`, but if we treat those two characters as bitmask - there is very little we can do with this problem unless we are sure there will be no more flags. In fact having two flags not will find out "bad implementations" earlier.

codefromthecrypt:
this does not address sampled. Please get feedback from others not in the
room please

SergeyKanzhelev:
Created https://github.com/w3c/distributed-tracing/issues/167 for discussion

SergeyKanzhelev:
@bogdandrutu @adriancole can you sign off this PR as it just adds clarification for flags introduced in previous PR. Let's discuss renaming and removing in a separate issue.

codefromthecrypt:
I wasn't there at the meeting to say what you've written was clear or not. Not sure bogdan was either. Whoever was there that resulted in this change should ideally be the one clicking approve as they would know.

AloisReitbauer:
I looked at the text and I think the language is still confusing for people. I think we need to cut the text to two things: What the sender intends the receiver to do. How the receiver can change previous decisions.

SergeyKanzhelev:
@AloisReitbauer what people? It is already split into meaning and mutations. Please suggest changes in text

SergeyKanzhelev:
closed in favor of #180

PR #164, opened 2018-08-23
explanation why mutations of tracestate are not required

### LINKED ISSUE #117: Allow to mutate mutliple keys of tracestate
Spec should allow to mutate multiple `tracestate` entries in a single mutation. Here are scenarios that might be enabled:

## 1. Share base SDK

Single base SDK that has it's logic of updating `tracestate` key may need an extensibility model for special type applications with additional `tracestate` key required.

Application Insights is designed for high-tenant apps. SDK today requires to propagate an `app-id` between components which allows to build a full trace efficiently. By knowing `app-id` of a caller backend knows which telemetry store to query for data.

There are also internal applications using hierarchical identifiers to address spans. Hierarchical identifiers will not necessarily be implemented in Application Insights SDK. Also not every component is an internal app and it may not support hierarchical identifier.

That's said internal applications will need two keys in `tracestate`. One that is set and maintained by base SDK and containing `app-id`. It will enable high-tenancy. Second - for hierarchical identifier and additional scenarios it brings.

One alternative here is to host smaller `tracestate` inside one of `tracestate`'s key. This smaller `tracestate` will also need key/value store and values will point to specific `span-id` to know who was the parent in it's own trace graph. However this makes field value look worth without benefit. 

## 2. Share trace properties

It is beneficial for many scenarios to share some common trace characteristics like sampling score. See this https://github.com/w3c/distributed-tracing/issues/8#issuecomment-383267825.

Sharing a key in `tracestate` allows to enable even more vendors collaboration scenarios. Alternative for this is parsing each-other headers. 

Drawback of documenting trace properties is a risk that `tracestate` will be used as a [baggage](https://github.com/w3c/distributed-tracing/pull/110#issuecomment-383626477).

SergeyKanzhelev:
Closes #117 and #80 for now.

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
+++ b/trace_context/HTTP_HEADER_FORMAT_RATIONALE.md
@@ -31,6 +31,42 @@ interfere with identification of the tracing system responsible for an entry.
 - The typical name will be a single word in latin and the value will be a
 copy of the `traceparent` format or an opaque string.
 
+## Mutations of `tracestate`
+
+Two questions that comes up frequently is whether the `tracestate` header HAVE TO
+be mutated on every mutation of `span-id` to identify the vendor which made this
+change and whether two different vendors can modify the `tracestate` entries in
+a single component.
+
+This requirement may improve interoperability between vendors. For instance,
+vendor may check the first `tracestate` key and provide some additional value
+for the customer by adjusting data collection in the current component via the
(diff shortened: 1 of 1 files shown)


bogdandrutu (approved):
I think it is a good idea to not require the tracestate mutation.

SergeyKanzhelev:
I'll wait with the merge to give opportunity for comments

PR #166, opened 2018-08-24
Implement Test Harness

reyang:
This is the test harness which will be used for standard compliance testing.
1. test harness
2. initial set of test cases
3. README.md


AloisReitbauer (commented):
Let us discuss this in the next meeting than rather pushing it now,

SergeyKanzhelev:
@AloisReitbauer please post feedback here. In-person discussions can help clarifying something, but written feedback helps to move it forward.

NatalieWolfe:
Gave the test harness a whirl with our node agent. One issue is our agent samples requests, it doesn't trace everything. Feels kind of like cheating to have a "testing mode" for the agent to run under just to pass the tests. The only difference would be ignoring the sample limit. Any suggestions?

For the harness itself, it looks like there is no way to change the endpoint it hits without opening the dev console and clearing all storage for the site. The first time I opened it I did not enter a url, just to see what the UI looked like, and it rendered the harness unusable until I cleared out all storage from Chrome.

It could also be safer at guarding against bad implementations. If the tested application does not respond with any trace headers then the error you get is a cryptic `cannot read property "headers" of undefined` message. Had to dive into the implementation of the harness to see what was going wrong.

I'm still getting some test failures that I'm not sure about, but those could actually be issues in our implementation. Wanted to get my initial feedback in before spending too much time debugging.

reyang:
Hi @NatalieWolfe, thanks for giving it a try, like the feedbacks!

1. Currently my suggestion would be having a specific endpoint for testing purpose with sample limit ignored. As people discussed during the W3C meeting, we intend to have different endpoints exposed by the "service to be tested" based on scenarios (e.g. http://myservice.org/test-add-stuff-to-tracestate, http://myservice.org/test-pass-through-proxy).

 2. I got feedbacks that we want to converge to a single programming language and focus more on the console mode. I will remove the JavaScript/browser part and add a Python test script - with the test endpoint as the command line arguments. To your original question, there is a way without having to clear the browser cookie - from Chrome, open a new tab, put the URL with your test endpoint in the anchor (e.g. http://127.0.0.1:8080#http://your-test-endpoint.org:5000/foobar).

3. Definitely! This is the initial PR to get feedback on the design, I will add more check for robustness (e.g. currently there is no prevention to malicious request, no timeout management).

Again, thank you for trying out and the feedbacks are very valuable. I will make changes and switch to Python CLI mode next week, please have a try and let me know your feedbacks.

test/test.py
```diff
@@ -0,0 +1,188 @@
...
+	traceparent_name_re = re.compile(r'^traceparent$', re.IGNORECASE)
+	traceparent_format = r'^([0-9a-f]{2})-([0-9a-f]{32})-([0-9a-f]{16})-([0-9a-f]{2})$'
+	traceparent_format_re = re.compile(traceparent_format)
+
+	def get_traceparent(self, headers):
+		retval = []
+		for key, value in headers:
+			if self.traceparent_name_re.match(key):
+				retval.append((key, value))
+		self.assertEqual(len(retval), 1, 'expect one traceparent header, got more {!r}'.format(retval))
```
discostu105: this assertion also hits if there are zero traceparent headers, so the "got more" text is misleading.

discostu105:
Gave the latest (python-only) version of the testharness a try.

Looks really good. Most tests fail for dynatrace right now, but these are subject to be fixed :). Things I'd wish for:
 * The output lists all failed tests with a failure-message. That's fine, but I'd somehow also wished to see the list of passed test (to know which tests were executed at all).
 * A debug-mode which spits out the details about every received request. Would help a bit to diagnose failing tests.
 * For debugging a specific test, it would be cool if we could filter for one test per commandline argument.

But these are just nice-to-have requests. IMHO this is already a very good start.

discostu105:
@NatalieWolfe 
> One issue is our agent samples requests, it doesn't trace everything. Feels kind of like cheating to have a "testing mode" for the agent to run under just to pass the tests. The only difference would be ignoring the sample limit. Any suggestions?

Isn't an implementation supposed to forward the tracecontext, even if a request is not sampled?

test/test.py
reyang: @discostu105 nice catch! Fixed.

reyang:
@discostu105 Appreciate the suggestions!

1. Sure, I'll make it happen. (update: it's done)
2. I plan to add a DEBUG flag which dumps all the HTTP transactions. (update: HARNESS_DEBUG is added, please have a try)
3. It is supported, I'll update the README file to mention this (probably will come up with a better command line syntax). You can give it a try first:
```
> python -m unittest test.TraceContextTest.test_traceparent_header_name
```
update: new syntax supported -

```python test.py http://127.0.0.1:5000/test TraceContextTest.test_traceparent_duplicated```

discostu105:
@reyang thanks for those improvements. They are really helpful! After a few fixes, the dynatrace implementation now passes all tests :)

    C:\workspaces\arch\distributed-tracing-testharness.git\test>python test.py http://localhost:4500/api/forward
    harness listening on http://127.0.0.1:7777
    test_multiple_requests (__main__.AdvancedTest) ... ok
    test_both_traceparent_and_tracestate_missing (__main__.TraceContextTest) ... ok
    test_traceparent_duplicated (__main__.TraceContextTest) ... ok
    test_traceparent_header_name (__main__.TraceContextTest) ... ok
    test_traceparent_header_name_valid_casing (__main__.TraceContextTest) ... ok
    test_traceparent_included_tracestate_missing (__main__.TraceContextTest) ... ok
    test_traceparent_trace_flags_illegal_characters (__main__.TraceContextTest) ... ok
    
    ----------------------------------------------------------------------
    Ran 7 tests in 0.797s
    
    OK

SergeyKanzhelev (approved):

SergeyKanzhelev:
@discostu105 it's time to contribute more tests then =)

NatalieWolfe:
@discostu105 

>> One issue is our agent samples requests, it doesn't trace everything. Feels kind of like cheating to have a "testing mode" for the agent to run under just to pass the tests. The only difference would be ignoring the sample limit. Any suggestions?
>
> Isn't an implementation supposed to forward the tracecontext, even if a request is not sampled?

Forwarding the trace context is fine, but some tests [specifically check that a new span is created](https://github.com/w3c/distributed-tracing/blob/master/test/test.py#L116-L125) or that [a new trace is created](https://github.com/w3c/distributed-tracing/blob/master/test/test.py#L127-L138). When not sampling, our agent acts just as a proxy without modifying the trace context at all. Thus in sampling mode, these tests fail.

reyang:
@discostu105 @NatalieWolfe we can have different group of test cases (e.g. adding tags to test cases) for different mode:
1. **service mode** - the service is expected to process traceparent header - by inheriting the trace_id and creating a new span_id
1.1. **simple mode** - the service wouldn't modify tracestate
1.2. **interop mode** - the service would prepend vendor specific key-value pair to tracestate
2. **pass-thru mode** - traceparent and tracestate are not modified
3. **security gateway mode** - always restart trace

What do you think?

reyang:
@discostu105 is the sample rate configurable or can we make it configurable?

I think a viable solution would be running the agent with sample rate equals to 0% and 100% separately.
1. agent sample rate = 0%, this would make the agent a full proxy, we can run all the **pass-thru mode** test cases against it.
2. agent sample rate = 100%, this would make the agent behave like the **service mode**.

PR #188, opened 2018-10-03
allow empty headers

SergeyKanzhelev:
Implementing `tracestate` for various languages proved it's very hard to conform to the empty `tracestate` header test case. Code becoming too complex. Thus allowing empty value to simplify implementation. 

Also addresses https://github.com/w3c/distributed-tracing/issues/171

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -207,6 +207,11 @@ Platforms and libraries MUST expect header name in any casing and SHOULD send he
 
 ## Header value
 
+Multiple `tracestate` headers are allowed. Values from multiple headers in
+incoming requests SHOULD be combined in a single header according to the
+[RFC7230](https://tools.ietf.org/html/rfc7230#page-24) and send as a single
+header in outgoing request.
+
 This section uses the Augmented Backus-Naur Form (ABNF) notation of [RFC5234](https://tools.ietf.org/html/rfc5234), including the DIGIT rule in [appendix B.1 for RFC5234](https://tools.ietf.org/html/rfc5234#appendix-B.1). It also includes the OWS rule from [RFC7230 section 3.2.3](https://tools.ietf.org/html/rfc7230#section-3.2.3).
 
 `DIGIT` rule defines number `0`-`9`.
@@ -215,16 +220,25 @@ The `OWS` rule defines an optional whitespace. It is used where zero or more whi
 
 The `tracestate` field value is a `list` as defined below. The `list` is a series of `list-members` separated by commas `,`, and a `list-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `list-member`s are ignored. There can be a maximum of 32 `list-member`s in a `list`.
 
+Empty and whitespace-only list elements are allowed as a list value. But the use
+of empty `tracestate` header should be considered an exception. The reason for
+allowing of empty value in `tracestate` is a difficulty for implementor to
+recognize the empty value when multiple `tracestate` headers were sent.
+Whitespace characters are allowed for a similar reason as some frameworks will
+inject whitespace after `,` separator automatically even in case of an empty
+header.
+
 A simple example of a `list` with two `list-member`s might look like: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`.
 
-```
+``` abnf
 list  = list-member 0*31( OWS "," OWS list-member )
 list-member = key "=" value
+list-member = OWS
 ```
 
 Identifiers are short (up to 256 characters) textual identifiers.
 
-```
+``` abnf
 key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
 key = lcalpha 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" ) "@" lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
 lcalpha    = %x61-7A ; a-z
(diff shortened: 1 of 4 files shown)


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -207,6 +207,11 @@ Platforms and libraries MUST expect header name in any casing and SHOULD send he
 
 ## Header value
 
+Multiple `tracestate` headers are allowed. Values from multiple headers in
+incoming requests SHOULD be combined in a single header according to the
```
yurishkuro, line 211: Why SHOULD?

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -215,16 +220,25 @@ The `OWS` rule defines an optional whitespace. It is used where zero or more whi
 
 The `tracestate` field value is a `list` as defined below. The `list` is a series of `list-members` separated by commas `,`, and a `list-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `list-member`s are ignored. There can be a maximum of 32 `list-member`s in a `list`.
 
+Empty and whitespace-only list elements are allowed as a list value. But the use
```
yurishkuro: I recommend using consistent terminology: list members or list elements, not both.

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -215,16 +220,25 @@ The `OWS` rule defines an optional whitespace. It is used where zero or more whi
...
+of empty `tracestate` header should be considered an exception. The reason for
```
yurishkuro: What does it mean "to be considered exception"? Does http even allow headers without values?

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: The idea is to increase interoperability by making sure it's always a single string with defined order of elements and easy to calculate length. From implementation perspective it is also easier if we talking about libraries, not generic proxies. 

What's your main concern with this "SHOULD"?

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: I clarified

trace_context/HTTP_HEADER_FORMAT.md
yurishkuro: In the HTTP protocol multiple headers and single header with comma-separated values are interchangeable. I don't see the point of putting more stringent restrictions, because a compliant HTTP middleware is free to swap those two representations.

trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -215,16 +220,26 @@ The `OWS` rule defines an optional whitespace. It is used where zero or more whi
...
 The `tracestate` field value is a `list` as defined below. The `list` is a series of `list-members` separated by commas `,`, and a `list-member` is a key/value pair separated by an equals sign `=`. Spaces and horizontal tabs surrounding `list-member`s are ignored. There can be a maximum of 32 `list-member`s in a `list`.
 
+Empty and whitespace-only list members are allowed. Libraries and
+platforms MUST accept empty `tracestate` headers, but SHOULD avoid
+sending them. The reason for allowing of empty list members in
+`tracestate` is a difficulty for implementor to recognize the empty
+value when multiple `tracestate` headers were sent. Whitespace
+characters are allowed for a similar reason as some frameworks will
+inject whitespace after `,` separator automatically even in case of an
+empty header.
```
yurishkuro, line 230: I understand the whitespace use case, but don't understand "empty list member" use case. Under what conditions would that occur?

SergeyKanzhelev: We tried python, C# and Go. And there are libraries which will merge headers for you instead of giving the list. Also some libraries will insert spaces with the comma, some wouldn't. So make spec loose to simplify implementors life

trace_context/HTTP_HEADER_FORMAT.md
SergeyKanzhelev: That's why it SHOULD, not MUST. Same as here:

> Platforms and libraries MUST expect header name in any casing and SHOULD send header name in lower case.

yurishkuro: This is why I object to this. If you MUST treat different inbound formats anyway, putting additional restrictions on the outbound format adds 0 value to anyone, just becomes an arbitrary thing in the spec with no justification.

trace_context/HTTP_HEADER_FORMAT.md
yurishkuro: if libraries join headers into a string, they may add whitespace, but they will not create empty list members unless the inbound headers have empty values, which I think is not allowed by HTTP spec.

discostu105:
I support this PR. Had the same experience when implementing the current spec to conform this test-case. I had to add extra code to detect an empty header and reject a trace based on that. Allowing empty tracestate headers would make it easier.

discostu105 (approved):

AloisReitbauer (approved):

trace_context/HTTP_HEADER_FORMAT.md
AloisReitbauer: We discussed not defining trace-state as a list. Technically, this FORBIDS an implementation to split the header into multiple headers.

SergeyKanzhelev:
@yurishkuro I'll update the terminology. For the SHOULD - I think it's a reasonable request to make for the implementors to merge if possible. It will make less mistakes and implementation that fail to read additional headers

PR #191, opened 2018-10-03
failure to parse clarification (already covered in tests)

SergeyKanzhelev:
Addresses: https://github.com/w3c/distributed-tracing/issues/172

### DIFF
--- a/trace_context/HTTP_HEADER_FORMAT.md
+++ b/trace_context/HTTP_HEADER_FORMAT.md
@@ -47,6 +47,8 @@ which tracing system corresponds with `traceparent`. In this case, since
 
 # TraceParent field
 
+Field `traceparent` identifies the request in a tracing system.
+
 ## Header name
 
 In order to increase interoperability across multiple protocols and encourage successful integration by default it is recommended to keep the header name lower case. Header name is a single word without any delimiters like hyphen (`-`).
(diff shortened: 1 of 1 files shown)


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -195,7 +197,12 @@ Implementation should follow the following rules when parsing headers with an un
 
 # Tracestate field
 
-The `tracestate` HTTP header field conveys information about request position in multiple distributed tracing graphs.
+The `tracestate` HTTP header field conveys information about request
+position in multiple distributed tracing graphs. This header is a
+companion header for the `traceparent`. If library or platform failed to
+parse `traceparent` - it MUST NOT attempt to parse the `tracestate`.
+Note, that opposite it not true - failure to parse `tracestate` MUST NOT
```
reyang, line 204: Would removing "Note, that opposite is not true -" make the doc easier to read?

reyang (approved):

trace_context/HTTP_HEADER_FORMAT.md
zoidyzoidzoid: What if we break it up into multiple paragraphs, mentioning that `tracestate` adds information to the trace we create using the `traceparent`? Emphasising that `tracestate` is additional optional information, but `traceparent` is required, if there is a `tracestate`.

SergeyKanzhelev: calling `tracestate` optional can be misleading as it may be read as "optional to propagate". This paragraph is about parsing and validation rules.

PR #196, opened 2018-10-22
Don't allow spaces in tracestate to make parsing easier

AloisReitbauer:


trace_context/HTTP_HEADER_FORMAT.md
```diff
@@ -279,7 +279,7 @@ The `tracestate` field value is a `list` as defined below. The `list` is a serie
 A simple example of a `list` with two `list-member`s might look like: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`.
 
 ```
-list  = list-member 0*31( OWS "," OWS list-member )
+list  = list-member 0*31("," list-member )
```
reyang, line 282: It is hard to forbid it due to external dependency. Currently we allow multiple tracestate headers, which makes it possible for HTTP frameworks to generate OWS automatically.

For example:
tracestate: A=1
tracestate: B=2

It is possible that an HTTP framework will preprocess the header and give something like **tracestate: A=1, B=2** to the tracecontext library.

reyang:
I understand that we want the spec to be clean and simple, and we want the implementations to be efficient.
My general feeling is that HTTP 1.x protocol is not clean/simple/efficient, it has a long history and is designed with the history compatibility in mind.
The minimum goal for Distributed Tracing HTTP spec is that we should not go against the HTTP spec.

discostu105:
JFYI - as it turns out the IETF spec for Dictionary just changed to make it ordered which fits our requirements petty neatly. During discussion in the WG we tenatively decided to go for this datatype now: https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-08#section-3.1.

FYI @mnot

discostu105:
This is now invalid now, if we want to change to IETF Dictionary (https://github.com/w3c/distributed-tracing/issues/197)

reyang:
Thanks for the updates @discostu105.

PR #215, opened 2018-12-05
Draft of binary format: moved to other repository

SergeyKanzhelev:
# PR was moved

This PR was moved here: https://github.com/w3c/trace-context-binary/pull/2

### DIFF
--- a/spec/Binary/20-BINARY_FORMAT.md
+++ b/spec/Binary/20-BINARY_FORMAT.md
@@ -9,25 +9,86 @@ those as size of the field followed by the value.
 # `Traceparent` binary format
 
 The field `traceparent` encodes the version of the protocol and fields
-`trace-id`, `span-id` and `recorded`. Each field starts with the one byte field
-identifier with the field value following immediately after it. Field
+`trace-id`, `parent-id` and `recorded`. Each field starts with the one byte
+field identifier with the field value following immediately after it. Field
 identifiers are used as markers for additional verification of the value
 consistency and may be used in future for the versioning of the `traceparent`
 field.
 
 ``` abnf
-traceparent     = version version_format
+traceparent     = version version_format  
+version         = 1BYTE                   ; version is 0 in the current spec
 version_format  = "{ 0x0 }" trace-id "{ 0x1 }" parent-id "{ 0x2 }" recorded
 trace-id        = 16BYTES
 parent-id       = 8BYTES
-trace-flags   = 1BYTE   ; only the least significant bit is counted
+trace-flags     = 1BYTE  ; only the least significant bit is used
 ```
 
 Unknown field identifier (anything beyond `0`, `1` and `2`) should be treated as
 invalid `traceparent`. All zeroes in `trace-id` and `parent-id` invalidates the
 `traceparent` as well.
 
-## `Traceparent` example
+## Serialization of `traceparent`
+
+Implementation SHOULD serialize fields into the field ordering sequence. In
+other words, `trace-id` field should be serialized first, `parent-id` second and
+`trace-flags` - third.
+
+Field identifiers should be treated as unsigned byte numbers and should be
+encoded in big-endian bit order.
+
+When `trace-id` is represented as a byte array - first element of an array MUST
+be copied first. When array is represented as a memory block of 16 bytes -
+serialization of `trace-id` would be identical to `memcpy` method call on that
+memory block. Same applies to `parent-id` field.
+
+If padding of the field is required (`traceparent` needs to be serialized into
+the bigger buffer) - any number of bytes can be appended to the end of the
+serialized value.
+
+## De-serialization of `traceparent`
+
+Let's assume the algorithm takes a buffer and can set and shift cursor in the
+buffer as well as validate whether the end of the buffer was reached or will be
+reached after reading the given number of bytes. De-serialization of
+`traceparent` should be done in the following sequence:
+
+1. If buffer is empty - return invalid status `BUFFER_EMPTY`. Set a cursor to
+   the first byte.
+2. Read the `version` byte at the cursor position. Shift cursor to `1` byte.
+3. If all three fields (`trace-id`, `parent-id`, `trace-flags`) already read -
+   return them with the status `OK` if `version` is `0` or status
+   `DOWNGRADED_TO_ZERO` otherwise.
+4. If at the end of the buffer return invalid status `TRACEPARENT_INCOMPLETE`.
+   Otherwise read the field identifier byte at the cursor position. Field
+   identifier should be read as unsigned byte assuming big-endian bits order.
+    1. If `0` - check that remaining buffer size is more or equal to `16` bytes.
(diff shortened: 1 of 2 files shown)


spec/Binary/20-BINARY_FORMAT.md
```diff
@@ -1,3 +1,54 @@
...
+
+The field `traceparent` encodes the version of the protocol and fields
+`trace-id`, `span-id` and `recorded`. Each field starts with the one byte field
+identifier with the field value following immediately after it. Field
+identifiers are used as markers for additional verification of the value
+consistency and may be used in future for the versioning of the `traceparent`
+field.
+
+``` abnf
+traceparent     = version "0" trace-id "1" parent-id "2" recorded
```
watson: I'm tempted to leave out the field identifiers 🤔

In the HTTP header format the `-` between each field doesn't individually identify each field and is to my understanding primarily there to make the value more readable - a use-case that's not required for binary data to the same extend.

What would be a real world example where the field identifiers would save us from parsing bad data?

watson: The `version` field should probably be defined above just to make it clear that it's one byte. I suggest updating this to be similar to how it's defined in the HTTP header spec ([2.2.2 Field value](https://w3c.github.io/trace-context/#field-value)), where there is a `value` field specifying the entire binary format consisting of a `version` and a `version-format` and then the `version-format` is defined on its own line.

spec/Binary/20-BINARY_FORMAT.md
```diff
@@ -1,3 +1,54 @@
...
+``` js
+{0,
+  0, 75, 249, 47, 53, 119, 179, 77, 166, 163, 206, 146, 157, 0, 14, 71, 54,
+  1, 52, 240, 103, 170, 11, 169, 2, 183,
+  2, 1}
+```
+
+This corresponds to:
+
+- `traceId` is
```
watson: Consider unifying how fields are referenced (`trace-id` vs `traceId` etc)

spec/Binary/20-BINARY_FORMAT.md
```diff
@@ -1,3 +1,54 @@
...
+
+## `tracestate` binary format
+
+List of up to 32 name-value pairs. Each list member starts with the 1 byte field
+identifier `0`. The format of list member is a single byte key length followed
+by the key value and single byte value length followed by the encoded value.
+
+``` abnf
+tracestate      = list-member 0*31( list-member )
+list-member     = "0" key-len key value-len value
```
watson, line 117: Similar to my comment about traceparent, I'm not sure if a field identifier is needed

watson (commented):
Great to see this getting started! 💯

spec/Binary/20-BINARY_FORMAT.md
```diff
@@ -1,3 +1,54 @@
...
+  `{75, 249, 47, 53, 119, 179, 77, 166, 163, 206, 146, 157, 0, 14, 71, 54}` or
+  `4bf9273577b34da6a3ce929d000e4736`.
+- `spanId` is `{52, 240, 103, 170, 11, 169, 2, 183}` or `34f067aa0ba902b7`.
+- `traceOptions` is `1` with the meaning `recorded` is true.
+
+## `tracestate` binary format
+
+List of up to 32 name-value pairs. Each list member starts with the 1 byte field
+identifier `0`. The format of list member is a single byte key length followed
+by the key value and single byte value length followed by the encoded value.
```
yurishkuro, line 112: One byte for value length limits the value length to 256, is that what we want?

spec/Binary/20-BINARY_FORMAT.md
```diff
@@ -1,3 +1,54 @@
...
+
+``` abnf
+tracestate      = list-member 0*31( list-member )
```
yurishkuro, line 116: There should be a number recording the number of items in the list.

yurishkuro:
Why not define the binary format using protobuf IDL instead of inventing some custom encoding?

SergeyKanzhelev:
@yurishkuro 

> Why not define the binary format using protobuf IDL instead of inventing some custom encoding?

I got a recommendation to not define it as protobuf as it's proprietary protocol/tech. That's said proposed format is basically how it will look in protobuf.

spec/Binary/20-BINARY_FORMAT.md
SergeyKanzhelev: It is aligned with the currently defined limits for tracestate. Also I'd prefer to have it defined as a `byte` instead of `varint` as `varint` will very rarely used - only for very long tracestate values while it's significantly more complex to implement

spec/Binary/20-BINARY_FORMAT.md
SergeyKanzhelev: filed identifiers may help to make fields optional in future and version the protocol. There is a section in http how to try to parse `traceparent` of a higher version - we need a similar section here. Kind of a best effort parsing.

Basically it's a proven technique from the protobuf that doesn't cost too much

SergeyKanzhelev: makes sense. Good idea

yurishkuro:
> I got a recommendation to not define it as protobuf as it's proprietary protocol/tech. That's said proposed format is basically how it will look in protobuf.

You just got another recommendation :-)  Should we have a vote then?

There are many binary format specs out there. I think it would be best if we _pick one_ rather than come up with a custom n+1 spec. There's nothing wrong with having the format defined by a widely used spec imo, it does not force the implementation in any way.

spec/Binary/20-BINARY_FORMAT.md
SergeyKanzhelev: oh, thanks for noticing

spec/Binary/20-BINARY_FORMAT.md
SergeyKanzhelev: Yeah. Perhaps termination symbol would be better in this case. So implementation can write and calculate limits the same time and stop at any moment. Would this solutions work better you think?

SergeyKanzhelev: Calculating the length ahead of time may slow down algorithms. Also the number of potential elements is small so pre-allocating of space on deserialization is not such a big problem. So suggestion is to not have length in the beginning and only use termination characters when buffer give is too large.

SergeyKanzhelev:
>You just got another recommendation :-) Should we have a vote then?

Before voting - do you agree that we should provide binary encoding rules that basically copies proto algorithm as not everybody can take dependency on proto? 


Also I tried to do it with proto and it feels like an overkill. For instance, instead of `byte` length that is easy to read we will get `varint` with complex algorithm of serializing/deserializing.

What other binary format specs you can suggest?

SergeyKanzhelev:
# PR was moved

This PR was moved here: https://github.com/w3c/trace-context-binary/pull/2

PR #217, opened 2018-12-14
AMQP draft: PR was moved to different repository

SergeyKanzhelev:
# PR was moved

https://github.com/w3c/trace-context-amqp/pull/1

### DIFF
--- a/spec/AMQP/20-AMQP_FORMAT.md
+++ b/spec/AMQP/20-AMQP_FORMAT.md
@@ -117,5 +117,16 @@ truncating or reusing some fields.
 
 ## Batching
 
-TODO: Batching of messages – recommended semantics of trace context fields when
-batching is happening. Should anything be said about it?
+The specification defines a binding to the AMQP "message format 0", i.e. the
+layout that is defined in the Messaging section of the the AMQP spec, but does
+not provide any further detail on HOW those messages are being transferred.
+
+AMQP is enormously efficient when filling link credit with sending messages in
+sequence without any special overlaid batching. And is often used without
+batching.
+
+Message brokers like Azure Service Bus, Azure Event Hub and others have a
+special message format that can batch multiple AMQP messages into one. However
+the way to bind to that format is by specifying a binding to the individual
+messages and then have the proprietary batching model pick that up via the
+binding to the standard message.
--- a/spec/AMQP/20-AMQP_FORMAT.md
+++ b/spec/AMQP/20-AMQP_FORMAT.md
@@ -1,33 +1,118 @@
 # AMQP format
 
-The AMQP binding will define for traceparent and tracestate to be added to the
-message in the application-properties section (as “w3c:tracestate” and
-“w3c:traceparent”) if they are added by the original message publisher. Once the
-message has been created, it’s no longer permissible to edit the bare message,
-so if it were necessary to annotate the message inside the middleware as it
-transits, that will happen in the “message-annotations” section, using the
-“application-properties” as a base. The ultimate receiver can then pick between
-the full context including infrastructure and the pure app context by either
-looking into “message-annotations” or “application-properties”.
- 
-1. *Protocol justification*. I suggest just reference W3C specs as they did in cloud events. And keep names of “headers”.
-2. *Size-sensitive environments*. There is a good reason to keep the format and sizes unchanged. For the size-sensitive environments
-If you ONLY INITIATE the message outside of existing context – there may be a possibility to save on size by truncating fields.
-Discuss how to use binary protocol instead of textual.
-Differences with message-id. Discuss how message-id is a deduplication concept where Trace-Context identifies the context of worker who either put or read the message.
-Batching of messages – recommended semantics of TraceContext fields when batching is happening.
+The Advanced Message Queuing Protocol (AMQP) is an open internet protocol for
+business messaging. It defines a binary wire-level protocol that allows for the
+reliable exchange of business messages between two parties. AMQP can be used as
+a protocol for asynchronous communication between components of an application.
+From the distributed tracing and telemetry correlation perspective it is a
+known problem to be able to correlate a component that placed message and
+component that processed it later. This specification describes how trace
+context MUST be encoded into AMQP messages.
 
+## Trace context fields placement in a message
 
-Failed post or get – what is relationship between failed operation trace
-   context and message trace context.
-Log entire app properties to get correlation.
+AMQP defines message as a payload with the additional annotations sections.
+There are two annotation sections this specification refers -
+"application-properties" and "message-annotations". See
+[3.2](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format)
+of OASIS Advanced Message Queuing Protocol (AMQP) Version 1.0, Part 3:
+Messaging.
 
-Mutation of trace context by queue provider – should we allow/recommend to mutate trace context fields by trace provider. Especially for the span-id.
-   Security/privacy consideration – we can refer to W3C spec again. And adjust
-   it if needed.
+AMQP message section "application-properties" is immutable collection of
+properties that is defined by message publisher and can be read by the message
+consumer. Message brokers cannot mutate those properties. Section
+"message-annotations" is designed for message brokers to use and can be mutated
+during the message processing.
 
-application properties is "unmodifiable" so "message-annotations" overrides
-those in application
+Fields `traceparent` and `tracestate` SHOULD be added to the message in the
+application-properties section by message publisher. Once the message has been
+created, it’s no longer permissible to edit the bare message. So if it were
+necessary to annotate the message inside the middleware as it transits, that
+MUST happen in the “message-annotations” section, using the
+“application-properties” as a base. Message reader SHOULD construct the full
+trace context by reading `traceparent` and `tracestate` fields from the
+“message-annotations” first and if not exist - from “application-properties”.
 
-traceparent 1.6.22 (list of binary) or binary - make rationale
-tracestate 1.6.23 (map of string/string type)
+## Trace context and failed read/write operations
+
+This specification defines how to propagate context from publisher thru broker
+to ultimate reader. It does not define any additional transport level correlation
+constructs that can be used to investigate failed publish or read operations.
+
+It is recommended, however, for AMQP implementations to make the best effort
+attempt to read the trace context from the message and use it while reporting
+such problems.
+
+## `traceparent` AMQP format
+
+The field `traceparent` MUST be encoded and decoded using [binary
+protocol](..\extension-binary.html) and stored as a binary type defined in
+section
+[1.6.19](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-binary)
+of AMQP specification.
+
+Property name MUST be `traceparent` - all lowercase without delimiters.
+
+## `tracestate` AMQP format
+
+The field `tracestate` MUST be encoded and decoded as a string to string map.
+See definition of type map in section
+[1.6.23](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-map)
+of AMQP specification.
+
+---
+
+*NOTE about the strings encoding*
+
+Please note that strings are defined as UTF-8 in AMQP. Should anything be
+different from the HTTP encoded opaque strings? Should we allow to benefit from
+wider character set for better encoding of opaque values? Or this will be too
+error prone?
+
+---
+
+Property name MUST be `tracestate` - all lowercase without delimiters.
+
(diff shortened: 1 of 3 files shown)


spec/AMQP/20-AMQP_FORMAT.md
```diff
@@ -1,3 +1,118 @@
...
+properties that is defined by message publisher and can be read by the message
+consumer. Message brokers cannot mutate those properties. Section
+"message-annotations" is designed for message brokers to use and can be mutated
+during the message processing.
+
+Fields `traceparent` and `tracestate` SHOULD be added to the message in the
+application-properties section by message publisher. Once the message has been
+created, it’s no longer permissible to edit the bare message. So if it were
+necessary to annotate the message inside the middleware as it transits, that
+MUST happen in the “message-annotations” section, using the
```
lmolkova, line 31: so the trace context MAY be added to the message-annotations or application-properties depending on the way it was added. Assuming message-annotations could be accessed and changed at any moment,  why do we need to use application-properties at all?
it creates a bit of extra complexity and perf impact.

spec/AMQP/20-AMQP_FORMAT.md
```diff
@@ -1,3 +1,118 @@
...
+## Size sensitive environments
+
+TODO: There are many good reason to keep the format and sizes of fields
+unchanged. For the size-sensitive environments, if you ONLY INITIATE the message
+outside of existing context – there may be a possibility to save on size by
+truncating or reusing some fields.
+
+## Batching
+
+TODO: Batching of messages – recommended semantics of TraceContext fields when
```
lmolkova: I think we should have a discussion and general guidance for batching, but it probably is related more to the instrumentation and tracing system behavior.

My (very small) experience with AMQP involves Azure ServiceBus and EventHubs, I can miss a lot of details here, but my take on this was:

- SB/EH message could be represented by AMQP message. Or a batch of SB/EH messages is a single AMQP message. If it is the typical case for AMQP, **we should give guidance to the AMQP implementations on how to treat these two cases**.

- every  SB/EH message carry trace context - they probably were produced in different contexts under different spans. Such messages should be annotated during creation, not during publishing. This is instrumentation guidance.

- AMQP message constructed from the batch has new trace context, potentially unrelated to each message in the batch. Relation should be established by something like opencensus links. But again  this is instrumentation and tracing system details, not the protocol.

spec/AMQP/20-AMQP_FORMAT.md
SergeyKanzhelev: I've added rationale section for this. My understanding that it is desire-able and easier to use app properties. So message annotations is simply a fallback option.

spec/AMQP/20-AMQP_FORMAT.md
```diff
@@ -1,3 +1,121 @@
...
+during the message processing.
+
+Fields `traceparent` and `tracestate` SHOULD be added to the message in the
+application-properties section by message publisher. Once the message has been
+created, it’s no longer permissible to edit the bare message. So if it were
+necessary to annotate the message inside the middleware as it transits, that
+MUST happen in the “message-annotations” section, using the
+“application-properties” as a base. Message reader SHOULD construct the full
+trace context by reading `traceparent` and `tracestate` fields from the
+“message-annotations” first and if not exist - from “application-properties”.
```
calohmn, line 34: We have been looking into where to put tracing context information in a message as well (in the context of adding OpenTracing support to Eclipse Hono and to EnMasse) and have been discussing whether to use the "message-annotations" section or the "delivery-annotations" for that. We chose the "delivery-annotations" - see the discussion here: EnMasseProject/enmasse#1212.

One point here was about the part of the AMQP spec where it is said that "message annotations SHOULD be propagated across every delivery step".
There might be cases where such a requirement looks useful, but in other cases it might be desirable for a component to not forward tracing info any further - to not propagate tracing info to _external_ application level message consumers.

Have you considered using delivery-annotations instead of message-annotations so that the propagation of tracing context information can be defined on a "per hop" basis? 
Thereby tracing context information could either be directly associated with the message (via the immutable application-properties) or  - to override this - via delivery-annotations with the flexibility of either propagating it further or not.

spec/AMQP/20-AMQP_FORMAT.md
yurishkuro: @SergeyKanzhelev - I am not quite convinced by that rationale. Having to look for trace context in two different locations and support some sort of overlay logic seems overly complex & unnecessary. What benefits does it provide?

spec/AMQP/20-AMQP_FORMAT.md
SergeyKanzhelev: My understanding is that "message-annotation" are overridable. So you getting benefit of them being propagated, but still can edit them or remove them when needed. Do you know what other isues "message-annotations" may bring? Or my understanding of "message-annotations" is incorrect?

@clemensv I'd appreciate if you can chime in here.

spec/AMQP/20-AMQP_FORMAT.md
SergeyKanzhelev: My understanding is that applicaiton properties is easier to implement and would work fine in all cases, except when broker need to change this context and they cannot change applicaiton properties according to spec.

SergeyKanzhelev: I've unresolved conversation to keep feedback coming

spec/AMQP/20-AMQP_FORMAT.md
clemensv: Several thoughts on this. 

First, I like the integration as it stands and I think this is good overall. 

On the application-properties vs. message-annotations vs. delivery-annotations question, I'm where @rgodfrey stood on the issue that @calohmn quoted: https://github.com/EnMasseProject/enmasse/issues/1212#issuecomment-396868699

I think manipulations of the context made during routing ought to go into the message annotations.

I think there are three grand scenarios:

- Application trace context through a simple AMQP brokered entity. If you send a message into a queue from an app and then pull that message from the queue with the same or another app and nothing else interesting happens, application-properties is the right place, because you're not doing any tracing of the broker, you're just passing the context alongside the message.

- Application context plus routing context. If you have a chain of AMQP entities as you might have in a complex routing scenario, you might want to enrich the context as the message is being passed on. In that case, the prior scenario still exists, i.e. you will still want the option to have the context information as set by the message producer and use that for application level tracing unencumbered by whatever the routing layer might be tracking. As the routing layer gets it hand on the message for tracing, it basically forks the context off into the message-annotations and that's being manipulated on the route. As the message reaches the destination, you now have the app-level context from the producer in the application-properties AND the routing context that was layered on top of that in the message-annotations.

- Hop to hop tracing: A delivery-annotations usage scenario would be purely additive to either of the prior two, allowing any intermediary to spawn a new context for itself or to propagate a context only via subset of hops. I think that is fairly esoteric at this point and I wouldn't take that into a spec unless someone shows a hard use-case.

SergeyKanzhelev: Added verbatim to rationale document and resolving this conversation. @calohmn feel free to re-open

yurishkuro (commented):
Has this been implemented anywhere?

spec/AMQP/20-AMQP_FORMAT.md
clemensv: The specification should define a binding to the AMQP "[message format 0](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#definition-MESSAGE-FORMAT)", i.e. the layout that is defined in the [Messaging section of the the AMQP spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format), but should not go into further detail on HOW those messages are being transferred. 

AMQP is enormously efficient when filling link credit with sending messages in sequence without any special overlaid batching. 

Service Bus and Event Hubs indeed have a special message format that can indeed batch multiple AMQP message format messages into one, but the way to bind to that format is by specifying a binding to the core message format and then have the proprietary batching model pick that up via the binding to the standard message.

SergeyKanzhelev:
@yurishkuro I'm working with Azure IoT team to implement it. Binary format for `traceparent` was already adopted by Open Census for GRPC.

spec/AMQP/20-AMQP_FORMAT.md
```diff
@@ -1,3 +1,132 @@
...
+
+## `traceparent` AMQP format
+
+The field `traceparent` MUST be encoded and decoded using [binary
+protocol](..\extension-binary.html) and stored as a binary type defined in
+section
+[1.6.19](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-binary)
+of AMQP specification.
+
+Property name MUST be `traceparent` - all lowercase without delimiters.
```
calohmn, line 57: Considering the case that the `traceparent` and `tracestate` fields are transported via _message-annotations_: 

As for the base type of the message-annotations, the AMQP spec says (http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-annotations):

> The annotations type is a map where the keys are restricted to be of type symbol or of type ulong. All ulong keys, and all symbolic keys except those beginning with "x-" are reserved. Keys beginning with "x-opt-" MUST be ignored if not understood. On receiving an annotation key which is not understood, and which does not begin with "x-opt", the receiving AMQP container MUST detach the link with a not-implemented error.

Concerning the names of `message-annotations`:

> A registry of defined annotations and their meanings is maintained [[AMQPMESSANN](http://www.amqp.org/specification/1.0/message-annotations)].

The question here: Are the names of the `traceparent` and `tracestate` fields supposed to be somehow registered as reserved with OASIS. Otherwise it looks to me as if the prefix "x-opt-" should be used.

spec/AMQP/20-AMQP_FORMAT.md
```diff
@@ -1,3 +1,132 @@
...
+
+## `tracestate` AMQP format
+
+The field `tracestate` MUST be encoded and decoded as a string to string map.
+See definition of type map in section
+[1.6.23](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-map)
+of AMQP specification.
```
calohmn, line 64: Considering the case that the `traceparent` and `tracestate` fields are transported via _application-properties_: 

About application-properties, the AMQP spec says (http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-application-properties):
> The keys of this map are restricted to be of type string (which excludes the possibility of a null key) and the values are restricted to be of simple types only, that is, excluding map, list, and array types. 

This looks incompatible with wanting to store a `tracestate` AMQP map in there.

spec/AMQP/21-AMQP_FORMAT_RATIONALE.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Binary `traceparent` vs. list of binary values
+
+There are multiple ways to implement a `traceparent`. It can either be
+implemented as string (http-like), binary protocol (like for grpc) or list of
+separate binary values (`trace-id`, `parent-id`, `trace-flags`).
+
+Strings duplicating the size of a field, using list of binaries will require to
+redefine the way the field serialized, parsed, and versioned. So re-using binary
+protocol looks like a logical solution.
```
grs, line 21: Where is that 'binary protocol' defined? (I found https://w3c.github.io/trace-context/extension-binary.html but it seems to be empty?). AMQP has quite a complex type-system that will already be required to decode the message, so it might be simpler to use that to encode traceparent in some way rather than requiring further decoding of the field.

SergeyKanzhelev: It is in another PR: https://github.com/w3c/trace-context/pull/215 Sorry for confusion.

@grs what type would you suggest for the header? We discussed options and didn't find the right match

SergeyKanzhelev:
@calohmn thank you for your review! I'll look into it. @clemensv I'd appreciate your input here as well.

spec/AMQP/21-AMQP_FORMAT_RATIONALE.md
grs: Actually, to @calohmn's point re the restrictions on application-property values, perhaps you are right and the binary format is best (it is not very arduous to parse). (Alternatively an amqp list or a map with integer keys would be similar to the binary encoding).

SergeyKanzhelev: @grs yeah. It may be better to switch to binary protocol than. At least some consistency - either text (like HTTP or MQTT) or binary (like GRPC or AMQP).

SergeyKanzhelev:
# PR was moved

https://github.com/w3c/trace-context-amqp/pull/1

Issue #203, opened 2018-11-23
abstract needs editing


dret:
the abstract has been written with the old title and spec name in mind. it needs some editing to make more sense for the new title/identity.

SergeyKanzhelev:
closed via #246

PR #221, opened 2018-12-19
[WIP] Changed abstract to define the problem we are solving and the proposed solution

AloisReitbauer:
Still WIP. Create PR for communication

addresses #203

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ bin
 
 # Vim
 .swp
+
+#VSCode
+.vscode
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
> Still WIP. Create PR for communication

please let know when it's ready for review

SergeyKanzhelev:
Any updates on this PR?

SergeyKanzhelev:
Some feedback - I'd suggest to keep a reference to other specs in the abstract. Something like 

> To correlate events between components, these components need to exchange and store a piece of information called context. Typically context consists of an originating event identifier, an originating component identity and other event properties. This specification describes how event identifiers should be formed and used to correlate events. Other specifications cover context propagation aspects like user-defined properties propagation, storage format, etc.

So specification will be scoped to a specific problem.

SergeyKanzhelev:
I think the minimum we need to make to address #203 is to make it clear that correlation context is not part of this spec. This sentence:

> Correlation context carries user-defined properties.

All other wording updates may happen in the next milestone

danielkhan:
Closed in favor of https://github.com/w3c/trace-context/pull/259

PR #242, opened 2019-01-11
[WIP] Editorial changes to HTTP Header Format

### LINKED ISSUE #168: Value format section is confusing
Working to add support for the tracestate header, I found this to be confusing:
https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#value-format

Main confusions:
* Do we always have to propagate the traceparent in the tracestate? Based on the examples this seems to be the case, but I think this should not be the default.
* Propagating the traceparent in the tracestate is a vendor specific implementation and probably should not be added in the public (official) examples. Every vendor can choice to do this based on their need, for example propagating only the span-id.

### LINKED ISSUE #234: Decide, if reuse of last trace ID in traceparent needs to be mentioned in spec
In https://github.com/w3c/trace-context/blame/cr--edit-processing-model/spec/20-HTTP_HEADER_FORMAT.md#L51 we mention the reuse of the last trace ID in traceparent.
This looks like an implementation detail and we should discuss if this should go into the spec.

### LINKED ISSUE #238: Decide, if we should explain OWS though we refer to the spec where it's defined
In https://github.com/w3c/trace-context/blame/cr--edit-processing-model/spec/20-HTTP_HEADER_FORMAT.md#L312 we describe optional whitespace OWS though we refer to RFC7230 before. Should we replicate other specs in our spec or is the link sufficient?

danielkhan:
Fixes #238 
Fixes #168 
Fixes #234

### DIFF
--- a/spec/20-HTTP_HEADER_FORMAT.md
+++ b/spec/20-HTTP_HEADER_FORMAT.md
@@ -128,8 +128,6 @@ hexadecimal string.
 
 **Example:** `00f067aa0ba902b7`.
 
-**Note:** `parent-id` is sometimes also referred to as `span-id` as some
-telemetry systems call the execution of a client call a *span*.
 
 ##### Rules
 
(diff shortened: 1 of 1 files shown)


danielkhan:
Fixes #238 #168 #234

Issue #239, opened 2019-01-10
Header limits are obsolete and should be removed


danielkhan:
In https://github.com/w3c/trace-context/blame/cr--edit-processing-model/spec/20-HTTP_HEADER_FORMAT.md#L387 we describe the limits which still reflects the 512 character limit we agreed on before coming up with the 5 entry limit.
This is now described in https://github.com/w3c/trace-context/blame/cr--edit-processing-model/spec/20-HTTP_HEADER_FORMAT.md#L375
Should we remove the limits paragraph entirely?

SergeyKanzhelev:
Is it possible to close on this item separately from the big CR PR? Few considerations:

- Rationale doc should be updated with the discussions and reasoning
- There are two items that we need to make sure to address https://github.com/w3c/trace-context/issues/175 and https://github.com/w3c/trace-context/issues/109
- My understanding that the `MUST propagate 5` thing is still too aggressive of a requirement (Maybe even more aggressive than limiting to 512 characters if there are abuse. At least this is what we discussed in Amazon office during the KubeCon and should re-hash in rationale doc again.). Discussion was that SaaS/platform vendor may have its own limit published, which will make an incentive for tracing vendors to limit `tracestate` entries they require. Basically follow the @yurishkuro suggestion in #109

danielkhan:
ok - let's discuss that independently of the PR.

AloisReitbauer:
We cannot move to PR if we do not reach a decision on this topic. I propose we add the trimming procedure to the PR and add a note that size requirements are still under discussion. What is currently in the spec does not reflect the latest discussions, so it at least needs to be removed.

SergeyKanzhelev:
Let's clean up duplicates:

- [Can a tracer/application only carry trace-parent and drop trace-state?](https://github.com/w3c/trace-context/issues/218)
- [Explain/remove 512 limit of tracestate in spec](https://github.com/w3c/trace-context/issues/109)
- [Action on tracestate header larger than 512 is confusing](https://github.com/w3c/trace-context/issues/175)

Basically, we need to update rationale document with the logic of
decision and discussions we had. Suggest truncation strategies and
advice on the limits.

I'll close issues above. I'll sent a proposed update for this limit.

danielkhan:
Let's continue discussion in #248

danielkhan:
Let's continue discussion in #248

SergeyKanzhelev:
[...]

PR #248, opened 2019-02-08
Make tracestate limit the responsibility of a platform

SergeyKanzhelev:
Proposed update for the `tracestate` limit. Basically make it a platform responsibility. See #239

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -392,7 +392,7 @@ tracing system may propagate them as they came, combine into a single
 header or split into multiple headers differently following the RFC
 specification.
 
-Platforms and tracing vendors SHOULD propagate at lease 512 characters
+Platforms and tracing vendors SHOULD propagate at least 512 characters
 of a combined header. This length includes commas required to separate
 list items. But does not include optional white space (`OWA`)
 characters.
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -404,7 +404,7 @@ propagating `tracestate` SHOULD be weighted against the value of
 monitoring scenarios enabled for the end users.
 
 In situation when `tracestate` needs to be truncated due to size
-limitations platform of tracing vendor MUST truncate whole entries.
+limitations, platform of tracing vendor MUST truncate whole entries.
 Entries larger than `128` characters long SHOULD be removed first. Then
 entries SHOULD be removed starting from the end of `tracestate`. Note,
 other truncation strategies like safe list entries, blocked list entries or
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -407,7 +407,7 @@ In situation when `tracestate` needs to be truncated due to size
 limitations platform of tracing vendor MUST truncate whole entries.
 Entries larger than `128` characters long SHOULD be removed first. Then
 entries SHOULD be removed starting from the end of `tracestate`. Note,
-other truncation strategies like whitelist entries, blacklist entries or
+other truncation strategies like safe list entries, blocked list entries or
 size-based truncation MAY be used, but highly discouraged. Those
 strategies will decrease interoperability of various tracing vendors.
 
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -383,23 +383,32 @@ Rather, the entry would be rewritten to only include the most recent position:
 
 **Limits:**
 
+The `tracestate` field contains essential information for request
+correlation. Platforms and tracing systems MUST propagate this field.
 There might be multiple `tracestate` headers in a single request
 according to [RFC7230 section
-3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2). Maximum length of a
-combined header MUST be less than 512 characters. This length includes commas
-required to separate list items. But SHOULD NOT include optional white space
-(OWA) characters.
+3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2). Platform or
+tracing system may propagate them as they came, combine into a single
+header or split into multiple headers differently following the RFC
+specification.
 
-`tracestate` field contains essential information for request correlation.
-Platforms and tracing systems MUST propagate this header. Compliance with the
-specification will require storing of `tracestate` as part of the request
-payload or associated metadata. Allowing the long field values can make
-compliance to the specification impossible. Thus, the aggressive limit of 512
-characters was chosen.
+Platforms and tracing vendors SHOULD propagate at lease 512 characters
+of a combined header. This length includes commas required to separate
+list items. But does not include optional white space (`OWA`)
+characters.
 
-If the `tracestate` value has more than 512 characters, the tracer CAN decide to
-forward the `tracestate`. When propagating `tracestate` with the excessive
-length - the assumption SHOULD be that the receiver will drop this header.
+There are systems where propagating of 512 characters of `tracestate`
+may be expensive. In this case the maximum size of propagated
+`tracestate` header SHOULD be documented and explained. Cost of
+propagating `tracestate` SHOULD be weighted against the value of
+monitoring scenarios enabled for the end users.
+
+In situation when `tracestate` MUST be truncated due to size limitations
+platform of tracing vendor SHOULD truncate whole list entries starting
+from the end of tracestate. Note, other truncation strategies like
+whitelist entries, blacklist entries or size-based truncation MAY be
+used, but highly discouraged. Those strategies will decrease
+interoperability of various tracing vendors.
 
 ## Examples of HTTP headers
 
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -403,12 +403,13 @@ may be expensive. In this case the maximum size of propagated
 propagating `tracestate` SHOULD be weighted against the value of
 monitoring scenarios enabled for the end users.
 
-In situation when `tracestate` MUST be truncated due to size limitations
-platform of tracing vendor SHOULD truncate whole list entries starting
-from the end of tracestate. Note, other truncation strategies like
-whitelist entries, blacklist entries or size-based truncation MAY be
-used, but highly discouraged. Those strategies will decrease
-interoperability of various tracing vendors.
+In situation when `tracestate` needs to be truncated due to size
+limitations platform of tracing vendor MUST truncate whole entries.
+Entries larger than `128` characters long SHOULD be removed first. Than
+entries SHOULD be removed starting from the end of `tracestate`. Note,
+other truncation strategies like whitelist entries, blacklist entries or
+size-based truncation MAY be used, but highly discouraged. Those
+strategies will decrease interoperability of various tracing vendors.
 
 ## Examples of HTTP headers
 
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -405,7 +405,7 @@ monitoring scenarios enabled for the end users.
 
 In situation when `tracestate` needs to be truncated due to size
 limitations platform of tracing vendor MUST truncate whole entries.
-Entries larger than `128` characters long SHOULD be removed first. Than
+Entries larger than `128` characters long SHOULD be removed first. Then
 entries SHOULD be removed starting from the end of `tracestate`. Note,
 other truncation strategies like whitelist entries, blacklist entries or
 size-based truncation MAY be used, but highly discouraged. Those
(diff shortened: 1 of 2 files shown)


AloisReitbauer (changes requested):
Needs language polishing

bogdandrutu (approved):

bogdandrutu:
Looks good the message overall. I am not checking for the language.

danielkhan:
I am not sure about it and I am happy to discuss this but isn't that a lot of ambiguity and reasoning around our thought-process for a spec?

AloisReitbauer (changes requested):
This is not what we defined in the last meeting. Following this approach one entry can have 512 bytes. I would define a maximum size per entry and total number of entries. 

We should make it clear that dropping tracestate will break other tracing systems.

AloisReitbauer:
This was the last decision made during TPAC:

Decision: Size limit
* Every entry limit of 128 bytes (including key). If violated it’s ALLOWED to evict a violating entry.
* Max number of entries is 5. If that’s violated, it’s ALLOWED to evict entries that exceed the limit (but only at the end!)
* There won’t be an overall limit of 512 anymore.
* An implementer is does NOT HAVE TO enforce those rules

yurishkuro:
> Every entry limit of 128 bytes (including key)

any written records of justification?

AloisReitbauer:
* Typical tracestate entries are between 36 and 77 bytes (according to #109)! In 512 bytes, this would allow between (6X and 14X tracers).
* To allow some head room, let's assume a entry-length of ~128 bytes. (would allow 4X tracers)
* Assuming, the most common number of tracers is 2 (APM vendor + Cloud provider). There might be cases of multiple APM vendors + multiple Cloud providers, but cases of 3/4 tracing systems will be the exception.
* "Edge" devices would usually only have one tracestate entry (since they initiate a call)

spec/20-http_header_format.md
```diff
@@ -383,23 +383,32 @@ Rather, the entry would be rewritten to only include the most recent position:
...
-
-If the `tracestate` value has more than 512 characters, the tracer CAN decide to
-forward the `tracestate`. When propagating `tracestate` with the excessive
-length - the assumption SHOULD be that the receiver will drop this header.
+3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2). Platform or
+tracing system may propagate them as they came, combine into a single
+header or split into multiple headers differently following the RFC
+specification.
+
+Platforms and tracing vendors SHOULD propagate at lease 512 characters
```
discostu105: at "least"

AloisReitbauer: just copied over from the notes :-)

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,92 @@ configuration - certain header mutations may still be required. An example may b
...
+like positioning request in multiple distributed tracing graphs and
+providing backward compatibility with older protocols. But `tracestate`
+is not limited to this use and will drive innovations and new scenarios
+going forward.
+
+Opaque nature of `tracestate` value raises many questions and biggest
+one is what guarantees we can request from protocol implementors on
+propagating and storing this field.
+
+On one hand - field value should be small so implementors can satisfy
```
discostu105: "On the one hand, "

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,92 @@ configuration - certain header mutations may still be required. An example may b
...
+the requirement to pass the value all the time. It is especially
+critical for various messaging systems where metadata like `tracestate`
+is not paid for by the customer and provided for free.
+
+On other hand - in order for the field to be useful there should be some
```
discostu105: "On the other hand, "

yurishkuro:
> * Typical tracestate entries are between 36 and 77 bytes

@wu-sheng is that true of SkyWalking? You were telling me that you may pass the caller (name+endpoint) in the context, which can easily be much longer strings.

SergeyKanzhelev:
@AloisReitbauer I explained in rationale portion of this PR why any arbitrary limits are not helpful. **TL;DR;** any limit spec will define will be too big for somebody to propagate and too small to somebody to use. Also limits defined as hard number of characters will not encourage vendors to minimize amount of information to propagate via `tracestate` and to limit this information to essentials only. Cloud vendors in their turn has a good incentive of providing better monitoring experience for customers to increase the amount of data they will propagate.

I understand that the proposal you've listed was discussed during Leon meeting, but alternative proposal from @yurishkuro that is reflected in this PR was suggested even earlier. And this is what we discussed later during KubeCon Seattle.

More from rationale doc in this PR:

> #### 2. Limit based on number of entries

> This proposal was made to eliminate the problem of a "greedy" vendor utilizes the full length of a `tracestate` and ultimately blocking the interoperability.

> Problem with this proposal is that limits proposed for the individual entry were still quite high for cloud providers.

> Also even though seemingly this proposal makes interoperability better, abuse of using multiple entries by a single vendor is still unavoidable.


>  #### 3. Make max length the decision of the implementor

> Header should be small so providers can satisfy the requirement to pass the value all the time.	
> Options above "protected" vendors from implementors who will take optional nature of `tracestate` and will not propagate `tracestate`. However the reality is that for many platforms even larger `tracestate` fields propagation is not a big issue.


> 512 bytes looks like a reasonable compromise.	So there was a proposal that the spec will suggest the length limits, and define the algorithm of trimming `tracestate`. However the final decision of an actual implementation to the platform.

spec/20-http_header_format.md
SergeyKanzhelev: thx!

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,103 @@ configuration - certain header mutations may still be required. An example may b
...
+### Total size limit
+
+The field `tracestate` opens up rich interoperability and extensibility
+scenarios for vendors. There are defined use cases for the `tracestate`
+like positioning request in multiple distributed tracing graphs and
+providing backward compatibility with older protocols. But `tracestate`
+is not limited to this use and will drive innovations and new scenarios
+going forward.
+
+Opaque nature of `tracestate` value raises many questions and biggest
```
mtwo: Replace with:

The opaque nature of `tracestate` introduces some tension between guarantees to vendors around propagating vendor-specific data with traces vs. storage and propagation constraints

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,103 @@ configuration - certain header mutations may still be required. An example may b
...
+tempting and hard to implement feature.
+
+Talking about abuse - one thing this working group is doing is working
+on user-defined context propagation as part of [Correlation
+Context](https://w3c.github.io/correlation-context/) spec. Correlation
+context will provide a relieve valve for people who want to have freedom
+propagating relatively big payload across the components of a
+distributed trace.
+
+Specification allows removing and adding as many `tracestate` entries as
```
mtwo: Replace the first sentence with "This specification allows removing and adding as many `tracestate` entries as an implementation needs.

mtwo (approved):
Two grammar nits. Please fix before submitting, but I'm approving the changes.

danielkhan:
I found a few grammar problems like that one here:
> Options above "protected" vendors from implementors who will take
optional nature of `tracestate` and will not propagate `tracestate`.
However the reality is that for many platforms even larger `tracestate`
fields propagation is not a concern.

At the very least, I suggest to rephrase it to

> The options above "protected" vendors from implementors that will take
the optional nature of `tracestate` and will not propagate it.
However, the reality is that for many platforms propagating an even larger `tracestate` field is not a concern.

Still, it could be further edited down for clarity and conciseness.

How do we want to deal with that? Go with it and do a review after? But then we will again end up re-hashing discussions we had already.

SergeyKanzhelev:
@mwear, @victorNewRelic can you please review from NewRelic perspective?

@yurishkuro any more concerns?

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,103 @@ configuration - certain header mutations may still be required. An example may b
...
+distributed trace.
+
+This specification allows removing and adding as many `tracestate`
+entries as an implementation needs. This freedom is required to satisfy
+privacy and interoperability concerns. Thus all suggestions
+specification is making regarding the size of `tracestate` field will
+only be a recommendation that would improve vendors interoperability and
+cannot be "enforced" in practice.
+
+Possible solutions:
```
yurishkuro: Are these "considered and rejected"? The caption is confusing.

mwear:
@SergeyKanzhelev: We are ok with the limits suggested here.

spec/21-http_header_format_rationale.md
SergeyKanzhelev: I addressed feedback. Also just a note, it's in rationale doc, not a spec

AloisReitbauer (approved):

spec/20-http_header_format.md
```diff
@@ -383,23 +383,33 @@ Rather, the entry would be rewritten to only include the most recent position:
...
+
+There are systems where propagating of 512 characters of `tracestate`
+may be expensive. In this case the maximum size of propagated
+`tracestate` header SHOULD be documented and explained. Cost of
+propagating `tracestate` SHOULD be weighted against the value of
+monitoring scenarios enabled for the end users.
+
+In situation when `tracestate` needs to be truncated due to size
+limitations platform of tracing vendor MUST truncate whole entries.
+Entries larger than `128` characters long SHOULD be removed first. Than
```
devinsba:
```suggestion
Entries larger than `128` characters long SHOULD be removed first. Then
```

plehegar:
(waiting on #260 for merge)

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,109 @@ configuration - certain header mutations may still be required. An example may b
...
+distributed trace.
+
+This specification allows removing and adding as many `tracestate`
+entries as an implementation needs. This freedom is required to satisfy
+privacy and interoperability concerns. Thus all suggestions
+specification is making regarding the size of `tracestate` field will
+only be a recommendation that would improve vendors interoperability and
+cannot be "enforced" in practice.
+
+There solutions to address this problem were discussed:
```
nikmd23:
```suggestion
Three solutions to address this problem were discussed:
```

spec/21-http_header_format_rationale.md
```diff
@@ -98,17 +90,109 @@ configuration - certain header mutations may still be required. An example may b
...
+#### 1. Declare the arbitrary max length
+
+This is simplest and easy to understand solution. Proposals for the
+limit varied from `20` (size of a single `parent-id` with small
+identifier) to `1024` to fit up to 5 vendors with large tracestate
+entries. `512` is a nice median length.
+
+Extremely small proposal like `64` wasn't received well by vendors,
+while larger limits was unacceptable/undesirable for cloud vendors as
+COGS of implementing it are quite high.
```
nikmd23:
```suggestion
costs of implementing it are quite high.
```

spec/20-http_header_format.md
```diff
@@ -383,23 +383,33 @@ Rather, the entry would be rewritten to only include the most recent position:
...
+characters.
+
+There are systems where propagating of 512 characters of `tracestate`
+may be expensive. In this case the maximum size of propagated
+`tracestate` header SHOULD be documented and explained. Cost of
+propagating `tracestate` SHOULD be weighted against the value of
+monitoring scenarios enabled for the end users.
+
+In situation when `tracestate` needs to be truncated due to size
+limitations platform of tracing vendor MUST truncate whole entries.
```
nikmd23: Pedantic: added missing comma

```suggestion
limitations, platform of tracing vendor MUST truncate whole entries.
```

spec/20-http_header_format.md
```diff
@@ -383,23 +383,33 @@ Rather, the entry would be rewritten to only include the most recent position:
...
+Entries larger than `128` characters long SHOULD be removed first. Then
+entries SHOULD be removed starting from the end of `tracestate`. Note,
+other truncation strategies like whitelist entries, blacklist entries or
```
nikmd23: whitelist and blacklist [can be problematic terms](https://github.com/graphite-project/graphite-web/issues/1569). In general, we should try stay away from them. 

```suggestion
other truncation strategies like safe list entries, blocked list entries or
```

nikmd23 (changes requested):
I generally approve, but recommend accepting my change about whitelist/blacklist terminology.

SergeyKanzhelev:
@nikmd23 thanks for review

nikmd23 (approved):

Issue #247, opened 2019-02-06
Mark "Other Protocols" section non-formative and simply point to the registry


SergeyKanzhelev:
Section https://w3c.github.io/trace-context/#other-protocols is going to be updated and changed going forward as new protocols got supported.

@plehegar what's a good practice here? Have a link to a folder of specs with other protocols (like a GitHub link) or mark section as non-formative and update it even after spec was marked CR? Any other options here?

nikmd23:
@plehegar has suggested that the spec link to a separate "registry" document, that can be updated more easily.

So essentially, the spec links to a separate document that is a "note". That note will link to other protocols. 

This makes sense to me, and allows us to provide guidance on other protocols without churning the spec or dealing with the W3C processes.

AloisReitbauer:
LGTM

plehegar:
Here is an initial draft: https://w3c.github.io/trace-context-protocols-registry/

danielkhan:
LGTM

nikmd23:
Thanks @plehegar. That looks great. Should we close this issue in favor of #254? @SergeyKanzhelev has listed out the work required to complete the registry there.

PR #249, opened 2019-02-08
Mark other protocols section point to the registry

SergeyKanzhelev:
#247 

and added link to CloudEvents extension.

This change will add the note "non-normative" to the spec. Supposedly it means that we will be able to keep updating it without the whole document versioning.

image

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/pull/249.html" title="Last updated on Feb 15, 2019, 12:45 AM UTC (e702be2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/249/2bd4746...e702be2.html" title="Last updated on Feb 15, 2019, 12:45 AM UTC (e702be2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -59,7 +59,7 @@
         
         <section data-include="spec/20-http_header_format.md" data-include-format='markdown'></section>
 
-        <section data-include="spec/40-other-protocols.md" data-include-format='markdown' class="informative"></section>
+        <section data-include="spec/40-other-protocols.md" data-include-format='markdown'></section>
 
         <section data-include="spec/50-privacy.md" data-include-format='markdown'></section>
         <section data-include="spec/51-security.md" data-include-format='markdown'></section>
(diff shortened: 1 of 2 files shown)


bogdandrutu (approved):

plehegar:
See #247 for an alternative approach.

plehegar:
Thinking about this further, is there now a reason not to make the link to the registry normative? Surely, we want to encourage implementation to follow what is in the registry rather than coming up with their own mapping. Imho, we should make the registry to document normative, and let it deal with the question on whether a particular mapping is normative or not. That's an other added bonus of haivng this indirection.

SergeyKanzhelev:
@plehegar makes sense. Changed

AloisReitbauer (approved):

AloisReitbauer:
@plehegar please approve if this is fine for you.

plehegar (approved):

Issue #211, opened 2018-11-30
Test case for invalid traceparent restart should validate that all children has the same ID


SergeyKanzhelev:
Spec tells to restart the trace if incoming traceparent has invalid `trace-id` (like all zeroes). Test case validates that child request will have a different trace ID. But doesn't validate that ALL children will share this trace ID. This may lead to implementations that restart traces lazily which is not desireable.

SergeyKanzhelev:
PR was merged. Closing the issue

PR #253, opened 2019-02-12
improve test coverage for multiple requests scenario

reyang:
This would address #211. @SergeyKanzhelev

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -805,13 +805,12 @@ class AdvancedTest(TestBase):
 			traceparent = self.get_traceparent(response['headers'])
 			trace_ids.add(traceparent.trace_id.hex())
 			parent_ids.add(traceparent.parent_id.hex())
-		self.assertEqual(len(trace_ids), 1)
 		self.assertEqual(len(parent_ids), 3)
 
 	def test_multiple_requests_with_illegal_traceparent(self):
 		'''
 		harness sends an invalid traceparent and asks vendor service to callback multiple times
-		expects a new trace_id generated and used for all the callbacks
+		expects new trace_id(s) generated
 		'''
 		trace_ids = set()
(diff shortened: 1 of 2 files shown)


plehegar:
(this is ok to merge despite the ipr check failure)

reyang:

SergeyKanzhelev (approved):
Thank you!

test/test.py
```diff
@@ -800,6 +808,22 @@ def test_multiple_requests(self):
 		self.assertEqual(len(trace_ids), 1)
 		self.assertEqual(len(parent_ids), 3)
 
+	def test_multiple_requests_with_illegal_traceparent(self):
```
bogdandrutu, line 810: I feel this is very specific to the test implementation, the fact that all 3 requests will have the same trace_id is not part of the specs in my opinion. We just need to check that they have a valid trace_id in my opinion, if an implementation of the test will create a new root span (new trace id) for every request I don't think that is necessary bad.

reyang: @bogdandrutu you're right.
It seems that we've only mentioned "restart trace" here https://github.com/w3c/trace-context/blob/master/spec/20-http_header_format.md#versioning-of-traceparent.

@SergeyKanzhelev what's your opinion?

SergeyKanzhelev: Yes, thinking about it - @bogdandrutu you are correct. We never ask for both - incoming and outgoing requests to be recorded necessarily. So if you only monitor outgoing calls - this test will not hold. @reyang would you mind just checking for correctness?

reyang: Test cases updated.

bogdandrutu (approved):

discostu105 (approved):

PR #246, opened 2019-02-02
Consolidates all work done on individual CR* branches

danielkhan:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/pull/246.html" title="Last updated on Feb 26, 2019, 4:49 PM UTC (02adfa9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/246/3e0c447...02adfa9.html" title="Last updated on Feb 26, 2019, 4:49 PM UTC (02adfa9)">Diff</a>

Issue #235, opened 2019-01-10
Decide, if we have to suggest an intrinsic format of trace-id


danielkhan:
In https://github.com/w3c/trace-context/blame/cr--edit-processing-model/spec/20-HTTP_HEADER_FORMAT.md#L105 we suggest that the random part should go on the right side to simplify sampling decisions.
Is this something that should go into the spec?

AloisReitbauer:
We should tell people that they should strive for maximum randomness. Add note that implementations should look into the randomness of their IDs

yurishkuro:
also that when you receive an inbound trace ID, you cannot make assumptions that some bits of it are more random than others.

danielkhan:
In our meeting on 2019-01-15 it was agreed that the wording should be trimmed down to just advice implementors to ensure randomness of the ID and not encode any information.
A randomness as of UUIDv4 is deemed to be good enough.

PR #259, opened 2019-02-28
Changes to wording, structure, grammar and spelling

danielkhan:
This PR contains is the first stage to bring the ongoing changes in various CR* branches into master.
It should only contain editorial changes. 
The processing model was omitted on purpose to bring it in as soon as this PR has landed.
Consequently, some other changes that were already discussed and were included in and where part of PR #246 will be added in a subsequent PR.
This is to keep the changeset clean and easier to review.

***
<a href="https://pr-preview.s3.amazonaws.com/danielkhan/distributed-tracing/pull/259.html" title="Last updated on Apr 1, 2019, 5:53 PM UTC (51fb31d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/259/117394f...danielkhan:51fb31d.html" title="Last updated on Apr 1, 2019, 5:53 PM UTC (51fb31d)">Diff</a>


AloisReitbauer (approved):
LGTM

discostu105 (approved):

index.html
```diff
@@ -1,83 +1,71 @@
...
 
-        <section data-include="spec/40-other-protocols.md" data-include-format='markdown'></section>
+<body>
+  <section id='sotd' data-include="spec/02-sotd.md" data-include-format='markdown'></section>
+  <section id='abstract' data-include="spec/01-abstract.md" data-include-format='markdown'></section>
 
-        <section data-include="spec/50-privacy.md" data-include-format='markdown'></section>
-        <section data-include="spec/51-security.md" data-include-format='markdown'></section>
-        
-        <section class="informative">
```
SergeyKanzhelev, line 79: Please revert this back

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
-Distributed tracing is a set of tools and practices to monitor the health and reliability of a distributed application. A distributed application is an application that consists of multiple components that are deployed and operated separately. It is also known as micro-service.
+# Problem Statement
```
SergeyKanzhelev: I'd suggest to close this PR https://github.com/w3c/trace-context/pull/221 to avoid double-reviewing

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
-Distributed tracing is a set of tools and practices to monitor the health and reliability of a distributed application. A distributed application is an application that consists of multiple components that are deployed and operated separately. It is also known as micro-service.
+# Problem Statement
 
-The main concept behind distributed tracing is event correlation. Event correlation is a way to correlate events from one component to the events from another. It allows to find the cause-and-effect relationship between these events. For instance – find which user action in a browser caused a failure in the business logic layer.
+Distributed tracing is a methodology implemented by tracing tools to follow, analyze
+and debug a transaction across multiple software components. Typically, a trace
+traverses more than one component which requires it to be uniquely identifiable
+across all participating systems. Passing along this unique identification is
+provided by *trace context propagation*.
```
SergeyKanzhelev: I don't understand the part "is provided by trace context propagation" part of a sentence. Perhaps "provided" is not the right word here. What does this sentence mean?

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
-Distributed tracing is a set of tools and practices to monitor the health and reliability of a distributed application. A distributed application is an application that consists of multiple components that are deployed and operated separately. It is also known as micro-service.
+# Problem Statement
 
-The main concept behind distributed tracing is event correlation. Event correlation is a way to correlate events from one component to the events from another. It allows to find the cause-and-effect relationship between these events. For instance – find which user action in a browser caused a failure in the business logic layer.
+Distributed tracing is a methodology implemented by tracing tools to follow, analyze
+and debug a transaction across multiple software components. Typically, a trace
```
SergeyKanzhelev: You can use `<a>distributed trace</a>` instead of `trace` so it will be referencing to glossary.

SergeyKanzhelev: Otherwise you need to put a definition of term "trace" here

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
-Distributed tracing is a set of tools and practices to monitor the health and reliability of a distributed application. A distributed application is an application that consists of multiple components that are deployed and operated separately. It is also known as micro-service.
+# Problem Statement
 
-The main concept behind distributed tracing is event correlation. Event correlation is a way to correlate events from one component to the events from another. It allows to find the cause-and-effect relationship between these events. For instance – find which user action in a browser caused a failure in the business logic layer.
+Distributed tracing is a methodology implemented by tracing tools to follow, analyze
+and debug a transaction across multiple software components. Typically, a trace
+traverses more than one component which requires it to be uniquely identifiable
+across all participating systems. Passing along this unique identification is
```
SergeyKanzhelev: scoping this work down to unique identifier of the trace is very limiting. Why the original paragraph about cause-and-effect events correlation wasn't good enough?

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+provided by *trace context propagation*.
 
-To correlate events between components, these components need to exchange and store a piece of information called context. Typically context consists of an originating event identifier, an originating component identity and other event properties. Context has two parts. The first part is a trace context. Trace context consists of properties crucial for event correlation. The second part is correlation context. Correlation context carries user-defined properties. These properties may be helpful for correlation scenarios. But they are not required and components may choose to not carry or store them.
+Today, trace context propagation is implemented individually by each vendor.
+In multi-vendor environments, this causes interoperability problems, like:
 
-Unifying the format of distributed tracing context as well as aligning on semantic meaning of the values is the main objective of this working group. The goal is to share this with the community so that various tracing and diagnostics products can operate together.
+- Traces that are collected by different tracing vendors cannot be linked together
```
SergeyKanzhelev: Again, replace `Traces` to `<a>Distributed traces</a>`

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+  as there is no shared unique identifier.
+- Traces that cross boundaries between different tracing vendors break as there
+  is no uniformly agreed set of identification that is forwarded.
+- Cloud platform vendors, as well as hardware vendors, cannot guarantee to support
```
SergeyKanzhelev: Please include SaaS vendors (e.g. facebook API) and proxies (e.g. Envoy). Which are neither of cloud nor hardware vendor

spec/01-abstract.md
SergeyKanzhelev: are first and second items identical?

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+Distributed tracing is a methodology implemented by tracing tools to follow, analyze
+and debug a transaction across multiple software components. Typically, a trace
+traverses more than one component which requires it to be uniquely identifiable
+across all participating systems. Passing along this unique identification is
+provided by *trace context propagation*.
 
-To correlate events between components, these components need to exchange and store a piece of information called context. Typically context consists of an originating event identifier, an originating component identity and other event properties. Context has two parts. The first part is a trace context. Trace context consists of properties crucial for event correlation. The second part is correlation context. Correlation context carries user-defined properties. These properties may be helpful for correlation scenarios. But they are not required and components may choose to not carry or store them.
+Today, trace context propagation is implemented individually by each vendor.
```
SergeyKanzhelev: perhaps say "vendor of distributed tracing solution" to add clarity what is "vendor" here

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+In multi-vendor environments, this causes interoperability problems, like:
```
SergeyKanzhelev, line 10: this is not necessarily multi-vendor environment as you use vendor as "tracing vendor". Thrid bullet is NOT about multi-vendor environment. It's about platforms forwarding

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
-Unifying the format of distributed tracing context as well as aligning on semantic meaning of the values is the main objective of this working group. The goal is to share this with the community so that various tracing and diagnostics products can operate together.
+- Traces that are collected by different tracing vendors cannot be linked together
+  as there is no shared unique identifier.
+- Traces that cross boundaries between different tracing vendors break as there
+  is no uniformly agreed set of identification that is forwarded.
+- Cloud platform vendors, as well as hardware vendors, cannot guarantee to support
+  trace context propagation as there is no standard to follow.
+
+In the past, these problems did not have a significant impact as most applications
+were monitored by a single tracing tool and stayed within the boundaries of a single
```
SergeyKanzhelev: perhaps use "tracing vendor" or "vendor" instead of "tracing tool" for consistency? Or I missing the difference assumed here?

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+platform provider. Today, an increasing number of applications are highly
+distributed and leverage multiple middleware services and cloud platforms.
```
SergeyKanzhelev, line 23: I'd suggest to end this paragraph with "This transformation of modern applications calls for distributed tracing context propagation standard". Or something to this effect. Otherwise it feels not-complete.

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+
+## Solution
+
+The trace context specification defines a universally agreed-upon format for the
+exchange of trace context propagation data - referred to as *trace context*. Trace
```
SergeyKanzhelev, line 31: Perhaps we can add trace context in glossary. I guess *italic* is fine for now

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+context solves the problems described above by
+
+- providing a unique identifier for individual traces, allowing traces of multiple
```
SergeyKanzhelev: not only traces, but also for individual requests

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+  providers to be linked together.
+- providing an agreed-upon mechanism to forward vendor-specific trace data and
```
SergeyKanzhelev, line 36: solution includes a new concept of "forward vendor-specific trace data" which was not mentioned as a problem in Problem Statement section. I suggest to state the problem in previous section

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+  avoid broken traces when multiple trace tools participate in a single transaction.
+- providing an industry standard that platform- and hardware-providers can support.
+
+A unified approach for propagating trace data improves visibility into the behavior
+of distributed applications, facilitating problem and performance analysis.
+The interoperability provided by trace-context is a prerequisite to manage modern
```
SergeyKanzhelev, line 43: please keep consistency and don't use "-" in trace-context

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+
+The trace context specification defines a universally agreed-upon format for the
+exchange of trace context propagation data - referred to as *trace context*. Trace
+context solves the problems described above by
+
+- providing a unique identifier for individual traces, allowing traces of multiple
+  providers to be linked together.
+- providing an agreed-upon mechanism to forward vendor-specific trace data and
+  avoid broken traces when multiple trace tools participate in a single transaction.
+- providing an industry standard that platform- and hardware-providers can support.
```
SergeyKanzhelev: do you need dash after the words platform and hardware?

spec/40-other-protocols.md
```diff
@@ -1,13 +1,10 @@
 # Other protocols
 
-Trace Context propagation is crucial to enable distributed tracing scenarios
-for applications spans multiple components. Http is one of the communication
-protocols used for cross components correlation. Extensions of this
-specification as well as specifications defined by external organizations define
-the format of trace context serialization and
-deserialization for other protocols. Note, that those extensions may be at a
-different maturity level than this specification.
+While trace context is defined for HTTP the authors acknowledge it is also relevant
```
SergeyKanzhelev, line 3: Is it common to use "authors" in a spec? This sentence reads snobbish for me =).

spec/40-other-protocols.md
```diff
@@ -1,13 +1,10 @@
...
+for other protocols. Extensions of this specification define the format of trace
+context serialization anddeserialization for other protocols.
```
SergeyKanzhelev: anddeserialization - add space

spec/40-other-protocols.md
```diff
@@ -1,13 +1,10 @@
...
 
-Trace Context propagation is crucial to enable distributed tracing scenarios
-for applications spans multiple components. Http is one of the communication
-protocols used for cross components correlation. Extensions of this
-specification as well as specifications defined by external organizations define
-the format of trace context serialization and
-deserialization for other protocols. Note, that those extensions may be at a
-different maturity level than this specification.
+While trace context is defined for HTTP the authors acknowledge it is also relevant
+for other protocols. Extensions of this specification define the format of trace
```
SergeyKanzhelev: why have you dropped "specifications defined by external organizations"? THis is the whole point of having a registry with external contributions

spec/40-other-protocols.md
```diff
@@ -1,13 +1,10 @@
 # Other protocols
```
SergeyKanzhelev: I'd just revert to the original wording

spec/10-overview.md
```diff
@@ -1,16 +1,13 @@
 # Overview
 
-Trace context consists of a set of properties crucial for event correlation in distributed applications. Trace context positions requests in a <a>distributed trace</a>. If you look at a <a>distributed trace<a> as a graph of correlated requests - a single request can be part of multiple graphs. For example, let's say component `A` calls component `B` which in turn calls `C`. One distributed tracing system may see component `A` calling component `B`. Another may see all three components communicating. So the call from `B` to `C` will carry trace context about service `A` as well as information about its direct communication upstream to `B`.
+Trace context is split into two individual propagation fields supporting interoperability and vendor-specifc extensability:
```
SergeyKanzhelev: typo in "extensability"

spec/10-overview.md
```diff
@@ -1,16 +1,13 @@
...
+Trace context is split into two individual propagation fields supporting interoperability and vendor-specifc extensability:
 
-The situation when a single request is a part of multiple graphs is becoming more common. One distributed application may have components monitored by different distributed tracing systems. Those distributed tracing systems may not be easily replaceable as they might be provided by cloud vendors or distributed as pre-built components.
+- `traceparent` describes the position of the incoming request in its trace graph in a portable, fixed-length format. Its design focuses on fast parsing. Every tracing tool MUST properly set `tracestate` even when it only relies on vendor-specif informaiton in `tracestate`
```
SergeyKanzhelev: You meant traceparent in this sentence, right?  "Every tracing tool MUST properly set `tracestate` even when it only relies on vendor-specif informaiton in `tracestate`"

SergeyKanzhelev: also two typos - "informaiton " and "specif"

SergeyKanzhelev: Also I believe we were using "SHOULD", not "MUST" - one can propagate `tracestate` without `traceparent`. Is it new or copy/paste from somewhere?

spec/10-overview.md
```diff
@@ -1,16 +1,13 @@
...
+- `tracestate` extends `traceparent` with vendor-specific data represented by a set of name/value pairs. Storing information in `tracestate` is optional.
 
-The trace context is designed to allow extensibility for all distributed tracing systems. It requires them to respect context set by other systems.
+Tracing tools can provide two levels of compliant behaviour interacting with trace context:
 
-Trace context is represented by a set of name/value pairs describing the identity of every http request. Two propagation fields carry the common and vendor-specific properties that make up the trace context.
+- At a minimum they MUST propagate the `traceparent` and `tracestate` headers and guarantee  traces are not broken. this behaviour is also referred to as forwarding a trace.
+- In addition they CAN also choose to participat in a trace by modifying the `traceparent` header and relevant parts of the `tracestate` header containing their proprietray information. This is also referred to as participating in a trace.
```
SergeyKanzhelev: typo "participat "

spec/10-overview.md
```diff
@@ -1,16 +1,13 @@
...
 
-* `traceparent` describes the position of the incoming request in its trace graph in a portable, fixed-length format. Its design focuses on fast parsing.
-* `tracestate` maps all graphs the incoming parent is a part of in potentially vendor-specific formats. For example, if a request crosses tracing systems, there will be one entry in `tracestate` for each system.
-
-Notably, the `tracestate` field is unreliant on data in the `traceparent`.
-
-Libraries and platforms MUST propagate `traceparent` and `tracestate` headers to guarantee that trace will not be broken.
+A tracing tool can choose to change this behavior for each individual request to a component it is monitoring. The decision on whether to participate or forward a trace is called sampling. A request is considered when the tracing tool participates in the trace.
```
SergeyKanzhelev: > whether to participate or forward a trace is called sampling.

I don't think I agree with this definition of sampling. Sampling is about collecting data, not propagating context.

SergeyKanzhelev: > A request is considered when …

Something missing here. Word "sampled"?

spec/10-overview.md
SergeyKanzhelev: Was the definition of tracestate as a way to identify request in multiple graphs moved somewhere? I believe it is crutial for understainding the semantics

spec/50-privacy.md
```diff
@@ -14,8 +14,10 @@ Note, both privacy concerns of `traceparent` field are theoretical rather than p
...
 ## Privacy of tracestate field
 
-The  `tracestate` field may contain any opaque value in any of the keys. The main purpose of this header is to provide additional vendor-specific trace-identification information across different distributed tracing systems. 
+The  `tracestate` field may contain any opaque value in any of the keys. The main purpose of this header is to provide additional vendor-specific trace-identification information across different distributed tracing systems.
 
 Tracing systems MUST NOT include any personally identifiable information in the `tracestate` header.
 
 Platforms and tracing systems extremely sensitive to personal information exposure MAY implement selective removal of values corresponding to the unknown keys. This mutation of the `tracestate` field is not forbidden, but highly discouraged. As it defeats the purpose of this field for allowing multiple tracing systems to collaborate.
+
+*NOTE*: Should we restart the trace in this situation?
```
SergeyKanzhelev: please remove the note if this is not work in progress

spec/50-privacy.md
```diff
@@ -2,7 +2,7 @@
...
 
-Trace vendors MUST assess the risk of header abuse. This section provides some considerations and initial assessment of the risk associated with storing and propagating these headers. Tracing systems or platforms may choose to inspect and remove sensitive information from the fields before allowing the tracing system to execute code that potentially can propagate or store these fields. All mutations should, however, conform to the list of mutations defined in this specification.
+Trace vendors MUST assess the risk of header abuse. This section provides some considerations and initial assessment of the risk associated with storing and propagating these headers. Trace vendors may choose to inspect and remove sensitive information from the fields before allowing the tracing system to execute code that potentially can propagate or store these fields. All mutations should, however, conform to the list of mutations defined in this specification.
```
SergeyKanzhelev: "Trace vendors" - very limiting in this context. Platforms and proxies may provide this functionality like they do for cookies.

SergeyKanzhelev (changes requested):

spec/50-privacy.md
danielkhan: I think we should come up with a final and concise wording for that at some point.
Is 'participants' acceptable?

spec/50-privacy.md
danielkhan: Can I take care of leftover notes when all changes have landed?
They will be gone by the time we give the 'go' for CR.

spec/10-overview.md
danielkhan: @AloisReitbauer could you comment on that as you did the original change?

spec/10-overview.md
danielkhan: Fixed in https://github.com/w3c/trace-context/pull/259/commits/1cb67acf85dc660d9daa69670785b53262e03ad3

index.html
danielkhan: This commit changes the order of the TOC. Is there a specific reason why it should be reverted?

spec/01-abstract.md
danielkhan: Done

index.html
danielkhan: This changes the order of the TOC - is there a specific reason you want this to be reverted back?

spec/01-abstract.md
danielkhan: This has been changed by request. If I remember correctly @yurishkuro was involved in this discussion.
Before it was phrased ''is called trace context propagation".
This was considered to be incorrect.

Trace context propagation provides the functionality to pass along a unique identifier.
Please suggest another wording if you feel this is incorrect.

spec/10-overview.md
danielkhan: I think what the text says is:

Collecting data: Sampling.
Not collecting data: Forwarding

So this follows our wording about participating and forwarding systems and is not about propagating context.

index.html
SergeyKanzhelev: I meant to revert the glossary.

spec/01-abstract.md
SergeyKanzhelev: got it. "called" instead of "provided by" makes sentence understandable. But I can see where @yurishkuro is coming from. Perhaps this feedback is similar to feedback I gave further below (not sure):

> scoping this work down to unique identifier of the trace is very limiting. Why the original paragraph about cause-and-effect events correlation wasn't good enough?

Perhaps you can reuse parts of an original abstract for better definitions.

spec/10-overview.md
SergeyKanzhelev: hm. System may change traceparent without actually collecting data. This is a perfectly normal behavior. Especially considering that sampling may happen later in collected data uploading pipeline. 

So even with this clarification it is not good definition of sampling

spec/50-privacy.md
SergeyKanzhelev: my understanding is that this PR was intended as 

>It should only contain editorial changes.

So let's not introduce unfinished sentences and notes

spec/50-privacy.md
SergeyKanzhelev: In this specific context this change is not correct. As for creating a glossary items - I agree, it is not a bad idea. `Participants` is a good word. Maybe too generic. Do you think "Tracing systems or platforms" is too long? BTW, we already have a mix of terms like "Tracing vendors " vs. "Tracing systems".

spec/50-privacy.md
danielkhan: I think that a note does not change the spec in any way but tells that there is something open to discuss. As such it is as editorial as it can get because it is not even intended to still be there in CR.
Regardless of that, I removed the note.

spec/50-privacy.md
danielkhan: I think that "Tracing systems or platform" is too verbose if it is spread all over the spec for every time we mean a system that participates in one way or the other.
Reducing this to one concise phrase is why I am very fond of the glossary section you introduced.

I changed it back to "Tracing systems or platforms" for now. Let's reduce the wording after CR.

spec/10-overview.md
danielkhan: How about:

> The decision on whether to record a trace is called sampling. A request is considered sampled when the tracing tool records a trace.

index.html
danielkhan: Done

spec/01-abstract.md
danielkhan: Changed to
<a>distributed trace</a>

spec/01-abstract.md
AloisReitbauer: The original paragraph introduced a lot of new concepts and was rather confusing.

spec/10-overview.md
AloisReitbauer: I agree with Sergey. Even we participate in a trace when we don't collect data. One reason is to pass "not-sampled" information downstream. 

We should simple only talk about forward and participating here. Whether the data is collected or not is not relevant to this spec.

spec/10-overview.md
AloisReitbauer: The original text was super confusing. A lot of concepts like trace graphs etc. People not familiar with the space will have a hard time to understand this. Also explaining multi hop trace text-only with support images makes things hard to grasp. 

I put in here how we solve the problem of interoperability and I believe this is what should be in this section.

I also see the need to have probably an entire section on how to properly use trace-context du build distributed traces. This will be super helpful and I am all in describing this. 

I am fine taking the first round of this section. "Building Distributed traces with Trace-Context"

spec/50-privacy.md
AloisReitbauer: I agree with Daniel. We should be clear about areas where we have open discussions and where behavior is not fully defined. We owe this to people implementing the spec. Ideally, we also link to an issue or PR.

Every other behavior would be unprofessional with regards to openness. If a potential implementor reads this and has a strong opinion he will also feel more motivated to provide feedback.

spec/50-privacy.md
AloisReitbauer: I like trace participants

spec/10-overview.md
danielkhan: That

> I agree with Sergey. Even we participate in a trace when we don't collect data. One reason is to pass "not-sampled" information downstream.
> 
> We should simple only talk about forward and participating here. Whether the data is collected or not is not relevant to this spec.

That would be the opposite of what @SergeyKanzhelev is saying @AloisReitbauer:

> I don't think I agree with this definition of sampling. Sampling is about collecting data, not propagating context.
Sampling is about collecting data, not propagating context

spec/10-overview.md
SergeyKanzhelev: Yes, it will be great to take the first part of it. Without multi-graph positioning explanation it's hard to understand the problem of interoperability and one of the key ideas behind `tracestate`. And examples of `tracestate` with `parent-id` in them looks out of place. As spec didn't explain the concept of using your `tracestate` entry for your own positioning. I agree that it needs better explanation and pictures. But I think simply removing this text makes spec less descriptive.

SergeyKanzhelev (changes requested):

spec/50-privacy.md
SergeyKanzhelev: If we have cases where we undecided like this - I'd suggest to put a bit more formal text like "Note, this specification is not defined exact recommendation of tracing system or platform behavior in this situation. Removal of `tracestate` entries may be harmful for interoperability and future versions of this specification may enforce restarting `traceparent` in these cases.". This will explains what comments says better.

As for this particular comment - spec allows to modify `tracestate` entries. So here depending on implementation of privacy-enforcement proxy or library - it may decide the behavior on its own. I can see how organization may have an outgoing traffic proxy that clears out internal-use-only `tracestate` entries. Forcing to restart in this situation will be damaging for interoperability

spec/01-abstract.md
mtwo: "tracing vendor" would also work

spec/40-other-protocols.md
mtwo: I think it's pretty standard to use authors in OSS organizations like this

spec/40-other-protocols.md
SergeyKanzhelev: if we go with this texting - I'd also suggest to specify "communication protocols". Just "protocols" may be confusing.

spec/40-other-protocols.md
SergeyKanzhelev: it's ok with me then.

spec/10-overview.md
danielkhan: Removed the part about sampling from overview in https://github.com/w3c/trace-context/pull/259/commits/fe0df40a00dfcaa62322a4feadd95b4361313326
@SergeyKanzhelev - does this work for you?

index.html
```diff
@@ -1,83 +1,88 @@
...
+  <section data-include="spec/50-privacy.md" data-include-format='markdown'></section>
+  <section data-include="spec/51-security.md" data-include-format='markdown'></section>
+
+  <section class="informative">
+    <h1>Glossary</h1>
+    <dt><dfn data-lt="distributed traces">Distributed trace</dfn></dt>
+    <dd>
+      A distributed trace is a set of events, triggered as a result
+      of a single logical operation, consolidated across various
+      components of an application. A distributed traces contains
```
mwear: Typo in the second sentence. "A distributed traces" should probably be "A distributed trace"

spec/10-overview.md
```diff
@@ -1,16 +1,13 @@
...
+- `traceparent` describes the position of the incoming request in its trace graph in a portable, fixed-length format. Its design focuses on fast parsing. Every tracing tool MUST properly set `tracestate` even when it only relies on vendor-specif informaiton in `tracestate`
+- `tracestate` extends `traceparent` with vendor-specific data represented by a set of name/value pairs. Storing information in `tracestate` is optional.
 
-The trace context is designed to allow extensibility for all distributed tracing systems. It requires them to respect context set by other systems.
+Tracing tools can provide two levels of compliant behaviour interacting with trace context:
 
-Trace context is represented by a set of name/value pairs describing the identity of every http request. Two propagation fields carry the common and vendor-specific properties that make up the trace context.
+- At a minimum they MUST propagate the `traceparent` and `tracestate` headers and guarantee  traces are not broken. this behaviour is also referred to as forwarding a trace.
```
mwear: Formatting: there's an extra space between "guarantee" and "traces"

spec/50-privacy.md
```diff
@@ -14,7 +14,7 @@ Note, both privacy concerns of `traceparent` field are theoretical rather than p
 
 ## Privacy of tracestate field
 
-The  `tracestate` field may contain any opaque value in any of the keys. The main purpose of this header is to provide additional vendor-specific trace-identification information across different distributed tracing systems. 
+The  `tracestate` field may contain any opaque value in any of the keys. The main purpose of this header is to provide additional vendor-specific trace-identification information across different distributed tracing systems.
```
mwear: Extra space between "The" and "`tracestate`".

spec/10-overview.md
AloisReitbauer: Let us still merge this one and handle the detailed usage in a separate PR. I don't see why this should prevent us from merging this PR

spec/50-privacy.md
AloisReitbauer: This is why I wanted to link to an issue. People can then follow the discussion thread and we will not have to revisit the same discussion over again. Does this work for you?

spec/40-other-protocols.md
AloisReitbauer: you had protocols in your text. but ok let us use communication protocols

danielkhan: Changed the title and added 'communication' once in the copy.

index.html
danielkhan: Fixed

spec/10-overview.md
danielkhan: Fixed

spec/50-privacy.md
danielkhan: Fixed

spec/01-abstract.md
danielkhan: Can we do these glossary replacements and wording changes in a second run?
This will otherwise just again complicate this whole PR.
The abstract section contains the word trace 17 times. Should we replace it every time or does it make sense to only use the full wording when the term is introduced?

spec/01-abstract.md
danielkhan: I think we need there one concise phrase that will be explained in the glossary. We can not mention each flavor of participant every time we speak about them.

AloisReitbauer:
Can some people please provide their review so we can move forward with this PR

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
-Distributed tracing is a set of tools and practices to monitor the health and reliability of a distributed application. A distributed application is an application that consists of multiple components that are deployed and operated separately. It is also known as micro-service.
+# Problem Statement
 
-The main concept behind distributed tracing is event correlation. Event correlation is a way to correlate events from one component to the events from another. It allows to find the cause-and-effect relationship between these events. For instance – find which user action in a browser caused a failure in the business logic layer.
+Distributed tracing is a methodology implemented by tracing tools to follow, analyze
+and debug a transaction across multiple software components. Typically, a
+<a>distributed trace</a> traverses more than one component which requires it to
+be uniquely identifiable across all participating systems.
+Passing along this unique identification is provided by *trace context propagation*.
```
yurishkuro: I recommend we add references to published papers that do a much better job of explaining the space of distributed tracing, to put this in context.

spec/20-http_header_format.md
```diff
@@ -1,34 +1,39 @@
...
+The `traceparent` header identifies an incoming request into a participating component.
 
-For example, a client traced in the congo system adds the following headers to
-an outbound http request.
+The `tracestate` header contains an ordered set of participating components
+as key-value pairs, where the key identifies the component and the value contains potentially
+proprietary data needed for processing by the component.
+As the components append their entry to the top of the set, while shifting
+other entries to the right, the left-most position implicitly tells whic∏h
+tracing system corresponds with the current `traceparent`.
```
yurishkuro: In this paragraph the use of the word component is inconsistent with earlier use. A component refers to a component of monitored distributed application. There's only one place here that uses that meaning, all other references are to tracing tools or instrumentation inside the components.

spec/20-http_header_format.md
```diff
@@ -1,34 +1,39 @@
...
+
+## Example
+
+A software system is traced by two components called `congo` and `rojo`.
```
yurishkuro: I suggest replacing these with the customary Foo and Bar. Avoids questions.

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
+2. Bug in caller
+3. Different load between caller service and callee service might force the
+   callee to downsample.
+
+More information can be found in the security section of this specification.
+
+The current version of the specification only supports a single flag called
+`recorded`.
+
+#### Recorded Flag (00000001)
```
yurishkuro: I thought we agreed to revert to using industry- (and statistics-) standard term "sampling". Let's not invent new non-standard terminology.

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
-Implementations HAVE TO ignore the `traceparent` when the `trace-id` is invalid.
-For instance, if it contains non-allowed characters.
+##### Rules
 
-### Parent-id
+* A `trace-id` with all bytes set to `0` is not allowed.
+* Implementations SHOULD generate globally unique values.
+* Implementations HAVE TO ignore the `traceparent` if the `trace-id` does not
+  comply with this specification.
+* If a system operates with a shorter `trace-id` - it SHOULD fill-in the extra
```
yurishkuro: s/SHOULD/MUST/, otherwise it won't comply with the point above

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
 
-Is the ID of this call as known by the caller. It is also known as `span-id` as
-a few telemetry systems call the execution of a client call a span. It is
-represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes `0`
-is considered invalid.
+  **Example:**
+  A system uses a 8-byte unique identifier like `3ce929d0e0e4736`. Instead of
+  setting the `trace-id` value to `0000000000000003ce929d0e0e4736` it is
+  recommended to generate a value like `4bf92f3577b34da6a3ce929d0e0e4736` where
+  `4bf92f3577b34da6a` is a random value or a function of time & host value.
```
yurishkuro: This doesn't make sense. Doing so will pretty much guarantee that the trace will break, either if the same component makes another request (different timestamp) or another component with the same 8-byte tracer makes a request.

If a tracer is fundamentally only capable of 8-byte trace-id (including in-process propagation), it cannot comply with this spec, UNLESS it uses 0-padding, which is the only stable way for it to ensure the same 16-byte trace-id

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
-`trace-id` value to `0000000000000003ce929d0e0e4736` it is recommended to
-generate a value like `4bf92f3577b34da6a3ce929d0e0e4736` where
-`4bf92f3577b34da6a` is a random value or a function of time & host value. Note,
-even though a system may operate with a shorter `trace-id` for <a>distributed trace</a>
-reporting - full `trace-id` should be propagated to conform to the
-specification.
+> **Note:** As some systems rely on randomness of the `trace-id` as a whole,
+> the provided ID MUST NOT contain encoded information that leads to non-random
+> parts (e.g. using a host ID as prefix).
+> A randomness as provided by UUID v4 is considered to be sufficient.
```
yurishkuro: I am not sure of the intention of this Note. How can some systems  **rely** on randomness? Isn't "globally unique" property sufficient?

IIRC we wanted to make a different, opposite point: you cannot rely on any part of the trace-id to represent anything but randomness. Even if your own  system encodes something in the bits (like the timestamp in X-Ray), when it receives the trace-id generated externally it cannot expect that.

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
-...
+// [...]
 boolean recorded = (traceFlags & FLAG_RECORDED) == FLAG_RECORDED
 ```
 
-Current version of specification only supports a single flag called `recorded`.
+**Note:** The encoded flags are recommendations given by the caller rather than
+strict rules to follow for three reasons:
+
+1. Trust and abuse
```
yurishkuro: Seriously, nobody is bothered by this "explain-nothing" wording? It's the 4th time I'm raising it.

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
+field". However there are set of recommendations for better vendor
 interoperability.
 
-1. If component made definitive recording decision - this decision SHOULD be
-   reflected in `recorded` flag.
-2. If component needs to make a recording decision - it SHOULD respect
+##### Recommendations
+
+1. If a component made a definitive recording decision - this decision SHOULD be
+2. reflected in the `recorded` flag.
```
yurishkuro: it's not `2.`

spec/20-http_header_format.md
```diff
@@ -91,242 +84,206 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
-   as a default option when trace is initiated by this component. There are two
-   additional options:
-    1. Component that makes deferred or delayed recording decision may
-       communicate priority of recording by setting `recorded` flag to `1` for a
-       subset of requests.
-    2. Component may also fall back to probability sampling to set flag
-       `recorded` to `1` for the subset of requests.
+4. If a component deferred or delayed the decision and only a subset of telemetry
+   will be recorded - the flag `recorded` should be propagated unchanged. And set
+   to `0` as a default option when a trace is initiated by this component. There
```
yurishkuro: This is certainly a safer recommendation, but it conflicts with the definition of the flag above which says flag=1 means "I may have sampled this trace". From this paragraph it sounds like we want to flip the definition to
  * flag=1: definitely sampled
  * flag=0: may or may not have sampled

yurishkuro (commented):
I don't have a strong opinion for or against merging this (prefer to merge and work with smaller diffs).

My main concern with the current text is that it plays fast and loose with the terminology. I suggest:
  * we agree on the exact terminology we want to use in this spec and stick to it, instead of using interchangeable terms (some of them conflicting in different contexts)
  * we refer to other sources, specifically to this paper http://www.pdl.cmu.edu/PDL-FTP/SelfStar/CMU-PDL-14-102.pdf, which provides a thorough overview of the tracing space. The trace context spec only addresses a subset of that space, namely tracing systems relying on metadata propagation. The paper is well-crafted in terms of terminology, we don't need to reinvent the wheel.

spec/20-http_header_format.md
danielkhan: @yurishkuro, this is part of the spec for **7 months** at least: https://github.com/w3c/trace-context/blame/d380547a5307d4fc418f4125f603aafce2103ada/trace_context/HTTP_HEADER_FORMAT.md#L96

The topic of randomness was again discussed [in this meeting](https://github.com/w3c/distributed-tracing-wg/blob/master/meeting_notes/2019-01-15.md).
AFAIK - you attended that meeting.

The current PR doesn't change the content or meaning but just the structure of the respective paragraph. Please create another PR or an issue if you'd like to re-discuss this whole topic.

spec/20-http_header_format.md
danielkhan: This was discussed in https://github.com/w3c/trace-context/issues/235
Please propose another wording if this does not reflect what was discussed.

spec/20-http_header_format.md
danielkhan: Please elaborate what you mean by this.

spec/20-http_header_format.md
danielkhan: Fixed

spec/20-http_header_format.md
danielkhan: This PR did not change the wording of this paragraph but just added a note that it needs more editing.

spec/20-http_header_format.md
danielkhan: This was not changed in this PR

spec/20-http_header_format.md
danielkhan: This was not changed in this PR.

spec/20-http_header_format.md
danielkhan: We agreed that we will do another round of edits where we reduce the variations in wording and relay the explanation to the glossary.

spec/10-overview.md
danielkhan: Fixed

spec/01-abstract.md
danielkhan: We will change this in a second round and reduce the variability in favor of a glossary.

spec/10-overview.md
AloisReitbauer: this is based on what is currently in the spec. We need to merge this one finally, so updated issue don't keep blocking it.

danielkhan: Fixed

spec/20-http_header_format.md
yurishkuro: There needs to be a complete sentence here, not a cryptic fragment / inside joke.

spec/20-http_header_format.md
```diff
@@ -362,70 +323,561 @@ nblk-chr = %x21-2B / %x2D-3C / %x3E-7E
...
 ```
 
+**Handling roundtrips:**
+
+If a tracing system key is `congo`, and a trace started in their system, went
+through a system with the key `rojo` and later returned to `congo`, the following
+`tracestate` would be *invalid*:
+
+```http
+// INVALID!
```
SergeyKanzhelev: what does invalid here means?

spec/01-abstract.md
AloisReitbauer: let us add tracing vendor here

spec/20-http_header_format.md
yurishkuro: Unfortunately, #235 refers to another discussion via a link that is broken.

danielkhan:
Yet in #235 we discussed the randomess and the hat to enforce.

spec/01-abstract.md
danielkhan: changed to tracing vendor as agreed on meeting 2019-03-13

index.html
```diff
@@ -1,83 +1,88 @@
...
-        </section>
-  </body>
-</html>
+
+  <section data-include="spec/10-overview.md" data-include-format='markdown'></section>
+  <section data-include="spec/11-fields-semantics.md" data-include-format='markdown'></section>
+
+  <section data-include="spec/20-http_header_format.md" data-include-format='markdown'></section>
+
+  <section data-include="spec/30-processing-model.md" data-include-format='markdown'></section>
```
SergeyKanzhelev: This file is not part of thie PR. Please remove

index.html
```diff
@@ -1,83 +1,88 @@
...
+      }],
+      wgPatentURI: "https://www.w3.org/2004/01/pp-impl/108594/status"
+    };
+  </script>
+</head>
 
-        <section data-include="spec/40-other-protocols.md" data-include-format='markdown'></section>
+<body>
+  <section id='sotd' data-include="spec/02-sotd.md" data-include-format='markdown'></section>
+  <section id='abstract' data-include="spec/01-abstract.md" data-include-format='markdown'></section>
```
SergeyKanzhelev, line 53: I'm very curious of this change. My understanding always was that spec has short abstract, than status and than longer introductiuon and spec itself.  @plehegar is it ok to swap those?

spec/20-http_header_format.md
```diff
@@ -91,242 +84,205 @@ parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
...
+
+More information can be found in the security section of this specification.
+
+The current version of the specification only supports a single flag called
+`recorded`.
+
+#### Recorded Flag (00000001)
 
-### Recorded Flag (00000001)
+> **NEEDS EDITING - this whole section should be reviewed for grammar and clarity.**
```
SergeyKanzhelev: please remove this note

spec/20-http_header_format.md
```diff
@@ -362,70 +323,561 @@ nblk-chr = %x21-2B / %x2D-3C / %x3E-7E
...
+* Libraries and platforms MUST accept empty `tracestate` headers, but SHOULD
+  avoid sending them. The reason for allowing of empty list members in `tracestate`
+  is a difficulty for implementors to recognize the empty value when multiple
+  `tracestate` headers are received. Whitespace characters are allowed for similar
+  reasons as some frameworks will inject whitespace after the `,` separator
+  automatically even when the header is empty.
+
+#### Limits
+
+> **PENDING - this was changed in favor of the 5 item limit** Maximum length of
```
SergeyKanzhelev: this is already resolved in separate PR

index.html
SergeyKanzhelev: Just to clarify on this - I looked at first 5 specs here https://www.w3.org/TR/ and all of them has a few sentences abstract in the beginning of the document before `sotd` section. And it makes sense as you want to quickly read what it is about and then decide whether you need to read it further. This change makes our spec different.

spec/01-abstract.md
SergeyKanzhelev: Having H1 inside the abstract section looks bad in rendered document. "Abstract" is already and H1 title.

https://pr-preview.s3.amazonaws.com/danielkhan/distributed-tracing/pull/259.html#overview

image

spec/01-abstract.md
```diff
@@ -1,7 +1,39 @@
...
+  is no uniformly agreed set of identification that is forwarded.
+- Cloud platform vendors, as well as hardware vendors, cannot guarantee to support
+  trace context propagation as there is no standard to follow.
+
+In the past, these problems did not have a significant impact as most applications
+were monitored by a single tracing tool and stayed within the boundaries of a single
+platform provider. Today, an increasing number of applications are highly
+distributed and leverage multiple middleware services and cloud platforms.
+
+## Solution
```
SergeyKanzhelev, line 28: why solution is a subsection of a problem?

SergeyKanzhelev (changes requested):
couple more comments that expand on the problem of a moved abstract

SergeyKanzhelev:
As discussed on Gitter - here is a summary of blocking from my perspective issues:

1. I think removing short abstract from the beginning of the document it wrong and should not be merged. See explanation in comment.
2. Abstract defines context as unique identification of a distributed trace. Comment was left with the request to say that it's also about causality of events. Otherwise definition is too limited.
3. Problem statement includes a list where two first items are identical. Third item is exclusionary to some distributed tracing participants. It makes abstract misleading. 
4. PR has conflicts, please resolve to make it merge-ready.

This PR was intended to improve understanding of the spec purpose and scope, but has problems that either limiting the scope of a spec or makes it unclear. Thus I believe those should be addressed before the merge.

We also should merge this PR only with assumption that:
1. Spelling mistakes will be addressed immediately as a separate PR.
2. PR that returns description of multi-graph identity of every request and how `tracestate` is used to describe this multi-graph identity will be worked on soon.

danielkhan (commented):
> 1. I think removing short abstract from the beginning of the document it wrong and should not be merged. See explanation in comment.

@AloisReitbauer please discuss with @SergeyKanzhelev 

> 2. Abstract defines context as unique identification of a distributed trace. Comment was left with the request to say that it's also about causality of events. Otherwise definition is too limited.

@SergeyKanzhelev, we agreed in the meeting on 2019-13-03 that your objections are valid and will be adressed in a separate PR where a complete description will be provided by @AloisReitbauer 

> 3. Problem statement includes a list where two first items are identical. Third item is exclusionary to some distributed tracing participants. It makes abstract misleading.

I think this refers to:
>> - Traces that are collected by different tracing vendors cannot be linked together
  as there is no shared unique identifier.

This is about correlation.

>> - Traces that cross boundaries between different tracing vendors break as there
  is no uniformly agreed set of identification that is forwarded.

This is about propagation.

>> - Cloud platform vendors, as well as hardware vendors, cannot guarantee to support
  trace context propagation as there is no standard to follow.

If there is no agreed upon standard, also platforms can not guarantee to propagate context.

1 != 2, please explain the problem of 3.

> 4. PR has conflicts, please resolve to make it merge-ready.

This will be taken care of.

> We also should merge this PR only with assumption that:
> 
> 1. Spelling mistakes will be addressed immediately as a separate PR.

The whole master branch needs to undergo a spelling check because - as it is now - it already contains wrong grammar and spelling errors. The contents of this PR will be checked with it as well.

> 2. PR that returns description of multi-graph identity of every request and how `tracestate` is used to describe this multi-graph identity will be worked on soon.

@AloisReitbauer will take that

spec/01-abstract.md
AloisReitbauer: We already discussed that in the next editing round we will clear out the spec and unify "vendor" usage. We decided so far on trace participant.

spec/01-abstract.md
AloisReitbauer: I think having it once in there for the first occurrence should be fine. Again this is not a major issue and should not block merging.

spec/01-abstract.md
AloisReitbauer: We had this discussion before. Will do this in the next clean up round.

spec/01-abstract.md
AloisReitbauer: Don't get what you mean. Can you please elaborate on this a bit more

spec/01-abstract.md
AloisReitbauer: again clean up work

index.html
plehegar: indeed. The abstract must appear before the SOTD.
[[
There must be a status section that follows the abstract, […]
]]
https://www.w3.org/pubrules/doc/rules/?profile=CR#sotd

spec/40-other-protocols.md
```diff
@@ -1,12 +1,10 @@
...
-Trace Context propagation is crucial to enable distributed tracing scenarios
-for applications spans multiple components. Http is one of the communication
-protocols used for cross components correlation. Extensions of this
-specification as well as specifications defined by external organizations define
-the format of trace context serialization and
-deserialization for other protocols. Note, that those extensions may be at a
-different maturity level than this specification.
+While trace context is defined for HTTP the authors acknowledge it is also relevant
+for other communication protocols. Extensions of this specification define the
+format of trace context serialization and deserialization for other protocols.
```
SergeyKanzhelev, line 6: was  `as well as specifications defined by external organizations  ` left out purposefully?

spec/40-other-protocols.md
```diff
@@ -1,12 +1,10 @@
...
+Note, that these extensions may be at a different maturity level than this specification.
+
+Please refer to the [protocols registry](https://w3c.github.io/trace-context-protocols-registry/) for the
```
SergeyKanzhelev: lost in merge: `[[!REGISTRY]]` - switch to normative references

spec/01-abstract.md
nikmd23: In HTTP, they'd call them intermediaries and service providers.
```suggestion
- Cloud platform vendors, intermediaries and service providers, cannot guarantee to support
```

spec/01-abstract.md
nikmd23:
```suggestion
In platform forwarding environments, this causes interoperability problems, like:
```

spec/01-abstract.md
SergeyKanzhelev:
```suggestion
were monitored by a single tracing vendor and stayed within the boundaries of a single
```

spec/01-abstract.md
SergeyKanzhelev:
```suggestion
The interoperability provided by trace context is a prerequisite to manage modern
```

spec/01-abstract.md
SergeyKanzhelev:
```suggestion
- providing an industry standard that platform and hardware providers can support.
```

SergeyKanzhelev (approved):

mtwo (approved):

Issue #252, opened 2019-02-12
What should we do when we restart a trace


AloisReitbauer:
In case you are restarting a trace you should record the old traceparent. We have not really defined this behaviour

_Originally posted by @AloisReitbauer in https://github.com/w3c/trace-context/pull/222/review_comment/create_

bogdandrutu:
I don't think this belongs to this project scope.

nikmd23:
Is it safe to assume that the "restarter" of the trace would have to handle that in whatever way makes sense for them? Perhaps that's stuffing something into tracestate, perhaps it's emitting a "bridge" telemetry item - or something else entirely.

AloisReitbauer:
I believe we have to define restarting a trace as creating a new trace-parent and clearing trace-state. If we would not clear trace-state this might have weird effects for a tracer b/c suddenly the traceid changed unexpectedly or may seem invalid.

SergeyKanzhelev:
@AloisReitbauer there are many cases when restart of a trace may be needed. Some implementation may need to restart trace for internal propagation and recover to original trace when going back to external services. We can recommend certain implementations in blog posts or reference implementations. But I don't think restricting one or another behavior should be part of a spec.

Furthermore, this is not the completely blocking factor, but it is important nore. This behavior is hard to test using the test harness to ensure compliance.

SergeyKanzhelev:
Discussion diverged from the originally posed "record all `traceparent`" proposal to discussion of this issue: https://github.com/w3c/trace-context/issues/146 

I think originally proposed recording of all `traceparent` is out of scope of a spec. Clearing the `tracestate` can be discussed separately. 

I vote for closing this issue as out of scope and keep #146

PR #260, opened 2019-03-01
traceparent restart should require tracestate cleanup rationale

### LINKED ISSUE #146: How and when are you allowed to restart a trace
We discussed the need and option to restart a trace.  The question came up during test case generation. 

Key questions are: 
- When are you allowed to restart a trace?
- Do you also clear ```tracestate``` when you restart?
- Are there multiple types of restart?
  - Soft restart keeping tracestate
  - Hard restart not keeping tracestate

SergeyKanzhelev:
Most of the text in this PR is for rationale, not specification. I marked it as editorial as this is what was assumed.

Hopefully this will close #146 and #252

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -242,7 +242,7 @@ base16(trace-flags) = 00  // not recorded
 
 ## Versioning of `traceparent`
 
-Implementations is opinionated about future version of the specification. Current
+This specification is opinionated about future version of the trace context. Current
 version of this specification assumes that the future versions of `traceparent`
 header will be additive to the current one.
 
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -124,7 +124,7 @@ a few telemetry systems call the execution of a client call a span. It is
 represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes `0`
 is considered invalid.
 
-Implementation HAVE TO ignore the `traceparent` when the `parent-id` is invalid.
+Implementations HAVE TO ignore the `traceparent` when the `parent-id` is invalid.
 For instance, if it contains non-allowed characters.
 
 ## Trace-flags
@@ -216,7 +216,7 @@ interoperability.
 ### Other Flags
 
 The behavior of other flags, such as (`00000100`) is not defined and reserved
-for future use. Implementation MUST set those to zero.
+for future use. Implementations MUST set those to zero.
 
 ## Examples of HTTP headers
 
@@ -242,11 +242,11 @@ base16(trace-flags) = 00  // not recorded
 
 ## Versioning of `traceparent`
 
-Implementation is opinionated about future version of the specification. Current
+Implementations is opinionated about future version of the specification. Current
 version of this specification assumes that the future versions of `traceparent`
 header will be additive to the current one.
 
-Implementation should follow the following rules when parsing headers with an
+Implementations should follow the following rules when parsing headers with an
 unexpected format:
 
 1. Pass thru services should not analyze version. Pass thru service needs to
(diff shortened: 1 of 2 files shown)


spec/20-http_header_format.md
```diff
@@ -453,7 +453,10 @@ Here is the list of allowed mutations:
 4. **Restarting trace**. All properties - `trace-id`, `parent-id`, `trace-flags`
    are regenerated. This mutation is used in the services defined as a front
    gate into secure networks and eliminates a potential denial of service attack
-   surface.
+   surface. Implementation SHOULD clean up `tracestate` collection on
+   `traceparent` restart. In rare cases when original `tracestate` entries must
```
AloisReitbauer: article missing: the original

AloisReitbauer: Be specific here. In which rare cases. Without examples adding this does not help the reader to decide.

AloisReitbauer (changes requested):

SergeyKanzhelev:
@AloisReitbauer can you please review again?

spec/20-http_header_format.md
```diff
@@ -453,7 +453,13 @@ Here is the list of allowed mutations:
 4. **Restarting trace**. All properties - `trace-id`, `parent-id`, `trace-flags`
    are regenerated. This mutation is used in the services defined as a front
    gate into secure networks and eliminates a potential denial of service attack
-   surface.
+   surface. Implementation SHOULD clean up `tracestate` collection on
```
mtwo: Implementation -> Implementations

spec/21-http_header_format_rationale.md
```diff
@@ -152,6 +152,34 @@ One variation is whether original or new header that you cannot recognize is pre
...
+than restored back when it is leaving those boundaries - keeping original
+`tracestate` entries will increase tracing vendors interoperability.
+Vendor-specific context will be propagated thru the these secure boundaries.
+
+Scenarios like one above require careful coding and understanding of what they
+are trying to achieve. It should be considered an exception. Implementations
+should discourage this type of restarts. For example, implementations may allow
+this scenario only by means of restarting the trace WITH `tracestate` clean up
+and than re-population of `tracestate` can only be implemented as an explicit
+copy/pasting of `tracestate` entries one by one.
```
mtwo, line 263: Modified for grammar / clarity

Developers sometimes feel a need to restart a trace due to security concerns,
however trace context should not contain information that can compromise your
application,  and tracing vendors should work through the risks of trusting
incoming identifiers without compromising on interoperability.

If restarting a trace is unavoidable, you SHOULD clear the `tracestate` list as well.
The data carried in `tracestate` carries information that often has a strong
association with the `traceparent`, particularly `trace-id`. Tracing vendors will
often assume that these values are associated, which is why both `tracestate`
SHOULD be cleared when `traceparent` is changed.

There are scenarios, however, which may require different behavior.
For example, if a trace was restarted when entered some secure boundaries and
than restored back when it is leaving those boundaries - keeping the original
`tracestate` entries will fully restore the trace back to normal. In this example,
vendor-specific context will be propagated through these secure boundaries.

Scenarios like the one above require careful coding and understanding of what they
are trying to achieve, and should be considered an exception. Implementations
should discourage this type of restarts. For example, implementations may allow
this scenario only by means of restarting the trace WITH `tracestate` clean up
and than re-population of `tracestate` can only be implemented as an explicit
copy/pasting of `tracestate` entries one by one.

mtwo (commented):
Made a suggestion for grammar / clarity

spec/20-http_header_format.md
SergeyKanzhelev: Also fixed the same accross the entire doc

spec/21-http_header_format_rationale.md
SergeyKanzhelev: thanks!

spec/20-http_header_format.md
```diff
@@ -242,11 +242,11 @@ base16(trace-flags) = 00  // not recorded
 
 ## Versioning of `traceparent`
 
-Implementation is opinionated about future version of the specification. Current
+Implementations is opinionated about future version of the specification. Current
```
mtwo: Implementations should be opinionated about future version of the specification. The current

SergeyKanzhelev: nice catch! I did a poor find/replace job, but it caught bigger problem. I believe this sentence should be reworded as it talks about spec being opinionated, not implementation. Fixed.

mtwo (approved):
Reviewed at Sergey's request

AloisReitbauer (approved):

nikmd23 (approved):
LGTM :shipit:

Issue #195, opened 2018-10-21
Question on test-harness cases around invalid tracestate


discostu105:
I've updated our implementation to pass almost all current test-cases, except for these:

    test_tracestate_member_count_limit
    test_tracestate_key_length_limit
    test_tracestate_key_illegal_vendor_format
    test_tracestate_key_illegal_characters
    test_tracestate_duplicated_keys

My struggle with these test-cases is that they require full parsing of tracestate. In my current C++ based implementation, I just find the position of our tracestate-entry and keep the "rest" of tracestate (before and after string) untouched.

Now, although it would be an interesting exercise to implement this algorithm as efficiently as possible (single-pass, allocation-free), I want to ask the more general question, if an implementer really needs to enforce all spec rules strictly like this and _reject_ a tracestate at violation? Of course an implementer must not produce data that violates the spec, but must it really check it all? Also, technically, I believe while the spec gives these requirements, but it does not say that a trace must be restarted if a violations happens (please correct me if that's wrong!).

Also: Are we going to assign a certain "badge" / "certification" to certain vendors if their implementation passes 100% tests? What does it mean? What if an implementer decides not to implement certain checks for said reasons?

discostu105:
We have discussed this today at W3C working group (see notes https://docs.google.com/document/d/12yecf7L1uibMIGP_YkTRM-KKCZZkluqmYqXxGFHD5J4/edit#).

We discussed not having theses verification tests mandatory. An implementer should only try to extract their own tracestate part. If that fails, optionally discard the whole tracestate, otherwise keep propagating the rest of the tracestate.

@SergeyKanzhelev , @reyang : what you do think?

reyang:
@discostu105 thanks for the updates!

I agree with the general rule _An implementer should only try to extract their own tracestate part. If that fails, optionally discard the whole tracestate, otherwise keep propagating the rest of the tracestate._ Let's
follow up in Pull Requests and discuss the details.

SergeyKanzhelev:
@reyang do we have categories supported by infra? We can mark these tests with the special "StrictValidation" category.

reyang:
@SergeyKanzhelev yes we can tag the test cases.

Do we want to update the spec first, or we don't want to define "StrictValidation" in the spec?

SergeyKanzhelev:
hm. Yes. spec is vague about parsing issues in `tracestate`. Perhaps some clarification is required. As a first step simply marking tests will be a good first step.

[...]

PR #281, opened 2019-04-23
Set strict-level on certain tracestate validation tests

discostu105:
This has been discussed some time ago on https://github.com/w3c/trace-context/issues/195.

FYI @reyang

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -250,3 +250,9 @@ Content-Type: application/json
 
 	OK
 	```
+## Strictness levels
+
+The test harness supports different levels of strictness. It can be configured via env variable `STRICT_LEVEL`:
+
+* `STRICT_LEVEL=2`: All tests included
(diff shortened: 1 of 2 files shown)


test/README.md
```diff
@@ -250,3 +250,9 @@ Content-Type: application/json
 
 	OK
 	```
+## Strictness levels
```
SergeyKanzhelev, line 253: I'd suggest to call it tags or categories. So there might be intersections. Like forward-only mode tests and strict mode the same time.

reyang (approved):
LGTM

SergeyKanzhelev (approved):
Approving as this PR make things better. I'd suggest though to look into categories instead of strictness levels

test/README.md
discostu105: Yes, categories would also be an option. Let's do that in another PR.

AloisReitbauer (approved):

PR #283, opened 2019-04-23
Removed Trace-ID from the tracestate in examples

SergeyKanzhelev:
#168

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -14,15 +14,17 @@ an outbound http request.
 
 ``` http
 traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
-tracestate: congo=BleGNlZWRzIHRohbCBwbGVhc3VyZS4
+tracestate: congo=t61rcWkgMzE
 ```
+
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

AloisReitbauer (approved):

bogdandrutu (approved):

spec/20-http_header_format.md
```diff
@@ -14,15 +14,17 @@ an outbound http request.
 
 ``` http
 traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
-tracestate: congo=BleGNlZWRzIHRohbCBwbGVhc3VyZS4
+tracestate: congo=t61rcWkgMzE
 ```
 
+Note, `t61rcWkgMzE` is base64 encoded parent ID `b7ad6b7169203331`.
```
nikmd23:
```suggestion
Note: In this case, the value `t61rcWkgMzE`, is the result of Base64 encoding the Trace-ID (`b7ad6b7169203331`), though such manipulations are not required in `tracestate`.
```

nikmd23 (approved):
I think the note that was added was a tiny bit confusing, and wasn't using the correct terminology in line with the rest of the spec. I've suggested an update for that.

PR #294, opened 2019-05-16
Update 01-abstract.md

duerst:
It would be good if the abstract contained a bit more context, so that people outside the WG can put this spec in context. I added just the minimum "HTTP" (and fixed what I think is a grammatical error), but I can imagine adding even a bit more context would be welcome by outside readers.


AloisReitbauer (approved):

SergeyKanzhelev (approved):

SergeyKanzhelev:

plehegar (approved):

discostu105 (approved):

PR #307, opened 2019-06-25
Writing edits

iredelmeier:
First round of changes from @robswhitmore.

A few notes:
* Had to remove the hard line breaks as they make editing and diffing edits much, much harder. We can always add them back in later on
* Capitalization of headers changed to be more consistent
* Splitting up the spec into multiple Markdown files when the generated version is a single file significantly complicates adding anchor links :/ Any thoughts on combining the files or some other workaround?
* @robswhitmore's changes also include some images for diagrams, which I haven't added yet. Is there anything special (directory management, etc.) around adding images?


plehegar:

plehegar (approved):

iredelmeier:
(@plehegar - my W3C stuff should also be signed now!)

AloisReitbauer (approved):

iredelmeier:
For anyone else looking at the delta, I forgot to mention that it's probably easiest to compare the two commits in the PR since the first **only** changes the line breaks.

Probably also worth doing a non-squash merge to keep the delta easy going forward.

discostu105 (approved):

mwear:
This PR introduces the convention of putting MUST, MAY, SHOULD, etc in italics. Based on this [discussion](https://github.com/w3c/trace-context/pull/314#discussion_r303669975) on #314, we should decide what our position is on using italics going forward.

mtwo:
> This PR introduces the convention of putting MUST, MAY, SHOULD, etc in italics. Based on this [discussion](https://github.com/w3c/trace-context/pull/314#discussion_r303669975) on #314, we should decide what our position is on using italics going forward.

Following up from our meeting, we decided that we will _not_ use italics and that we will remove them in a future PR

mwear:
Leaving this comment for posterity. We decided that MUST, MAY, SHOULD, etc will not be italicized going forward. I'll address this in a follow on PR.

Issue #310, opened 2019-07-02
Clarify 512 byte header size


mwear:
Under the "Header Value" section we say:

> The length of a combined header MUST be less than or equal to 512 bytes. If the length of a combined header is more than 512 bytes it SHOULD be ignored.

A few paragraphs later under "Limits" we say:

> Platforms and tracing vendors SHOULD propagate at least 512 characters of a combined header. This length includes commas required to separate list items. But does not include optional white space (OWA) characters.

I believe the intention is that tracers SHOULD propagate 512 bytes of tracestate, but can choose to propagate more or less.

We should also evaluate whether or not optional whitespace and commas should be included or not.

mwear:
We decided that tracing systems SHOULD propagate 512 bytes of tracestate, but can choose to propagate more or less based on their needs. 512 bytes is the total size of tracestate without special considerations for whitespace or commas.

mwear:
Closing. This was resovled in #313.

PR #313, opened 2019-07-03
Clarify tracestate header size

mwear:
This addresses issue #310.

Previously, the spec listed tracestate limits in two places, and used the phrasing at most 512 bytes, in one sentence, and at least 512 characters in another. The intent that is that tracing systems should propagate 512 characters, but can choose to propagate more or less depending on their situation. This also makes the change that white space characters  are included in the 512 character recommendation.

The two sentences have been consolidated and clarified under the tracestate limits section.

If there are any suggestions on wording, I'd be happy to address them.

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -299,8 +299,6 @@ chr      = %x20 / nblk-chr
 
 The `tracestate` value is the concatenation of trace graph key/value pairs
 
-The length of a combined header (`list` + `list-member`) _MUST_ be less than or equal to 512 bytes. If the length of a combined header is more than 512 bytes it _SHOULD_ be ignored.
-
 Example: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`
 
 Only one entry per key is allowed because the entry represents that last position in the trace. Hence vendors must overwrite their entry upon reentry to their tracing system.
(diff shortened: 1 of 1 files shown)


discostu105 (approved):

danielkhan (approved):

spec/20-http_header_format.md
```diff
@@ -394,9 +391,9 @@ Platforms and tracing systems MUST propagate this field. There might be multiple
 system may propagate them as they came, combine into a single header or split
 into multiple headers differently following the RFC specification.
 
-Platforms and tracing vendors SHOULD propagate at least 512 characters of a
-combined header. This length includes commas required to separate list items.
-But does not include optional white space (`OWA`) characters.
+Platforms and tracing vendors _SHOULD_ propagate at least 512 characters of a
```
yurishkuro: we don't use italic for MUST/SHOULD words, capitalization is enough.

yurishkuro (commented):
This PR is a small semantic change to the spec. Do we keep a  changelog?

mtwo (approved):
Approved pending resolution of Yuri's comment

AloisReitbauer (approved):

spec/20-http_header_format.md
mwear: We decided that we will not be using italics going foward, see [this comment](https://github.com/w3c/trace-context/pull/307#issuecomment-511920186), but will address removal in a future PR.

mwear:
This has been updated so that it's no longer conflicted after merging #307.

Issue #311, opened 2019-07-02
tracestate validation


mwear:
The processing model says:

> The tracer validates the tracestate header. If the tracestate header cannot be parsed the tracer deletes it.

What exactly does validating the tracestate header entail? Most of our size limits, etc are SHOULDs in the spec, so this is a little ambiguous. Also, do tracing systems that do not use tracestate need to do anything here, or can they just propagate it?

mwear:
We decided that tracers MAY choose to validate tracestate and can drop it, if it's invalid.

mwear:
Closing. This was resolved in #314.

PR #314, opened 2019-07-03
Make tracestate validation optional

mwear:
This addresses issue #311.

This PR qualifies tracestate validation as a MAY. The rationale is that validating tracestate forces every tracer to police the value of the header. This is a cost that some tracing systems may wish to avoid, especially those don't use tracestate.

If there are any suggestions on wording, or if this needs further clarification, let me know and I'd be happy to make any changes.

### DIFF
--- a/spec/30-processing-model.md
+++ b/spec/30-processing-model.md
@@ -27,7 +27,7 @@ If a `traceparent` header is received:
     * If the _version number is higher_ than supported by the tracer, the vendor uses the format defined in this specification (`00`) to parse `trace-id` and `parent-id`.
 The vendor will only parse the `trace-flags` values supported by this version of this specification and ignore all other values. If parsing fails, the vendor creates a new `traceparent` header and deletes the `tracestate`.
     * If the vendor _supports the version number_, it validates `trace-id` and `parent-id`. If either `trace-id`, `parent-id` or `trace-flags` are invalid, the vendor creates a new `traceparent `header and deletes `tracestate`.
-3. The vendor validates the `tracestate` header. If the `tracestate` header cannot be parsed the vendor deletes it.
+3. The vendor _MAY_ validate the `tracestate` header. If the `tracestate` header cannot be parsed the vendor _MAY_ discard the entire header. Invalid `tracestate` entries _MAY_ also be discarded.
 4. For each outgoing request the vendor performs the following steps:
     - The vendor _MUST_ modify the `traceparent` header:
 		1. **Update `parent-id`**. The value of property `parent-id` MUST be set to a value representing the ID of the current operation.


spec/30-processing-model.md
```diff
@@ -29,7 +29,7 @@ forwards Trace Context headers.
    `parent-id`.
        - If either `trace-id`, `parent-id` or `trace-flags` are invalid, the tracer
       creates a new `traceparent` header and deletes `tracestate`.
-5. The tracer validates the `tracestate` header. If the `tracestate` header
+5. The tracer _MAY_ validate the `tracestate` header. If the `tracestate` header
```
discostu105: "If the `tracestate` header cannot be parsed the tracer deletes it."

-> should we also change this to "If the `tracestate` header cannot be parsed the tracer, _MAY_ discard the whole header. It _MAY_ as well discard only the invalid entries."

mwear: I updated the comment with your suggestion with some slight edits. Let me know if it needs any further clarification or wordsmithing.

danielkhan (approved):

spec/30-processing-model.md
```diff
@@ -29,8 +29,7 @@ forwards Trace Context headers.
    `parent-id`.
        - If either `trace-id`, `parent-id` or `trace-flags` are invalid, the tracer
       creates a new `traceparent` header and deletes `tracestate`.
-5. The tracer validates the `tracestate` header. If the `tracestate` header
-   cannot be parsed the tracer deletes it.
+5. The tracer _MAY_ validate the `tracestate` header. If the `tracestate` header cannot be parsed the tracer _MAY_ discard the entire header. Invalid `tracestate` entries _MAY_ also be discarded.
```
yurishkuro: we don't use italics for MUST/MAY

mwear: The writing edits pr #307 introduces italics for MUST and MAY and I was writing with the assumption that PR would be merged. Do we strictly not want italics (if so, we should probably say so on that PR), or are they ok if we adopt them as a new convention?

yurishkuro: I would recommend not using italics, the upper-case already makes the words stand out, italics doesn't add anything new and makes plain  text harder to read. I've never seen italics in the ietf RFCs.

AloisReitbauer (approved):

spec/30-processing-model.md
mwear: We decided that we will not be using italics going foward, see [this comment](https://github.com/w3c/trace-context/pull/307#issuecomment-511920186), but will address removal in a future PR.

mwear:
This has been updated so that it's not longer conflicted after merging #307.

discostu105 (approved):

Issue #325, opened 2019-07-24
Conflicting guidlines for handling multiple tracestate headers


mwear:
There is conflicting information about how to handle multiple tracestate headers in the spec.

In section [3.3.1.1 tracestate Header Field Values](https://w3c.github.io/trace-context/#tracestate-header-field-values) we say:

> Multiple tracestate headers are allowed. Values from multiple headers in incoming requests SHOULD be combined in a single header according to Field Order [RFC7230], and sent as a single header in an outgoing request.

In section [3.3.1.5 tracestate Limits](https://w3c.github.io/trace-context/#tracestate-limits) we say:

> There might be multiple tracestate headers in a single request according to RFC7230 section 3.2.2. Vendors may propagate them as they came, combine them into a single header, or split them into multiple headers differently, following the RFC specification.

So, in one case we say that multiple tracestate headers must be combined before sending, and in the other, we say they can be combined, or sent separately.

We should clarify what we expect to happen in the multiple header case, and make sure the spec is consistent. I also wonder if we need to mention this at all, and could just say, tracestate can be split into multiple headers and should be processed according to [RFC7230 Section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2).

yurishkuro:
not really a conflict (should != must), but I agree that the "SHOULD combine" is unnecessary because the receiver still MUST be  able to cope with repeated header.

mwear:
Perhaps we can clean this up by combining the ideas from both sections, and mentioning it only once. I'd propose that we put the following under [3.3.1.1 tracestate Header Field Values](https://w3c.github.io/trace-context/#tracestate-header-field-values)

> Tracestate SHOULD be sent as a single header when possible, but MAY be split into multiple headers. When a request contains multiple tracestate headers they SHOULD be combined into a single header. Tracestate headers MUST be split or combined as specified by [RFC7329 Section 3.2.2 Field Order](https://tools.ietf.org/html/rfc7230#section-3.2.2)

If that sounds ok, I can get a PR together with those changes and we can handle the wordsmithing  there.

yurishkuro:
sgtm, but I think in  the 2nd sentence we need MUST, not SHOULD. Actually, I would rephrase it "When receiving multiple `tracestate` headers, then MUST be combined into a single `tracestate` per RFC7329".

mtwo:
[...]

PR #328, opened 2019-07-25
Clarify handling of tracestate as multiple header fields

mwear:
I opened #325 as I had interpreted two places that we mention handling tracestate as multiple headers as being in conflict. @yurishkuro pointed out, that they weren't technically conflicting as it is a SHOULD rather than MUST situation. Either way, it seems like something we can clarify. 

Below is a summary the discussion from #325:

In section [3.3.1.1 tracestate Header Field Values](https://w3c.github.io/trace-context/#tracestate-header-field-values) we say:

> Multiple tracestate headers are allowed. Values from multiple headers in incoming requests SHOULD be combined in a single header according to Field Order [RFC7230], and sent as a single header in an outgoing request.

In section [3.3.1.5 tracestate Limits](https://w3c.github.io/trace-context/#tracestate-limits) we say:

> There might be multiple tracestate headers in a single request according to RFC7230 section 3.2.2. Vendors may propagate them as they came, combine them into a single header, or split them into multiple headers differently, following the RFC specification.

I suggested that we combine these ideas, and mention them in only one of the sections: 3.3.1.1 tracestate Header Field Values.

Here's what I came up with:

> Tracestate MAY be sent or received as multiple header fields. Multiple tracestate header fields MUST be handled as specified by [RFC7230 Section 3.2.2 Field Order](https://tools.ietf.org/html/rfc7230#section-3.2.2). The `tracestate` header SHOULD be sent as a single field when possible, but MAY be split into multiple header fields. When sending `tracestate` as multiple header fields, it MUST be split according to [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.2.2). When receiving multiple `tracestate` header fields, they MUST be combined into a single header according to [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.2.2).

It might a little on the verbose side, and I'm open to any wordsmithing or suggestions, but I think having the ideas consolidated and in a single place will help clarify how multiple header fields should be handled.

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -240,7 +240,7 @@ Vendors MUST expect the header name in any case (upper, lower, mixed), and SHOUL
 
 ##### tracestate Header Field Values
 
-The `tracestate` field may contain any opaque value in any of the keys. Multiple `tracestate` headers are allowed. Values from multiple headers in incoming requests SHOULD be combined in a single header according to [Field Order](https://httpwg.org/specs/rfc7230.html#page-24) [[RFC7230](https://w3c.github.io/trace-context/#bib-rfc7230)], and sent as a single header in an outgoing request.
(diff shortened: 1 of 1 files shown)


mtwo (approved):

danielkhan (approved):

danielkhan:

plehegar (approved):

mtwo:
@iredelmeier has a coworker who wants to comment on this. We'll hold off merging until they get a chance to review.

mwear:
@iredelmeier Let us know if and when your coworker has had a chance to review these changes.

mwear:
Merging. I got the go ahead from @iredelmeier on the W3C Slack.

PR #332, opened 2019-09-19
Update test cases to reflect the latest spec change

### LINKED ISSUE #304: Add negative test cases for Tracestate value
We need to have negative test cases which test against invalid tracestate value.

### LINKED ISSUE #311: tracestate validation

### LINKED ISSUE #331: Update tracestate test to validate new tenant format
There was an issue opened back in July that proposed a change to the format of the tenant id: https://github.com/w3c/trace-context/issues/316

That work eventually got merged via: https://github.com/w3c/trace-context/pull/322

However, the `key with vendor` format regex wasn't updated here: https://github.com/w3c/trace-context/blob/master/test/tracecontext/tracestate.py#L6

I think the change should be as simple as something like this:

```
_KEY_WITH_VENDOR_FORMAT = r'[a-z0-9][_0-9a-z\-\*\/]{0,240}@[a-z][_0-9a-z\-\*\/]{0,13}'
```

reyang:
Fix #304.
Fix #331.


reyang:

test/tracecontext/tracestate.py
```diff
@@ -2,8 +2,8 @@
 import re
 
 class Tracestate(object):
-	_KEY_WITHOUT_VENDOR_FORMAT = r'[a-z][_0-9a-z\-\*\/]{0,255}'
-	_KEY_WITH_VENDOR_FORMAT = r'[a-z][_0-9a-z\-\*\/]{0,240}@[a-z][_0-9a-z\-\*\/]{0,13}'
+	_KEY_WITHOUT_VENDOR_FORMAT = r'[0-9a-z][_0-9a-z\-\*\/]{0,255}'
```
mwear: Based on the [EBNF](https://github.com/w3c/trace-context/blob/master/spec/20-http_header_format.md#key) in the spec (pasted below), I don't believe that the key can start with a digit unless it uses the vendor format. 

```ebnf
key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
key = ( lcalpha / DIGIT ) 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" ) "@" lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
lcalpha    = %x61-7A ; a-z
```

mwear (changes requested):
Hi @reyang. Thanks for updating the test cases! I have one comment about the `_KEY_WITHOUT_VENDOR_FORMAT` regex.

test/tracecontext/tracestate.py
reyang: You're right! I didn't put a close eye one it - I assumed that vendor / non-vendor key would both use DIGIT as the 1st char.
I'm surprised that we have different designs. Will try to dig out **why** we're doing this.

reyang: @mwear FYI - I have it fixed. Let's try to get this PR merged first. Will investigate **why** we have the deviation and probably create an issue to follow up.

mwear: @reyang The spec was changed to allow a more flexible format for the tenant id (the part before the `@` sign). This is because tenant ids are likely to have representations that will contain digits, possibly as their first character. So far there have not been any scenarios or requests for similar changes to the vendor name format.

mwear (approved):
Thanks for these updates @reyang!

test/tracecontext/tracestate.py
reyang: I see, thanks for the background @mwear :)

danielkhan (approved):
Thank you!

SergeyKanzhelev (approved):

AloisReitbauer (approved):

discostu105 (approved):

PR #351, opened 2019-11-03
moved content about trace-id into informative section as a first step…

SergeyKanzhelev:
… of addressing larger feedback on operating 64bit systems

As a first step fixing #349 the entire note about the behavior of `trace-id` was moved to informative section. The reason for this move is to enable us writing recommendation freely even if they break formal requirement of propagating entire `trace-id`. This will enable us to explain better how 64bit systems should operate for better inter-oprability avoiding the demand to re-start the entire `trace-id` every time.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 3, 2019, 6:33 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Ftrace-context%2Fa80774c4072eb86cf87a4a2df5a224f7ab3ca527%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/w3c/trace-context/a80774c4072eb86cf87a4a2df5a224f7ab3ca527/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/trace-context%23351.)._
</details>

PR #356, opened 2019-11-06
Fix the section on 64bit systems

### LINKED ISSUE #337: Revert advice against backfilling of zeros

### LINKED ISSUE #349: Inter-operability of tracing systems operating with shorter trace-id
This issue is tracking an overall resolution of inter-operability of tracing systems operating with shorter trace-id.

---

@bogdandrutu @adriancole @tylerbenson @jcchavezs @yurishkuro @nicmunroe @tedsuo @reyang @danielkhan (people who was involved in PR discussion) and others, please review this issue. It is written based on my understanding of the problems identified in a spec. Let's agree that this is the list of problems we have to address and then split this single issue into smaller issues to discuss how to resolve them.

Please comment/react on this issue to indicate that it reflects your understanding of a problems identified in spec.

---

There are existing tracing systems that currently operate with the 64bit `trace-id`, but still want to use the Trace Context headers for cross-process communication. The spec attempts to give an advice on how 128-bit, compliant to the spec tracing systems may cooperate wit the 64bit systems.

Let's call the logic of generating 128bit `trace-id` by 64bit system a backfill logic. While addressing concern of adding more clarity into the spec regarding `trace-id` backfill logic (see #337), and whether this backfill logic must apply to new traces or traces "in transition", more problems with the current spec was highlighted. See discussion in PR #344.

Here are the issues that were identified. 

## 1. Split trace problem

If app A calls two apps, B and C, with a 64-bit trace, and both B and C each backfill using different algorithm, then the trace become split. It now has two separate trace IDs.

This is an original problem raised in #337. Spec must be very clear on how to backfill `trace-id` in case of new `trace-id` generation comparing to the backfill logic on `trace-id` propagation.

## 2. Backfilling on left vs. right

Many existing systems dealing with the 64bit to 128bit transition already. These systems typically have a logic to backfill `trace-id` and look up traces given the subset of a `trace-id` bytes. It is common that these systems implement `trace-id` look ups based on right-most characters as a typical backfill logic puts zeroes in the first bytes of `trace-id`. 

This backfill logic contradict with the other paragraph of a spec asking to put the randomness to the left side of the `trace-id`. So backfill logic, in order to follow this requirement, must backfill the rightmost bytes.

This creates confusion and breaks inter-operability with the existing 64bit tracing systems.

## 3. Left padding of randomness.

Left-padding of randomness requirement creates another challenge for existing 64bit tracing systems. Keeping rightmost bytes constant will break compatibility with the tracing systems that operates on these rightmost characters.

This requirement must be either rewritten to make sure it works with 64bit systems or removed from the specification.

## 4. Zero vs. random backfill

We can differentiate tracing systems into three types:

- The *128bit* system: operating with 128bit trace-id (fully compliant). 
- The *64bit* system: operating with 64bit trace-id. Not capable to propagate `tracestate` and 64 extra bits of `trace-id` 
- The *128on64bit* system: records 64 bit trace-id. But capable to propagate extra 64 bit of trace-id from incoming request to outgoing request.

Backfill logic that is currently described in a spec may work for *128on64bit* systems. But it will not work for the *64bit* systems. Even on a first trace-id generation, *64bit* systems has no capabilities to preserve extra bytes and re-use them for subsequent outgoing calls from a single incoming call with the newly generated `trace-id`. Thus the requirements for *64bit* systems may be different. Or the current requirement can be changed to simple zero backfill.

## 5. Spec should prescribe the behavior of a 64bit-only system

*64bit* systems (as oppose to *128on64*) are quite harmful for inter-operability of tracing systems. Not only these systems fail to propagate an entire `trace-id`, they also fail to propagate `tracestate`.

Specification currently declares these systems non-compliant - uses the "MUST" language in `trace-id` propagation. However inter-operability with these tracing systems is quite important. So the note in specification on how 64bit systems should operate will go a long way of ensuring a better interoperability of various tracing systems.

The spec will benefit from being more explicit on how 64bit tracing systems must operate.

SergeyKanzhelev:
Fixes #349
Fixes #337


mtwo (approved):

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+The value of `trace-id` SHOULD be globally unique. This field is typically used
+for unique identification of a <a>distributed trace</a>. It is common for
+<a>distributed traces</a> to span various components, including, for example,
+cloud services. Cloud services tend to serve variety of clients and have a very
+high throughput of requests. So global uniqueness of `trace-id` is important,
+even when local uniqueness might seem like a good solution.
+
+### Randomness of `trace-id`
+
+Randomly generated value of `trace-id` SDHOULD be preferred over other
```
yurishkuro:
```suggestion
Randomly generated value of `trace-id` SHOULD be preferred over other
```

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+algorithms of generating a globally unique identifiers. Randomness of `trace-id`
+addresses some [secutiry](#security-considerations) and [privacy
```
yurishkuro:
```suggestion
addresses some [security](#security-considerations) and [privacy
```

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+concerns](#privacy-considerations) of exposing unwanted information. Randomness
+also allows tracing vendors to base sampling decisions on `trace-id` field value
+and avoid propagating an additional sampling context.
+
+As shown in the next section, it is important for better inter-operability with
+some existing systems, for `trace-id` to carry "uniqueness" and "randomness" in
+the right part of the `trace-id`.
```
yurishkuro: swap parts of the sentence

As shown in the next section, it is important for `trace-id` to carry "uniqueness" and "randomness" in the right part of the `trace-id`, for better inter-operability with some existing systems.

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+
+### Length of `trace-id` for the platforms operating with shorter identifiers
+
+There are system operates with a `trace-id` that is shorter than 16 bytes,
```
yurishkuro:
```suggestion
There are system that operate with a `trace-id` that is shorter than 16 bytes,
```

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+which still willing to adopt this specification.
```
yurishkuro:
```suggestion
which are still willing to adopt this specification.
```

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+
+If such system capable of propagating an entire `trace-id`, even while operating
```
yurishkuro:
```suggestion
If such a system is capable of propagating an entire `trace-id`, even while operating
```

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+internally on shorter identifier, system is encouraged to utilize the
+`tracestate` header to propagate shorter identifier. However, if system wants to
+use a subset of trace-id bytes as an identifier - rightmost part of a `trace-id`
```
yurishkuro:
```suggestion
use a subset of trace-id bytes as an identifier - the rightmost part of a `trace-id`
```

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+SHOULD be used.
+
+There are tracing systems that are not capable to propagate the entire 16 bytes
+of a `trace-id`. For better interoperability of these systems with the fully
+compliant systems, the following practices are recommended:
+
+1. Whenever a new `trace-id` needs to be generated - it SHOULD be left padded
```
yurishkuro: I think this whole point is irrelevant. How the system operates internally, including ID generation, is not important, because this specification would only matter on egress from such system.

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,52 @@
...
+   with zeroes. For example, for the identifier `53ce929d0e0e4736`, `trace-id`
+   must be set to `000000000000000053ce929d0e0e4736`.
+2. On incoming `traceparent`, the rightmost part of `trace-id` SHOULD be used as
```
yurishkuro: I would split this into two bullet items:
  * on ingress, the incoming trace-id should be truncated by dropping the left-most bytes
  * on egress, the outgoing trace-id should be formed by padding the internal 64bit id with 0 bytes on the left

justinfoote: I agree about reframing this section, but I don't see it so much as ingress vs egress, I see it as crossing W3C compliance boundaries.  

When a trace moves from a 64 bit system to a W3C compliant 128 bit system, the 128 bit system should left-pad with zeros.  

When a trace moves from a W3C compliant 128 bit system to a 64 bit system, we expect the 64 bit system to either propagate the full 128 bits (if possible) or propagate only the 64 least-significant bits of the trace id.

tedsuo: Just to add a further clarification: the conversions must happen in the new, complaint systems, not in the legacy systems. The legacy systems should be assumed to be unchangeable. So the egress/ingress directions are slightly different.

I also feel that it would be clearer to have two completely separate sections: one for describing new systems which use a non-complaint internal identifier, and one for new systems interoperating with legacy systems.

This would be my suggested change:

```markdown
### Handling `trace-id` for complaint platforms with shorter internal identifiers

There are tracing systems which use a `trace-id` that is shorter than 16 bytes, 
which are still willing to adopt this specification.

If such a system is capable of propagating a fully compliant `trace-id`, even 
while still requiring a shorter, non-complaint identifier for internal purposes, 
the system is encouraged to utilize the `tracestate` header to propagate the 
additional internal identifier. However, if a system would instead prefer to use 
the internal identifier as the basis for a fully complaint `trace-id`, it SHOULD 
be incorporated at the as rightmost part of a `trace-id`.

### Interoperating with legacy formats which use shorter identifiers

There are tracing systems which are not capable of propagating the entire 16 
bytes of a `trace-id`. When a fully compliant system wished to interoperarate 
with these legacy systems, the following practices are recommended:

1. Whenever a shorter `trace-id` is recieved from a legacy system, a compliant 
   `trace-id` SHOULD be generated by left padding the original identifier with 
   zeroes. For example, when receiving the identifier `53ce929d0e0e4736`, the 
   new `trace-id` should be set to `000000000000000053ce929d0e0e4736`.
2. When converting `traceparent` to a legacy format, the rightmost part of 
   `trace-id` SHOULD be used as the identifier when truncating. For instance, 
   if the the value of `trace-id` was `234a5bcd543ef3fa53ce929d0e0e4736` on an 
   incoming request, all outgoing requests to legacy systems would be sent a 
   `trace-id` with the value of `53ce929d0e0e4736`.
```

tedsuo:
Thanks @SergeyKanzhelev. One question: there was a suggestion to move this information to a "non-normative" section. Is that where this is now? Sorry for my ignorance, but I assumed the file name would be changing.

SergeyKanzhelev:
@tedsuo move to the non-normative section was already done: https://github.com/w3c/trace-context/pull/351

spec/60-trace-id-format.md
SergeyKanzhelev: I think I resolved it https://github.com/w3c/trace-context/pull/356/commits/75f04af39c509ba061d9c64e2979234dfe2274a6

Modified text suggested by @tedsuo a bit. And added a note about `tracestate` propagation

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,67 @@
...
+system may receive `234a5bcd543ef3fa53ce929d0e0e4736` as a `trace-id`, hovewer
+internally it will use `53ce929d0e0e4736` as an identifier.
+
+### Interoperating with existing systems which use shorter identifiers
+
+There are tracing systems which are not capable of propagating the entire 16
+bytes of a `trace-id`. For better interoperability between a fully compliant
+systems with these existing systems, the following practices are recommended:
+
+1. Whenever a shorter `trace-id` needs to be send, a compliant `trace-id` SHOULD
```
yurishkuro: sorry, I disagree with @tedsuo's comments on ingress/egress. The responsibility for conversion should be on the non-compliant system, not on the compliant sender/receiver. If that system is a blackbox that cannot be changed, it can be deployed with proxies / sidecar that will perform rewriting of the headers to the format it understands. But if my service is fully compliant and needs to talk to such legacy system, I should not be required to couple my service with understanding of what the legacy system understands or not.

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,67 @@
...
+
+Similar transformations are expected when tracing system converts other
+distributed trace context propagation formats to W3C Trace Context. Shorter
+identifiers SHOULD be left padded with zeros when converted to 16 bytes
+`trace-id` and rightmost part of `trace-id` SHOULD be used as a shorter
+identifier.
+
+Note, many existing systems that are not capable of propagating the whole
+`trace-id` will not propagate `tracestate` header either. However, such system
+can still use `tracestate` header to propagate additional that is known by this
```
yurishkuro: > propagate additional

missing word?

spec/60-trace-id-format.md
yurishkuro: This is why I was suggesting to phrase these in terms of ingress/egress to/from non-compliant region of the architecture.

SergeyKanzhelev: the word compliant here refers to compliant trace-id. I can replace it with 16 bytes `trace-id`

SergeyKanzhelev: I think @tedsuo wrote the paragraph with the idea on how to convert from traceparent to some other format with shorter ID. This is why it's harder to read with the different idea in mind. I added paragraph to talk about conversions

SergeyKanzhelev: added clarifications and rephrased a bit

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,74 @@
...
+system may receive `234a5bcd543ef3fa53ce929d0e0e4736` as a `trace-id`, hovewer
+internally it will use `53ce929d0e0e4736` as an identifier.
+
+### Interoperating with existing systems which use shorter identifiers
+
+There are tracing systems which are not capable of propagating the entire 16
+bytes of a `trace-id`. For better interoperability between a fully compliant
+systems with these existing systems, the following practices are recommended:
+
+1. Whenever a shorter identifier needs to be send, a 16 bytes `trace-id` SHOULD
```
yurishkuro: When a system creates an outbound message and needs to generate a fully compliant 16 bytes `trace-id` from a shorter identifier, it SHOULD left pad the original identifier with zeroes.

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,74 @@
...
+   be generated by left padding the original identifier with zeroes. For
+   example, the identifier `53ce929d0e0e4736`, SHOULD be converted to
+   `trace-id` value `000000000000000053ce929d0e0e4736`.
+2. When converting 16 bytes `trace-id` to a shorter identifier, the rightmost
```
yurishkuro: When a system receives an inbound message and needs to convert the 16 bytes `trace-id` to a shorter identifier, ...

yurishkuro (commented):
I made two suggestions to make it clear who is responsible for conversions

justinfoote:
@yurishkuro Since there is such disagreement about which system should be required to upgrade or downgrade trace ids, perhaps it's better not to mention the system in this spec.  I like the existing passive wording of `When converting a...`.

spec/60-trace-id-format.md
justinfoote: I think this description is backwards.  Bullet 1 here is about _upgrading_ a trace id from 64 bits to 128 bits.  

```suggestion
1. To convert an 8 byte `trace-id` to 16 bytes, the `trace-id` SHOULD
```

SergeyKanzhelev: this is too specific. I took @yurishkuro suggestion that basically conveys the same message. Please review. Thanks!

yurishkuro (approved):

justinfoote (approved):

SergeyKanzhelev:
Hi @bogdandrutu @adriancole @tylerbenson @jcchavezs @nicmunroe @reyang @danielkhan . There are no surprises in this PR. I believe it is reflecting overall agreement on how interoperability must work. If you have some time - please review.

Thank you!

spec/60-trace-id-format.md
```diff
@@ -1,20 +1,75 @@
...
-reporting, the full `trace-id` MUST be propagated to conform to the
-specification.
+This section suggests some best practices to consider when platform or tracing
+vendor implement `trace-id` generation and propagation algorithms. These
+practices will ensure better interoperability of different systems.
+
+### Uniqueness of `trace-id`
+
+The value of `trace-id` SHOULD be globally unique. This field is typically used
+for unique identification of a <a>distributed trace</a>. It is common for
```
reyang, line 10: is the `<a>` tag by mistake ```<a>distributed trace</a>```?

reyang (approved):

spec/60-trace-id-format.md
SergeyKanzhelev: no, this is ReSpec thing to point to glossary.

nicmunroe (approved):

tylerbenson (approved):

SergeyKanzhelev:
Thank you everybody for reviews!

PR #359, opened 2019-11-14
Clarify wording when no traceparent is received

### LINKED ISSUE #347: Wording in 4.2.2 is unclear
In section `4.2 No traceparent Received` the item number 2 currently reads:
>If no traceparent header is received, the vendor creates a new trace-id and parent-id that represents the current request.

It is unclear why there is `If no traceparent header is received` portion at all when the entire section is titled `No traceparent Received`. In this section it is an assumed fact that traceparent is not received, so questioning this fact in the form of `if` statement is confusing.

I suggest the following better wording which is also consistent with 4.3:

>Because the traceparent header is not present, the vendor creates a new trace-id and parent-id that represents the current request.

tigrannajaryan:
Resolves https://github.com/w3c/trace-context/issues/347

### DIFF
--- a/spec/30-processing-model.md
+++ b/spec/30-processing-model.md
@@ -11,7 +11,7 @@ This processing model describes the behavior of a vendor that modifies and forwa
 If no traceparent header is received:
 
 1. The vendor checks an incoming request for a `traceparent` and a `tracestate` header.
-2. If _no_ `traceparent` header is received, the vendor creates a new `trace-id` and `parent-id` that represents the current request.
+2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request.
 3. If a `tracestate` header is received without an accompanying `traceparent` header, it is invalid and MUST be discarded.
 4. The vendor SHOULD create a new `tracestate` header and add a new key/value pair.
 5. The vendor sets the `traceparent` and `tracestate` header for the outgoing request.


spec/30-processing-model.md
```diff
@@ -11,7 +11,7 @@ This processing model describes the behavior of a vendor that modifies and forwa
 If no traceparent header is received:
 
 1. The vendor checks an incoming request for a `traceparent` and a `tracestate` header.
-2. If _no_ `traceparent` header is received, the vendor creates a new `trace-id` and `parent-id` that represents the current request.
+2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request.
```
SergeyKanzhelev, line 14: I think item 2 and item 3 are two branches in a processing model. If not recieved - do this, if recieved - do that

SergeyKanzhelev (changes requested):
I think this change changes the meaning of the text

plehegar:

spec/30-processing-model.md
tigrannajaryan: This entire section is titled _No traceparent Received_. This means that the _if_ in item 2 is superfluous and asserts a condition which is always true in this section and thus is unnecessary.

mtwo (approved):

dyladan (approved):

danielkhan (approved):

yurishkuro (approved):

SergeyKanzhelev (approved):

PR #365, opened 2019-12-03
Add response headers

dyladan:
***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/365.html" title="Last updated on Mar 25, 2020, 6:37 PM UTC (6212e44)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/365/f9bfb5e...dynatrace-oss-contrib:6212e44.html" title="Last updated on Mar 25, 2020, 6:37 PM UTC (6212e44)">Diff</a>


spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+
+Vendors MUST ignore the `tracemeta` when the `parent-id` is invalid (for example, if it contains non-lowercase hex characters).
+
+#### trace-flags
+
+An <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a>  that controls tracing flags such as sampling, trace level, etc. These flags are recommendations given by the callee rather than strict rules to follow for three reasons:
+
+1. Trust and abuse
+2. Bug in the callee
+3. Different load between callee service and callee service might force caller to downsample.
```
yurishkuro: "callee service" twice

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+
+This is the ID of the calling request as known by the callee (in some tracing systems, this is known as the `span-id`, where a `span` is the execution of a client request). It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
+
+Vendors MUST ignore the `tracemeta` when the `parent-id` is invalid (for example, if it contains non-lowercase hex characters).
+
+#### trace-flags
+
+An <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a>  that controls tracing flags such as sampling, trace level, etc. These flags are recommendations given by the callee rather than strict rules to follow for three reasons:
+
+1. Trust and abuse
```
yurishkuro: this is not a sentence

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+2. Bug in the callee
```
yurishkuro: This is not a sentence

spec/21-http_response_header_format.md
dyladan: Taken straight from https://github.com/dynatrace-oss-contrib/trace-context/blob/add-response-headers/spec/20-http_request_header_format.md#trace-flags

If we update it here we should update it there too WDYT

spec/21-http_response_header_format.md
dyladan: Taken straight from https://github.com/dynatrace-oss-contrib/trace-context/blob/add-response-headers/spec/20-http_request_header_format.md#trace-flags

If we update it here we should update it there too WDYT

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+The `tracemeta` HTTP response header field identifies a completed request in a tracing system. It has four fields:
+
+* `version`
+* `trace-id`
+* `parent-id`
+* `trace-flags`
+
+### Header Name
+
+Header name: `tracemeta`
```
mtwo: My preference is for something like `traceresponse` rather than `tracemeta`, as `tracemeta` is overly broad for this use case

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+
+This is the ID of the whole trace forest and is used to uniquely identify a <a href="#dfn-distributed-traces">distributed trace</a> through a system. It is represented as a 16-byte array, for example, `4bf92f3577b34da6a3ce929d0e0e4736`. All bytes as zero (`00000000000000000000000000000000`) is considered an invalid value.
+
+If the `trace-id` value is invalid (for example if it contains non-allowed characters or all zeros), vendors MUST ignore the `tracemeta`.
+
+See [considerations for trace-id field
+generation](#considerations-for-trace-id-field-generation) for recommendations
+on how to operate with `trace-id`.
+
+#### parent-id
```
mtwo: This field is optional, correct?

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+
+This is the ID of the calling request as known by the callee (in some tracing systems, this is known as the `span-id`, where a `span` is the execution of a client request). It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
+
+Vendors MUST ignore the `tracemeta` when the `parent-id` is invalid (for example, if it contains non-lowercase hex characters).
+
+#### trace-flags
```
mtwo, line 75: This field is optional, correct?

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,184 @@
...
+* `version`
+* `trace-id`
+* `parent-id`
+* `trace-flags`
+
+### Header Name
+
+Header name: `tracemeta`
+
+In order to increase interoperability across multiple protocols and encourage successful integration, by default vendors SHOULD keep the header name lowercase. The header name is a single word without any delimiters, for example, a hyphen (`-`).
```
justinfoote: I'd like to remove this phrase: :
> by default vendors SHOULD keep the header name lowercase

The _by default_ part makes me think that this functionality maybe should be configurable, which doesn't make sense to me.  
And I'm not sure what the verb _keep_ means in this context.  

```suggestion
The header name is a single word without any delimiters, for example, a hyphen (`-`).
```

Also, this is described more clearly two lines down: 
> Vendors MUST expect the header name in any case (upper, lower, mixed), and SHOULD send the header name in lowercase.

spec/21-http_response_header_format.md
dyladan: I like the name `traceresponse`. I was trying to think of something that implied describing an existing trace as opposed to defining a new one.

spec/21-http_response_header_format.md
dyladan: Yes

spec/21-http_response_header_format.md
dyladan: Yes

spec/21-http_response_header_format.md
dyladan: This wording was copied directly from the forward header spec:

traceparent: https://github.com/w3c/trace-context/blob/e32fa09fb7c2796ac2d10f5bb73cae249c4ae21b/spec/20-http_header_format.md#header-name
tracestate: https://github.com/w3c/trace-context/blob/e32fa09fb7c2796ac2d10f5bb73cae249c4ae21b/spec/20-http_header_format.md#header-name-1

Should it be changed in all three places?

spec/21-http_response_header_format.md
dyladan: I'll add a note in the paragraph.

spec/21-http_response_header_format.md
dyladan: I'll add a note in the paragraph.

spec/21-http_response_header_format.md
justinfoote: I don't know whether it makes sense to change it in all three places.  It doesn't feel like the highest priority.  

I'm OK with us leaving the wording as is.

dyladan: What about this?

```suggestion
In order to increase interoperability across multiple protocols and encourage successful integration, the header name SHOULD be lowercase. The header name is a single word without any delimiters, for example, a hyphen (`-`).
```

justinfoote: 👍

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,190 @@
...
+
+This section describes the binding of the distributed trace context to the `traceresponse` HTTP header.
+
+## Traceresponse Header
+
+The `traceresponse` HTTP response header field identifies a completed request in a tracing system. It has four fields:
+
+* `version`
+* `trace-id`
+* `parent-id`
```
tylerbenson: I think using "parent" in a response is rather confusing.  Is it referring to the span that initiated the request (and is likely receiving the response), or the span that is sending the response?

Depending on the answer, it should probably be renamed to something else.

dyladan: It is actually neither. The motivational use case for this was the web browser initial page load. The page load happens before the instrumentation js is loaded, so the initial request is sent without a traceparent header. The first instrumented service can then create a "fake" span which it assigns to be it's parent and returns the id of the fake span to the client. The client then uses that id when it creates the span for the page load. So it is the id that is actually assigned to the parent span, making `parent-id` a decent name IMO.

cce: I just wanted to chime in that we use response headers at SolarWinds (historically based on X-Trace headers used in Tracelytics, TraceView, and now AppOptics), and our response header includes a trace ID and event ID from the span that wrote the response in the header. So for us "parent" may not strictly be true (it's actually the child), but we can work with any name.

cce: Basically, for us, the response header communicates a [happened-before relationship](https://en.wikipedia.org/wiki/Happened-before) between the end of the responding span and the end of the caller span, and also confirms the successful communication of the callee's response, as well as the caller's trace ID, between caller and callee.

spec/21-http_response_header_format.md
dyladan: @yurishkuro do you think we should update this in both places (request and response headers), only update it here, or is it fine to just leave?

spec/21-http_response_header_format.md
dyladan: @yurishkuro do you think we should update this in both places (request and response headers), only update it here, or is it fine to just leave?

spec/21-http_response_header_format.md
yurishkuro: We should change it everywhere. This is some techno-jargon that people not familiar with the space find hard to understand, we should not have such language in the spec.

dyladan: Please peek at https://github.com/w3c/trace-context/pull/365/commits/98ccaeb7e955a8f5ef31d36dbca5aa9aa5c6899a

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,190 @@
...
+Vendors MUST ignore the `traceresponse` when the `parent-id` is invalid (for example, if it contains non-lowercase hex characters).
+
+The `parent-id` field is an optional part of the `traceresponse` response header.
+
+#### trace-flags
+
+An <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a>  that controls tracing flags such as sampling, trace level, etc. These flags are recommendations given by the callee rather than strict rules to follow for three reasons:
+
+1. An untrusted caller may be able to abuse a tracing system by setting these flags maliciously.
+2. A caller may have a bug which causes the tracing system to have a problem.
```
yurishkuro, line 80: 🎉 🎉 🎉

spec/21-http_response_header_format.md
yurishkuro: @cce I think you're describing a substantially different use case.

* "parent" makes sense to communicate back to the browser the ID it _should use for its span_ (because for some reason it could not have created that ID before sending the request to the server)
* @cce's case is to communicate back the server-side event ID

mtwo (approved):

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,190 @@
...
+The behavior of other flags, such as (`00000100`) is not defined and is reserved for future use. Vendors MUST set those to zero.
+
+
+## Returning the traceresponse Field
+
+Vendors MAY choose to include a `traceresponse` header on any response, regardless of whether or not a `traceparent` header was included on the request.
+
+Following are suggested use cases:
+
+- **Restarted trace**. When a request crosses a trust boundary, the called service may decide to restart the trace. In this case, the called service MAY return a `traceresponse` field indicating its internal `trace-id` and sampling decision.
```
justinfoote: Indenting these use cases with bullets doesn't really separate them well once the markdown is formatted.  How would you feel about giving them sub-headings instead?
```suggestion
### **Restarted trace**. When a request crosses a trust boundary, the called service may decide to restart the trace. In this case, the called service MAY return a `traceresponse` field indicating its internal `trace-id` and sampling decision.
```

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,190 @@
...
+Response
+```http
+traceresponse: 00-1baad25c36c11c1e7fbd6d122bd85db6--01
+```
+
+In this example, a participant in a trace with ID `4bf92f3577b34da6a3ce929d0e0e4736` calls a third party system that collects their own internal telemetry using a new trace ID `1baad25c36c11c1e7fbd6d122bd85db6`. When the third party completes its request, it returns the new trace ID and internal sampling decision to the caller. If there is an error with the request, the caller can include the third party's internal trace ID in a support request.
+
+**Note**: In this case, the `parent-id` was omitted from the response because, being a part of a different trace, it was not necessary for the caller.
+
+- **Load balancer**. When a request passes through a load balancer, the load balancer may wish to defer a sampling decision to its called service. In this instance, the called service MAY return a `traceresponse` field indicating its sampling decision.
```
justinfoote:
```suggestion
### **Load balancer**. When a request passes through a load balancer, the load balancer may wish to defer a sampling decision to its called service. In this instance, the called service MAY return a `traceresponse` field indicating its sampling decision.
```

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,190 @@
...
+Response
+```http
+traceresponse: 00---01
+```
+
+In this example, a caller (the load balancer) in a trace with ID `4bf92f3577b34da6a3ce929d0e0e4736` wishes to defer a sampling decision to its callee. When the callee completes the request, it returns the internal sampling decision to the caller.
+
+**Note**: In this case, both the `parent-id` and `trace-id` were omitted from the response. Because the trace was not restarted and only a sampling decision was requested by the caller, the `parent-id` and `trace-id` were not changed.
+
+- **Web browser**. When a web browser that does not natively support trace context loads a web page, the initial page load will not contain any trace context headers. In this instance, the server MAY return a `traceresponse` field for use by a tracing tool that runs as a script in the browser.
```
justinfoote:
```suggestion
### **Web browser**. When a web browser that does not natively support trace context loads a web page, the initial page load will not contain any trace context headers. In this instance, the server MAY return a `traceresponse` field for use by a tracing tool that runs as a script in the browser.
```

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,190 @@
...
+
+Example response:
+
+```http
+traceresponse: 00-4bf92f3577b34da6a3ce929d0e0e4736-d75597dee50b0cac-01
+```
+
+In this example, the server is telling the browser that it should adopt trace id `4bf92f3577b34da6a3ce929d0e0e4736` and parent id `d75597dee50b0cac` for the current operation.
+
+- **Tail sampling**. When a service that made a negative sampling decision makes a call to another service, there may be some event during the processing of that request that causes the called service to decide to sample the request. In this case, it may return its updated sampling decision to the caller, the caller may also return the updated sampling decision to its caller, and so on. In this way, as much of a trace as possible may be recovered for debugging purposes even if the original sampling decision was negative.
```
justinfoote:
```suggestion
### **Tail sampling**. When a service that made a negative sampling decision makes a call to another service, there may be some event during the processing of that request that causes the called service to decide to sample the request. In this case, it may return its updated sampling decision to the caller, the caller may also return the updated sampling decision to its caller, and so on. In this way, as much of a trace as possible may be recovered for debugging purposes even if the original sampling decision was negative.
```

justinfoote (approved):
Overall, I love this PR.  It matches what we discussed in Seattle, it has some great examples, it has good parity with the request header spec, and I love that you went back and updated the request header spec to add the word _request_.  

I added some minor formatting comments, which you can take or leave.  

Thanks for doing the legwork on this one, Dan!

spec/21-http_response_header_format.md
dyladan: https://github.com/w3c/trace-context/pull/365/commits/40568d32efaaf13ca52190681d938488c5779ceb

justinfoote: 🎉

spec/21-http_response_header_format.md
codefromthecrypt: parent doesnt make sense. parent passed to a server is the parent of the server span. the span id generated is more often the child of the caller. returning something named parent in the response is confusing and not intuitive. focusing on what you know. ex the span id that serviced the request, is more sensible if this is included at all.

also bear in mind response header processing at all cant be mandatory. this isnt mentioned explicitly. if you do decide to do this it is making this not implementable without significant rewrites of code.

dyladan: @adriancole @cce I think the issue may be that the name of the field does not fully convey it's intended semantic. Given the situation where service `foo` calls service `bar`, if `bar` returns a `parent-id` to `foo`, it is not saying "this is my span id", but rather "this is the span ID that you, my parent, should use."

The motivating use case is a situation like a browser where the initial request may not contain a span id, but the called service needs to use some id as its "parent," so it generates its own id but also generates an id that it uses as its parent. This parent id is then sent back up the wire to the calling service so that it can either use that id as its own span id, or it can at least link that id to its span.

Would changing the field name to something like `suggested-parent-id`, `assumed-parent-id`, `requested-span-id` or similar be more useful in conveying this semantic?

dyladan: Here is a more visual explanation of the use case:

image

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,197 @@
...
+The value is US-ASCII encoded (which is UTF-8 compliant).
+
+Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `255` is invalid. The current specification assumes the `version` is set to `00`.
+
+#### version-format
+
+The following `version-format` definition is used for version `00`.
+
+``` abnf
+version-format   = [trace-id] "-" [parent-id] "-" [trace-flags]
```
SergeyKanzhelev: Making fields optional while keeping the same format as `traceparent` carry potential issues. Some may decide to use the same optionality for `traceparent` (even though it's not allowed by spec) and break interoperability. Mostly because the same code may be used to parse and generate both headers.

This said, it will be great to confirm that all the scenario listed below in this proposal:
1. Would actually significantly benefit from fields optionality.
2. Define a clear logic on how to decide which parts to return.
3. Would not need more properties than listed in `traceresponse` and will ACTUALLY need the arbitrary `tracestate` on response instead.

codefromthecrypt:
as absence of parent implies root it is odd to create a routine to assume
that we should always assume there is a caller. I have heard privately that
another APM is able to do correlation without this mechanism. it would be
good if folks can speak up rather than create an odd behavior case which
assumes most requests are not only browser oriented but will be
instrumented browser side

danielkhan (approved):
Great work @dyladan!

spec/21-http_response_header_format.md
yurishkuro: Let's rename this to `proposed-parent-id` (it's not like this string is sent over the wire, a longer name does not hurt).

I would also prefer that the spec clearly states that in case of the caller already providing well-formed `traceparent` header, the `proposed-parent-id` should be omitted.

dyladan: https://github.com/w3c/trace-context/pull/365/files/41aa080c4d1a8c7132a8f8da2170dc861e6461ce..a412b5dd318f10baeab89933eae36e218a1958a8

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,197 @@
...
+Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `255` is invalid. The current specification assumes the `version` is set to `00`.
+
+#### version-format
+
+The following `version-format` definition is used for version `00`.
+
+``` abnf
+version-format   = [trace-id] "-" [proposed-parent-id] "-" [trace-flags]
+trace-id         = 32HEXDIGLC  ; 16 bytes array identifier. All zeroes forbidden
+proposed-parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
```
yurishkuro, line 51: nit: may want to fix whitespace/alignment

spec/21-http_response_header_format.md
```diff
@@ -0,0 +1,197 @@
...
+
+The `trace-id` field is an optional part of the `traceresponse` response header. If the request header contains a valid `traceparent` with a `trace-id`, and the callee does not use a different `trace-id`, the callee SHOULD omit the `trace-id` field from the `traceresponse`.
+
+See [considerations for trace-id field
+generation](#considerations-for-trace-id-field-generation) for recommendations
+on how to operate with `trace-id`.
+
+#### proposed-parent-id
+
+This is the ID of the calling request as known by the callee (in some tracing systems, this is known as the `span-id`, where a `span` is the execution of a client request). It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
```
yurishkuro, line 69:
```suggestion
This is the ID of the calling request proposed by the callee back to the caller in cases when the caller did not send the `traceparent` header (in some tracing systems, this ID is known as the `span-id`, where a `span` is the execution of a client request). It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
```

yurishkuro:
Extra use case: when server creates a different trace (e.g. a cloud service), it may want to return its trace ID to allow the client to correlate with it.

yurishkuro:
Merged as the current proposal. Open issues:
  * how to represent two distinct use cases: reporting server's trace/span ID vs. proposing span ID for the client
  * how to deal with optionality, especially given different use cases (maybe use dictionary)

codefromthecrypt:
@yurishkuro let me get this right. you are literally going to expect people to speculatively generate parent IDs when creating a root span, which effectively means root spans won't have a null parent anymore? Then, add some middleware to push this back (more overhead), also speculatively as there's no way to know if the caller actually would use it. you are ok doing this without clearing this with with basically all OSS implementations except Otel which expect a root span to have no parent?

dyladan:
@adriancole the working group decided on the call to merge this as a starting place today. In no way is it meant to be considered final and nothing is unchangeable. Also, note that the response header is an entirely optional part of the trace. If there is some material change you'd like to see feel free to join the call tomorrow or Friday (not sure what timezone you're in but the call is 2PM-5PM US East), file an issue, or open a PR.

codefromthecrypt:
@dyladan I raised protest to this before you merged it and you merged it anyway without answering to that. what good would raising an issue be if you will just merge things anyway?

https://github.com/w3c/trace-context/pull/365#issuecomment-584974324

yurishkuro:
> expect people to speculatively generate parent IDs when creating a root span, which effectively means root spans won't have a null parent anymore

I expect this to be at minimum configurable in the SDK, and perhaps even triggered by something in the request. It's an odd use case with which I don't have any experience.

codefromthecrypt:
In attempts to give a constructive way out, as this was merged IMHO by mistake.

Revert all text about `proposed-parent-id`, then re-introduce a PR with it, or better yet keep out until better studied and far more distinct stakeholders have a use case to pay for its complexity.

In normal OSS, you reduce change to what can be merged, and strip out stuff that isn't ready to be merged. That can work here, also even if retroactively.

codefromthecrypt:
I also opened this, as I was surprised to find tracestate is a rare thing we are required to not allow unless accompanied by traceparent https://github.com/w3c/trace-context/issues/401

PR #376, opened 2020-01-28
Minor grammar fix

### LINKED ISSUE #368: Fix grammar in higher-version flags parsing
The current text is:

> Parse the sampled bit of flags (2 characters from the third dash). Vendors MUST check that the 2 characters are either the end of the string or a dash.

Suggested possible edits:

> Parse the sampled bit of flags (2 characters from the third dash). Vendors MUST check that the 2 characters are either **at** the end of the string or **followed by** a dash.

> Parse the sampled bit of flags (2 characters from the third dash). Vendors MUST check that the 2 characters are **followed by** either the end of the string or a dash.

dyladan:
Fixes #368

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -214,7 +214,7 @@ Vendors MUST follow these rules when parsing headers with an unexpected format:
     - If the size of the header is shorter than 55 characters, the vendor should not parse the header and should restart the trace.
     - Parse `trace-id` (from the first dash through the next 32 characters). Vendors MUST check that the 32 characters are hex, and that they are followed by a dash (`-`).
     - Parse `parent-id` (from the second dash at the 35th position through the next 16 characters). Vendors MUST check that the 16 characters are hex and followed by a dash.
-    - Parse the `sampled` bit of `flags` (2 characters from the third dash). Vendors MUST check that the 2 characters are either the end of the string or a dash.
+    - Parse the `sampled` bit of `flags` (2 characters from the third dash). Vendors MUST check that the 2 characters are either at the end of the string or followed by a dash.
 
     If all three values were parsed successfully, the vendor should use them.
 


spec/20-http_header_format.md
```diff
@@ -214,7 +214,7 @@ Vendors MUST follow these rules when parsing headers with an unexpected format:
     - If the size of the header is shorter than 55 characters, the vendor should not parse the header and should restart the trace.
     - Parse `trace-id` (from the first dash through the next 32 characters). Vendors MUST check that the 32 characters are hex, and that they are followed by a dash (`-`).
     - Parse `parent-id` (from the second dash at the 35th position through the next 16 characters). Vendors MUST check that the 16 characters are hex and followed by a dash.
-    - Parse the `sampled` bit of `flags` (2 characters from the third dash). Vendors MUST check that the 2 characters are either the end of the string or a dash.
+    - Parse the `sampled` bit of `flags` (2 characters from the third dash). Vendors MUST check that the 2 characters are either at the end of the string or followed by a dash.
```
yurishkuro, line 217: I don't think the original text meant "followed by a dash".

dyladan: I interpreted it this way because a future version could extend the header length, and would do so by adding a dash and some new field. In this way it would be backwards compatible.

yurishkuro: but that does not match the current syntax for this field, which _ends_ with the flags byte.

dyladan: but implementations should allow for the possibility of extension. we dont want v1 parsers to break when a v2 or v3 header comes in

mtwo: @dyladan I agree that this would be a good change, but it changes the behaviour of parsers, which we can't modify now

dyladan: What does the original sentence even mean then if not this?

>Vendors MUST check that the 2 characters are either the end of the string or a dash

2 ways to interpret this:
1. Vendors must check that the end of the string is either these 2 characters, or that the end of the string is a dash.
2. Vendors must check that these two characters are either followed by the end of the string, or followed by a dash.

SergeyKanzhelev: I'm on @dyladan side here. Original text seems to be stating that the character after two digits must be dash or end of string

SergeyKanzhelev (approved):

plehegar (approved):

spec/20-http_header_format.md
bogdandrutu, line 217: Should we say:

Parse the `trace-flags` (2 characters from the third dash).

bogdandrutu: I would completely drop this requirement personally (don't know if it is possible).

danielkhan (approved):

PR #380, opened 2020-01-29
Clarify non-normative short trace-id considerations in section 8.4

### LINKED ISSUE #379: Clarify the requirements of short `trace-id` in section 8.4
Several of our engineers have been tripped up by the language in [section 8.4](https://w3c.github.io/trace-context/#interoperating-with-existing-systems-which-use-shorter-identifiers) -- they've interpreted it to mean that they should accept a traceparent with a short traceId, like this:
```
00-24160746bbaecabc-393951cc1a8d5e9e-00
```

There is a test asserting that implementations restart the trace if such a traceparent header is received: https://github.com/w3c/trace-context/blob/master/test/test.py#L303

I'd like us to clarify that the non-normative section 8.4 only applies to accepting short traceIds from legacy non-compliant tracers in a non-w3c payload format, and that it doesn't change the normative traceparent requirements in section 3.2.

justinfoote:
This PR changes section 8.4, which is a non-normative section describing considerations when dealing with `trace-id`.  This PR attempts to clarify that this non-normative section does not change the `trace-id` length requirement specified in section 3.2.  

Fixes #379


yurishkuro (approved):

dyladan (approved):

spec/60-trace-id-format.md
```diff
@@ -73,3 +74,9 @@ and receive an additional flag. Compliant systems will propagate this flag along
 all other key/value pairs. Existing systems which are not capable of
 `tracestate` propagation will truncate all additional values from `tracestate`
 and only pass along that flag.
+
+* This section provides guidance for dealing with short `trace-id`s generated by
```
dyladan, line 78: I think this point is wrong. we are not providing guidance for accepting short trace ids from non-compliant systems, but rather for systems which want to be compliant but can only process an 8 byte id.

@SergeyKanzhelev

justinfoote:
After conversation in the related issue, I'm closing this PR as unnecessary.

PR #322, opened 2019-07-24
Relax tenant id format

mwear:
This PR relaxes the tenant id format so that it can begin with a lowercase letter or number. Previously, it was required to start with a lowercase letter. This change was discussed in issue #316.

My BNF-fu might not be quite up to snuff, so please double check it and let me know if it needs any adjustments.

PR #386, opened 2020-03-18
Clarifies the ABNF on tracestate keys

codefromthecrypt:
I'm not sure if this interpretation is correct, as I'm reverse engineering it from the ABNF. I'm reverse engineering as I expect that folks reading the spec will treat ABNF authoritative vs clerical errors interpreting it.

The main thing is that the change to allow tenant format confused whether or not digits can prefix the basic format. Other changes help better explain the impact and assumptions around tenant format, for example that the intent is to reduce (but not prevent) collisions.

https://github.com/w3c/trace-context/pull/322/files#r393638103

I've allowed edits from maintainers. You may make changes to this and merge it, or close it out with derivative progress. However, I'm not likely to do a lot of revisions. The intent was to force a discussion to fix the underlying issue and I hope you do that one way or another, even if this text is not used at all.

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -278,7 +278,7 @@ keychar    = lcalpha / DIGIT / "_" / "-"/ "*" / "/" / "@"
 lcalpha    = %x61-7A ; a-z
 ```
 
-A `key` MUST begin with a lowercase letter and contain up to 256 characters including lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), forward slashes (`/`), and at signs (`@`).
+A `key` MUST begin with a lowercase letter or a digit and contain up to 256 characters including lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), forward slashes (`/`), and at signs (`@`).
 
 ##### Value
 
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -272,7 +272,6 @@ A `list-member` contains a key/value pair.
 
 The key is an identifier that describes the vendor.
 
-
 ``` abnf
 key = lcalpha 0*255( keychar )
 keychar    = lcalpha / DIGIT / "_" / "-"/ "*" / "/" / "@"
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -270,31 +270,16 @@ A `list-member` contains a key/value pair.
 
 ##### Key
 
-The key is an identifier that describes the vendor. There are two formats allowed, Basic and Tenant format. Both have a maximum length of 256 characters.
+The key is an identifier that describes the vendor.
 
 
 ``` abnf
-basic-key = lcalpha 0*255( keychar )
-tenant-key = ( lcalpha / DIGIT ) 0*240( keychar ) "@" lcalpha 0*13( keychar )
-keychar    = lcalpha / DIGIT / "_" / "-"/ "*" / "/"
+key = lcalpha 0*255( keychar )
+keychar    = lcalpha / DIGIT / "_" / "-"/ "*" / "/" / "@"
 lcalpha    = %x61-7A ; a-z
 ```
 
-###### Basic format
-
-Basic format MUST begin with a lowercase letter and contain lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), and forward slashes (`/`).
-
-###### Tenant format
-
-Tenant format uses an at sign (`@`) to separate a tenant ID from a vendor name. For example, the key `fw529a3039@dt` implies a tenant ID `fw529a3039` and vendor name `dt`.
-
-Tenant format MUST contain a single at sign (`@`). Characters before and after the at sign (`@`) MUST be defined in the Basic format. The up to 241 characters before the at sign (`@`) MUST begin with a lowercase letter or number. The up to 14 characters after the sign (`@`) MUST begin with a lowercase letter.
-
-Vendors SHOULD consider characters before the at sign (`@`) the tenant ID and characters after, the vendor name. Otherwise, a valid tenant ID in one system could be mistaken for the vendor name of another, leading to misinterpretation.
-
-This format was designed to ease locating all keys from the same vendor. For example, the vendor `dt` could use the index of `@dt=` to locate multiple values corresponding to tenants in their system.
-
-Note: The vendor name in Tenant format is the same as Basic format, except it is shorter (maximum 14 instead of 256 characters).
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
I'll make another update as I missed that the "vendor name" in tenant format is effectively a shorter version of the basic format.

Oberon00:
Why such complex restrictions anyway? Why not just add `@` to keychar and use `basic-key = 0*256 ( keychar )`?

BTW 256 (as oppposed to 255) as maximum length is odd, since you need 9 bit to represent 256.

codefromthecrypt:
I agree on both points @Oberon00, it is faster to parse also, if there can be multiple '@'. I had to think a while about why 256, but didn't mention it here.

This change so far is to reverse engineer text on the existing ABNF. To me simple is better, if '@' was just an allowed character, a lot of special casing and explanations go away.

I don't know how many sites are using tenant syntax anyway. I suspect most tracing sites can get away with shared knowledge about the format used or a non-standard convention of the same.

I would have expected at least a few diverse requests from different companies using such a syntax before it became standard. I can't see evidence at least in open source of it used at all, though I'm sure there's at least one commercial product doing tenant@vendor syntax. I do agree if doing such a syntax, one should be consistent, but I also think it was way too early to define special use case practices.

danielkhan:
@SergeyKanzhelev @discostu105 PTAL

discostu105:
We're using the @ symbol in Dynatrace, but I actually see not really a problem with @Oberon00 's suggestion to just allow the '@' character. 👍

codefromthecrypt:
If the overspecification of `@` could be considered an errata, that would be great. I scoured github and there are very few implementations of the header format yet, apart from what folks are doing that are involved in the spec here directly.

basically we can remove the dual text around the practice as it isn't relevant to pretend this is a common practice when there aren't even a lot of impls. this simplifies things quite a bit as especially when `@` is an allowed character, the shortcut approach is invalidated.

I'd prefer when finishing the impl for zipkin's brave client to not have to handle this, and pointing to a spec change that says we don't is the way out.

codefromthecrypt:
PS while OC and OT (and @felixbarny elastic) support pushing tracestate, I only found 2 projects using tracestate on github

@CodingFabian's gang are using this to stash context as described in order to survive a hop they don't own
https://github.com/instana/nodejs-sensor/commit/ef6fdf3b365ca7258afa14af45c55721093e69d7

@SergeyKanzhelev's gang are stashing an app ID in it
https://github.com/microsoft/ApplicationInsights-Java/blob/d1f387f2fc7478aad4ba276ee33d6f3045ca6c4c/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java

Besides Dynatrace who've already spoke up about using tracestate, not sure quorum, but it would be nice for some resolution of folks using this header on the topic of `@`

SergeyKanzhelev (approved):
It looks like a correct interpretation and an improvement for the spec readability.

SergeyKanzhelev:
Any objections to mark it as "non-substantive" as it doesn't change the format?

yurishkuro:
> Why not just add `@` to `keychar` and use `basic-key = 0*256 ( keychar )`?

wouldn't this (a) allow repeated `@` in the key, and therefore (b) break the clause "Otherwise, a valid tenant ID in one system could be mistaken for the vendor name of another, leading to misinterpretation"?

codefromthecrypt:
@yurishkuro on your point, yes. that's what I meant by overspecification. We'd need to delete the whole block on tenant stuff

felixbarny:
AFAIK, our Elastic agents don’t support adding tracestate, we just forward the headers.

basti1302:
We (Instana aka "Fabian's gang") do not give `@` any special meaning nor do we plan to. So +1 from me on removing the special handling of it in the spec.

codefromthecrypt:
As far as I can tell, the consensus here is to simplify the spec and just allow the `@` character. Similar to asterisks (`*`), there is no count restriction. I'll update the PR and thanks for the feedback

codefromthecrypt:
the only other thing I did was clarify the total character count for trace keys in text as someone unfamiliar with ABNF might guess it is 255 not 256

danielkhan (approved):

spec/20-http_request_header_format.md
```diff
@@ -273,15 +273,12 @@ A `list-member` contains a key/value pair.
...
+key = lcalpha 0*255( keychar )
+keychar    = lcalpha / DIGIT / "_" / "-"/ "*" / "/" / "@"
 lcalpha    = %x61-7A ; a-z
 ```
 
-**Note**: Identifiers MUST begin with a lowercase letter or a digit, and can only contain lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), and forward slashes (`/`).
-
-For multi-tenant vendor scenarios, an at sign (`@`) can be used to prefix the vendor name. Vendors SHOULD set the tenant ID at the beginning of the key. For example, \
-`fw529a3039@dt` - `fw529a3039` is a tenant ID and `@dt` is a vendor name. Searching for `@dt=` is more robust for parsing (for example, searching for all a vendor's keys).
+A `key` MUST begin with a lowercase letter and contain up to 256 characters including lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), forward slashes (`/`), and at signs (`@`).
```
justinfoote: This changed
```
MUST begin with a lowercase letter or a digit,
```
to
```
MUST begin with a lowercase letter
```

Was this change intentional?  New Relic, for one, definitely needs tenant IDs that start with digits.

spec/20-http_request_header_format.md
```diff
@@ -273,15 +273,12 @@ A `list-member` contains a key/value pair.
 The key is an identifier that describes the vendor.
 
 ``` abnf
-key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
-key = ( lcalpha / DIGIT ) 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" ) "@" lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+key = lcalpha 0*255( keychar )
```
dyladan: You changed the semantics of the vendor key to not allow the first character to be a digit. This is important for at least the newrelic folks.

```suggestion
key = ( lcalpha / DIGIT ) 0*255 ( keychar )
```

edit: disclaimer I do not work at newrelic and do not speak for them :)

dyladan (changes requested):

spec/20-http_request_header_format.md
justinfoote: To align with @dyladan's suggestion above:

```suggestion
A `key` MUST begin with a lowercase letter or a digit and contain up to 256 characters including lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), forward slashes (`/`), and at signs (`@`).
```

codefromthecrypt:
thanks for the help with the digit thing @dyladan @justinfoote I was going to raise that today because my unit tests broke :) appreciate you saving me the effort.

dyladan (approved):

justinfoote (approved):

PR #417, opened 2020-07-28
Check if keys exist before trying to access them

### LINKED ISSUE #416: Fix key errors in test suite
When running the tests, you can get a failure due to the tests checking for keys in an empty dictionary. Below, is an example of the error message. Because no tracestate was passed along by the server, the tracestate map is empty and causes this error. I think it's kind of confusing to have both Failure and Errors states that indicate a failure. I think we should just check if the keys exist prior to accessing them.

```
ERROR: test_tracestate_included_traceparent_included (__main__.TraceContextTest)
harness sends a request with both tracestate and traceparent
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test.py", line 785, in test_tracestate_included_traceparent_included
    self.assertEqual(tracestate['foo'], '1')
  File "/Users/stevenrivera-james/workspace/test_pr/lstracer-integration-tests/propagation-tests/propagation-server/tracecontext/tracestate.py", line 35, in __getitem__
    return self._traits[key]
KeyError: 'foo'
```

srjames90:
Fixes #416

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -445,6 +445,8 @@ class TraceContextTest(TestBase):
 			['tracestate', 'foo=1,bar=2'],
 		])
 		self.assertEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')
+		self.assertIn("foo", tracestate)
+		self.assertIn("bar", tracestate)
 		self.assertEqual(tracestate['foo'], '1')
 		self.assertEqual(tracestate['bar'], '2')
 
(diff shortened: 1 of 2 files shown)


reyang (approved):
LGTM.

SergeyKanzhelev (approved):

SergeyKanzhelev:

bogdandrutu (approved):

mtwo (approved):

PR #421, opened 2020-09-01
Improve wording on tracestate mutations

arminru:
"Update key value" sounds like changing the value of the key itself to me. Apart from that, "key value" is not aligned with the other occurrences saying "key/value pair". Ordering as create->update->delete also makes more sense IMHO since creating has to come first before one can update a value.
I also removed this non-normative and unclear statement:
> This is the most common mutation resuming the trace.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/pull/421.html" title="Last updated on Sep 2, 2020, 2:46 PM UTC (6e72191)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/421/36192a7...6e72191.html" title="Last updated on Sep 2, 2020, 2:46 PM UTC (6e72191)">Diff</a>


spec/20-http_request_header_format.md
```diff
@@ -352,6 +352,6 @@ Vendors receiving a `tracestate` request header MUST send it to outgoing request
 
 Following are allowed mutations:
 
-- **Update key value**. The value of any key can be updated. Modified keys MUST be moved to the beginning (left) of the list. This is the most common mutation resuming the trace.
-- **Add a new key/value pair**. The new key-value pair SHOULD be added to the beginning of the list.
+- **Add a new key/value pair**. The new key/value pair SHOULD be added to the beginning of the list.
+- **Update an existing value**. The value for any given key can be updated. Modified keys MUST be moved to the beginning (left) of the list. This is the most common mutation resuming the trace.
```
anuraaga: This isn't really related to the PR but maybe while we're here, I don't understand what resuming the trace means here. Anyways since a spec probably can't predict what is more common or not, maybe can remove the sentence?

arminru: I read this as participating in an existing trace by continuing it as opposed to starting off a new one.
I agree that this does not really fit into a technical specification but rather a usage guideline, for example, and would be fine with just removing that sentence.

dyladan: I think the sentence can just be removed as it is not a technical detail of the spec

arminru: Removed in 6e72191.

spec/20-http_request_header_format.md
```diff
@@ -352,6 +352,6 @@ Vendors receiving a `tracestate` request header MUST send it to outgoing request
 
 Following are allowed mutations:
 
-- **Update key value**. The value of any key can be updated. Modified keys MUST be moved to the beginning (left) of the list. This is the most common mutation resuming the trace.
-- **Add a new key/value pair**. The new key-value pair SHOULD be added to the beginning of the list.
+- **Add a new key/value pair**. The new key/value pair SHOULD be added to the beginning of the list.
```
dyladan, line 355: You changed the meaning of this one.

```suggestion
- **Add a new key/value pair**. The new key/value pair MUST be added to the beginning of the list.
```

reyang (approved):
LGTM.

spec/20-http_request_header_format.md
arminru: No I just reordered the items to create->update->delete since creating has to come first before one can update a value.

dyladan: Ah I see. It is confusing that an updated key MUST be moved to the left, but a newly added key only SHOULD be added on the left. If you delete your key, but do not follow the optional (SHOULD) for re-adding a new key, haven't you made a spec-noncompliant edit? I think this is a contradiction in the spec.

In any case, it is fine here as that is not the issue.

dyladan (approved):

danielkhan (approved):

PR #423, opened 2020-09-09
Remove contradiction in TraceState mutations

### LINKED ISSUE #422: Contradiction in TraceState requirements
It is confusing that an updated key MUST be moved to the left, but a newly added key only SHOULD be added on the left. If you delete your key, then add your key somewhere other than the start of the list (because that requirement is only SHOULD), haven't you edited the key without moving it left? I think this is a contradiction in the spec.

_Originally posted by @dyladan in https://github.com/w3c/trace-context/pull/421#r482283902_

dyladan:
The original contradiction:

1. Modified keys MUST be moved to the left of the list
2. Deletion is allowed
3. New keys SHOULD be added on the left, which implies they MAY be added
   somewhere else.
4. Deleting a key and adding it somewhere other than the left of the
   list breaks (1).

Fixes #422

When this fix is merged, it can be backported to level 1.

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/423.html" title="Last updated on Sep 9, 2020, 5:30 PM UTC (e0103bf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/423/5a99a94...dynatrace-oss-contrib:e0103bf.html" title="Last updated on Sep 9, 2020, 5:30 PM UTC (e0103bf)">Diff</a>


danielkhan (approved):
LGTM

yurishkuro:
Just to cover the bases for rationale - why not go for MUST in both "add" and "modify" cases?

I am also curious, now that it's been almost 2yrs since this was introduced, did any implementation actually take a dependency on the ordering of the keys? The original motivation for the "head of the list' rule was to allow the next hop to determine if the previous hop was instrumented by the same vendor. Does it matter in practice?  Worth noting that if a node is adding more than one key, then the whole ordering argument becomes less strong, since it can still add keys to the left in _any_ order.

dyladan:
@yurishkuro here's what we talked about at the meeting yesterday:

Changing requirements to MUST brings up a few issues. First, it disallows a usecase identified by @SergeyKanzhelev which is to have some sort of security sweeper which strips potentially security sensitive data from values. Second, it will always be possible to delete all keys and reinsert them in whatever order you want, which makes MUST requirements unenforceable here. Third, MUST would be a backwards incompatible change. There is also already wording in the spec regarding order-preservation of the keys.

Given these things, at the meeting we decided that SHOULD is the best way to move forward for now.

dyladan:
> I am also curious, now that it's been almost 2yrs since this was introduced, did any implementation actually take a dependency on the ordering of the keys? The original motivation for the "head of the list' rule was to allow the next hop to determine if the previous hop was instrumented by the same vendor. Does it matter in practice? Worth noting that if a node is adding more than one key, then the whole ordering argument becomes less strong, since it can still add keys to the left in _any_ order.

One important usecase which depends on key ordering is dropping keys if the header grows too long. Dropping the oldest keys requires order-preservation in order to determine relative age.

yurishkuro (approved):

mtwo (approved):

Issue #406, opened 2020-03-27
Remove proposed-parent-id from response header


dyladan:
After feedback from @adriancole and conversations during the march online workshop, it was decided to remove the proposed parent id for now.

yurishkuro:
Was there an alternative mechanism proposed to solve the original "web browser" issue? Please document it here, for posterity.

dyladan:
There is not. The issue is that there currently wouldn't even be a mechanism in browsers to use this feature, so until we get sign-on from browser vendors it won't be useable. You can't access arbitrary headers from the initial page load.

yurishkuro:
I was under the impression that this was an already existing use case, e.g. traceparent usage described here: https://github.com/census-instrumentation/opencensus-web

dyladan:
There are a handful of methods including cookies, server-timing, and injecting meta tags in web pages for communicating proposed parent ids back to the caller, but as far as I can tell, no API exists to read arbitrary response headers from the initial document load request.

dyladan:
This has been discussed in several of the bi-weekly meetings now. I'm updating this issue to the current state of the discussion.

We are currently hoping to keep this use-case in, but it depends on adding new browser APIs. We would like to solidify the use-case and try to see if browsers would implement this. If we cannot get browser adoption, there is no way to access the proposed parent from within the browser and the field should be removed.

codefromthecrypt:
TL;DR; I think this spec by adding features like proposed parent ID jumped too far without enough eyes on the problem. This is very much implementation not api. cc @cwensel and @jcchavezs

we can't say there's no solution as there have been very few independent minds on the problem, or folks offering solutions. It is moreover hard to help anyone solve this as discussions continue with no visible code to help people with. I don't understand how something can be added to a spec based on something no one can see. you just have to take someone's word on it? This should have an example code so people can help you.

Anyway, there's a massive difference between returning the context that serviced the request (its ids) and generating a "proposed parent ID" for a potential instrumented client which may never exist. The latter is something that shouldn't be in the spec at all. There are other ways, even if vendors doing browser based tracing today are not sharing their processes.

Here's one for zipkin:

In zipkin, only traceId and spanId are immutable fields. This means parentId is an updatable field: you can patch it the same as adding a late tag. It is just a field.

If you had the server's context (it thinks is root) sent back to you as `b3: 0412f83c4ae74844-0412f83c4ae74844-1`

[...]

PR #425, opened 2020-09-22
Create first draft of the use cases document

danielkhan:
This is the first shot at a document that collects use-cases for response headers.

### DIFF
--- a/response-headers-usecases.md
+++ b/response-headers-usecases.md
@@ -1,0 +1,13 @@
+# Use Cases for response Headers
+
+The [Response Header Specification](https://github.com/w3c/trace-context/blob/master/spec/21-http_response_header_format.md) defines the format for response headers.
+
+As response headers solve (largely) different use cases, this doc aims to collect them and also address possible implications, like privacy issues.
+
+This document is a loose, unordered and ongoing collection.
+
+## Usecase 1: Report the proposed parent ID back to the caller 
+
+A system, like a browser, can not create a trace ID and span ID.
+The callee could generate the parent ID and then send it back to the caller via response header.
+This parent ID would represent the parent of the spans created by the callee.


mtwo (approved):

bogdandrutu (approved):

response-headers-usecases.md
```diff
@@ -0,0 +1,13 @@
+# Use Cases for response Headers
+
+The [Response Header Specification](https://github.com/w3c/trace-context/blob/master/spec/21-http_response_header_format.md) defines the format for response headers.
+
+As response headers solve (largely) different use cases, this doc aims to collect them and also address possible implications, like privacy issues.
+
+This document is a loose, unordered and ongoing collection.
+
+## Usecase 1: Report the proposed parent ID back to the caller
```
jcchavezs, line 9: While it is good you describe use cases for such things I believe those should be self described. For example, why some systems can't create a trace ID and span ID, why the caller would need a parent ID? Does the parent ID is enough info without the trace ID?

Use case as it is feels like describing the what and not the why.

codefromthecrypt:
https://github.com/w3c/trace-context/issues/406 was intending to remove exactly this

dyladan (approved):

response-headers-usecases.md
codefromthecrypt: This topic has been well discussed and is niche at best. It doesn't deserve the "usecase 1" if anything "honorable mention"

Returning trace IDs is far more frequent for putting traces together without this odd notion of "proposed parent ID". Maybe you can put more routine things here by looking at the issues leading up to this or asking users.

For example, this issue from over 4 years ago catalogs some ideas with real people who were interested https://github.com/openzipkin/b3-propagation/issues/4

response-headers-usecases.md
```diff
@@ -0,0 +1,13 @@
+# Use Cases for response Headers
+
+The [Response Header Specification](https://github.com/w3c/trace-context/blob/master/spec/21-http_response_header_format.md) defines the format for response headers.
+
+As response headers solve (largely) different use cases, this doc aims to collect them and also address possible implications, like privacy issues.
+
+This document is a loose, unordered and ongoing collection.
+
+## Usecase 1: Report the proposed parent ID back to the caller 
+
```
Kanwaldeep, line 10: Here is another use case - Currently some of the Azure Control Plane APIs return Correlation Id as part of the response header which can then be used by the customer on reporting an issue or contacting us. We would like to have teams transition to using the standard W3C trace content and returning this as standard response header. Having the trace context in the form of the traceresponse header would make this easy for customers and enable us to pinpoint issues.

PR #426, opened 2020-10-06
Add wording suggesting random generation of span ids

### LINKED ISSUE #412: What's the recommended guideline for generating parent-id?
From here - https://www.w3.org/TR/trace-context/#parent-id

_This is the ID of this request as known by the caller (in some tracing systems, this is known as the span-id, where a span is the execution of a client request). It is represented as an 8-byte array, for example, 00f067aa0ba902b7. All bytes as zero (0000000000000000) is considered an invalid value._

_Vendors MUST ignore the traceparent when the parent-id is invalid (for example, if it contains non-lowercase hex characters)._

Current guidelines for parent-id mentions this should be an 8-byte array. But, we are missing guidelines/recommendations that this should be locally unique within the context of a trace? Also, how have SDKs implemented this to ensure span-id is unique within the context of a trace? We need to elaborate the guidelines similar to trace-id - https://www.w3.org/TR/trace-context/#considerations-for-trace-id-field-generation

dyladan:
Fixes #412

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/426.html" title="Last updated on Nov 3, 2020, 8:26 PM UTC (086b858)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/426/5a99a94...dynatrace-oss-contrib:086b858.html" title="Last updated on Nov 3, 2020, 8:26 PM UTC (086b858)">Diff</a>

### DIFF
--- a/spec/61-span-id-format.md
+++ b/spec/61-span-id-format.md
@@ -1,0 +1,20 @@
+## Considerations for `span-id` field generation
+
+This section suggests some practices to consider when implementing
+`span-id` generation algorithms to ensure
+interoperability between different systems.
+
+### Uniqueness of `span-id`
+
+The value of `span-id` MUST be unique within a <a>distributed trace</a>.
+If the value of `span-id` is not unique within a distributed trace,
+parent-child relationships between spans within the distributed trace
+may be ambiguous.
+
+### Randomness of `span-id`
+
+Values of `span-id` SHOULD be randomly generated. Randomness of `span-id`
+addresses some [security](#security-considerations) and [privacy
+concerns](#privacy-considerations) of exposing unwanted information.
+Randomness also ensures a high probability, though not a guarantee, of
+uniqueness within a distributed trace.
--- a/spec/61-span-id-format.md
+++ b/spec/61-span-id-format.md
@@ -6,7 +6,7 @@ interoperability between different systems.
 
 ### Uniqueness of `span-id`
 
-The value of `span-id` MUST be unique within a <a>distributed trace</a>.
+The value of `span-id` SHOULD be unique within a <a>distributed trace</a>.
 If the value of `span-id` is not unique within a distributed trace,
 parent-child relationships between spans within the distributed trace
 may be ambiguous.
--- a/index.html
+++ b/index.html
@@ -96,6 +96,7 @@
   <section data-include="spec/51-security.md" data-include-format='markdown'></section>
 
   <section class="informative" data-include="spec/60-trace-id-format.md" data-include-format='markdown'></section>
+  <section class="informative" data-include="spec/61-span-id-format.md" data-include-format='markdown'></section>
 
   <section class="appendix" data-include="spec/60-acknowledgments.md" data-include-format='markdown'></section>
 


spec/61-span-id-format.md
```diff
@@ -0,0 +1,20 @@
+## Considerations for `span-id` field generation
+
+This section suggests some practices to consider when implementing
+`span-id` generation algorithms to ensure
+interoperability between different systems.
```
bogdandrutu, line 5: nit, no need of a new line.

mtwo (approved):

spec/61-span-id-format.md
```diff
@@ -0,0 +1,20 @@
...
+parent-child relationships between spans within the distributed trace
+may be ambiguous.
+
+### Randomness of `span-id`
+
+Values of `span-id` SHOULD be randomly generated. Randomness of `span-id`
+addresses some [security](#security-considerations) and [privacy
+concerns](#privacy-considerations) of exposing unwanted information.
+Randomness also ensures a high probability, though not a guarantee, of
+uniqueness within a distributed trace.
```
bogdandrutu, line 20: Majority of the concerns are related to the "trace-id" which identifies a "request". Not sure if this justify this.

Kanwaldeep (approved):
LGTM

danielkhan (approved):
LGTM

spec/61-span-id-format.md
dyladan: This wording is just to discourage people from using PII in span ids. It is more of a concern with trace ids, but I think it still applies to span ids.

spec/61-span-id-format.md
```diff
@@ -0,0 +1,20 @@
+## Considerations for `span-id` field generation
+
+This section suggests some practices to consider when implementing
+`span-id` generation algorithms to ensure
+interoperability between different systems.
+
+### Uniqueness of `span-id`
+
+The value of `span-id` MUST be unique within a <a>distributed trace</a>.
```
yurishkuro:
```suggestion
The value of `span-id` SHOULD be unique within a <a>distributed trace</a>.
```
there is no way to guarantee it using random generators

spec/61-span-id-format.md
yurishkuro: I'm ok with this because it says SHOULD, leaving the door open for different algorithm. In general, I think the spec should be less opinionated about how IDs are generated, because nunyabiznis, and should be just talking about the representation.

danielkhan (approved):
LGTM

danielkhan (approved):

spec/61-span-id-format.md
yurishkuro: MUST here is a deal breaker for me

SergeyKanzhelev: even though it is informative section I tend to agree that MUST here is too strong of a requirement

SergeyKanzhelev (changes requested):
we need to change wording of MUST in span generation. Something like strive to. Must might be too aggressive

danielkhan:
@dyladan PTAL

yurishkuro (approved):

SergeyKanzhelev (approved):

Issue #390, opened 2020-03-24
Bogus trace and parent IDs when not sampling  


basti1302:
### Problem

I recently [implemented this specification](https://github.com/instana/nodejs-sensor/pull/204) for [Instana's Node.js tracer](https://github.com/instana/nodejs-sensor) and while doing so, came across a situation that might have room for improvement:

- A process instrumented by a tracing tool receives an HTTP request without the `traceparent` and `tracestate` header. 
- The tracing tool decides to _not_ record trace data.
- The instrumented process makes outgoing HTTP requests in the context of the initial incoming request.

Since there are not trace context headers in the incoming request, we can assume that this is the first process in line that is instrumented by a tracing tool, that is, the request is coming from a non-instrumented client. 

**What is the tracing tool supposed to do in this situation?**

According to [§3.2.2.5.1](https://www.w3.org/TR/trace-context/#sampled-flag), the decision to not record trace data SHOULD be communicated downstream:

> If a component made definitive recording decision - this decision SHOULD be reflected in the sampled flag.

This also implies that the tracing tool SHOULD send the `traceparent` header on outgoing requests, as there are not other means of communicating the recording decision downstream.

For a valid `traceparent` header, the tracing tool needs to set the `trace-id` and `parent-id` fields of that header. If it had received a `traceparent` header, it would simply use the IDs from that incoming header, but since the initial request did not have that header, this is not possible.

From my point of view, a compliant tracing tool can only choose between two options:
1. Create new bogus trace-id and parent-id values and send those downstream, together with the trace flags value `00`.
1. Do not set a `traceparent` header at all on outgoing requests.

The first option is quite ugly, as there now is a trace-id and a parent-id floating around that are not associated with any actual trace and span that is present in any of the participating APM systems. 
 
The second options is also not optimal because it does not allow to pass down the recording decision. It is also in conflict with the (non-normative) section [§4.2 No traceparent Received](https://www.w3.org/TR/trace-context/#no-traceparent-received), which says:

> the vendor creates a new trace-id and parent-id that represents the current request. [...]
> The vendor sets the traceparent and tracestate header for the outgoing request.

I would like to discuss/ask:

- whether this understanding of the spec is correct and shared by others (or if I have missed something),
- what other existing implementations are doing in this particular situation.

#### Do Not Send `traceparent`/`tracestate`

The spec could mandate explicitly to not send `traceparent` and `tracestate` at all in this particular situation. 

[...]

PR #433, opened 2020-10-27
Clarify behaviour when starting a trace and not recording

basti1302:
This is intended to "fix" #390.

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/433.html" title="Last updated on Dec 15, 2020, 5:56 PM UTC (40cf086)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/433/6d95bd9...instana:40cf086.html" title="Last updated on Dec 15, 2020, 5:56 PM UTC (40cf086)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -176,6 +176,8 @@ There are two additional options that vendors MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+**Note**: Communicating the recording decision downstream via the `sampled` flag obviously requires adding the `traceparent` header to outgoing requests in the first place. In cases where the incoming request did not have the `traceparent` header and the tracing tool decides not to record trace data and to communicate that decision downstream, this still entails generating a `trace-id` and `parent-id` (that is, starting a trace), even if those id values are not associated with any actual trace data. It is also valid behavior to not start a new trace (that is, not adding the traceparent headers to outgoing requests), and accept the fact that the decision to not record trace data cannot be communicated downstream.
+
 ##### Other Flags
 
 The behavior of other flags, such as (`00000100`) is not defined and is reserved for future use. Vendors MUST set those to zero.
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -176,8 +176,6 @@ There are two additional options that vendors MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
-**Note**: Communicating the recording decision downstream via the `sampled` flag obviously requires adding the `traceparent` header to outgoing requests in the first place. In cases where the incoming request did not have the `traceparent` header and the tracing tool decides not to record trace data and to communicate that decision downstream, this still entails generating a `trace-id` and `parent-id` (that is, starting a trace), even if those id values are not associated with any actual trace data. It is also valid behavior to not start a new trace (that is, not adding the traceparent headers to outgoing requests), and accept the fact that the decision to not record trace data cannot be communicated downstream.
-
 ##### Other Flags
 
 The behavior of other flags, such as (`00000100`) is not defined and is reserved for future use. Vendors MUST set those to zero.
--- a/spec/30-processing-model.md
+++ b/spec/30-processing-model.md
@@ -11,7 +11,7 @@ This processing model describes the behavior of a vendor that modifies and forwa
 If no traceparent header is received:
 
 1. The vendor checks an incoming request for a `traceparent` and a `tracestate` header.
-2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request. (Note: If the vendor does not sample this request and wants to communicate that sampling decision downstream via the `sampled` flag, it is acceptable to create a `trace-id` and `parent-id` that are not associated with any actual trace data. Not communicating the sampling decision downstream is equally acceptable.)
+2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request. (Note: If the vendor does not sample this request and wants to communicate that sampling decision downstream via the `sampled` flag, the vendor MAY create a `trace-id` and `parent-id` that are not associated with any actual trace data. The vendor MAY also decide to not communicate the sampling decision downstream.)
 3. If a `tracestate` header is received without an accompanying `traceparent` header, it is invalid and MUST be discarded.
 4. The vendor SHOULD create a new `tracestate` header and add a new key/value pair.
 5. The vendor sets the `traceparent` and `tracestate` header for the outgoing request.
--- a/spec/30-processing-model.md
+++ b/spec/30-processing-model.md
@@ -11,7 +11,7 @@ This processing model describes the behavior of a vendor that modifies and forwa
 If no traceparent header is received:
 
 1. The vendor checks an incoming request for a `traceparent` and a `tracestate` header.
-2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request.
+2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request. (Note: Even if deciding to not sample the request, generating IDs is important to be able to propagate that sampling decision downstream via the sampled flag.)
 3. If a `tracestate` header is received without an accompanying `traceparent` header, it is invalid and MUST be discarded.
 4. The vendor SHOULD create a new `tracestate` header and add a new key/value pair.
 5. The vendor sets the `traceparent` and `tracestate` header for the outgoing request.
--- a/spec/30-processing-model.md
+++ b/spec/30-processing-model.md
@@ -11,7 +11,7 @@ This processing model describes the behavior of a vendor that modifies and forwa
 If no traceparent header is received:
 
 1. The vendor checks an incoming request for a `traceparent` and a `tracestate` header.
-2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request. (Note: Even if deciding to not sample the request, generating IDs is important to be able to propagate that sampling decision downstream via the sampled flag.)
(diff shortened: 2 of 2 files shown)


spec/20-http_request_header_format.md
```diff
@@ -176,6 +176,8 @@ There are two additional options that vendors MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+**Note**: Communicating the recording decision downstream via the `sampled` flag obviously requires adding the `traceparent` header to outgoing requests in the first place. In cases where the incoming request did not have the `traceparent` header and the tracing tool decides not to record trace data, this still entails generating a `trace-id` and `parent-id` (that is, starting a trace), even if those id values are not associated with any actual trace data.
```
SergeyKanzhelev: side note: trace data is a good candidate to be added to the glossary. It is mentioned in spec couple times.

SergeyKanzhelev (approved):
LGTM. This PR is not changing semantics from what I understand

spec/10-overview.md
```diff
@@ -32,7 +32,7 @@ Trace context is split into two individual propagation fields supporting interop
 
 Tracing tools can provide two levels of compliant behavior interacting with trace context:
 
-- At a minimum they MUST propagate the `traceparent` and `tracestate` headers and guarantee traces are not broken. This behavior is also referred to as forwarding a trace.
-- In addition they CAN also choose to participate in a trace by modifying the `traceparent` header and relevant parts of the `tracestate` header containing their proprietary information. This is also referred to as participating in a trace.
+- At a minimum they MUST propagate the `traceparent` and `tracestate` headers from incoming requests to all outgoing requests, guaranteeing that traces are not broken. This behavior is also referred to as forwarding a trace.
+- In addition they CAN also choose to participate in a trace by modifying the `traceparent` header and relevant parts of the `tracestate` header containing their proprietary information. This is also referred to as participating in a trace. Participating tracing tools SHOULD also add the `traceparent` header to all outgoing requests, even when the associated incoming request did not have it (this is referred to as starting a trace).
```
basti1302: If we are worried about changing semantics, then this addition here, that participating tools SHOULD start a trace when the incoming request did not have W3C trace context headers, might need to be discussed a bit. From my point of view it is in line with the intention of the spec, but the spec never spelled that out explicitly before (or I might have missed it). 

Do we want to mandate always start a trace? Is SHOULD appropriate here (vs. MAY or MUST)? 

WDYT?

spec/20-http_request_header_format.md
basti1302: Tracing tool, too, I think.

spec/10-overview.md
SergeyKanzhelev: this is when tool decide to participate. Tool may decide not to participate in trace and do nothing. So no change for the spec from my POV

basti1302: Sure, this is only for participating tools. My point was that even for participating tools we didn't mandate to always start a trace until now. I think that was the mental model most folks had about this aspect, but it wasn't spelled out explicitly, I think.

But if you (and others) agree that this does not change semantics, I'm absolutely on board with that view. 👍

yurishkuro: The section you are changing starts with "interacting with trace context", which seems to imply there is an inbound context. If you're looking for completeness, there should be a separate section onwhat to do when there is no incoming trace context. In that situation both starting a trace and doing nothing should be valid options, IMO, i.e. the spec should not dictate starting a new trace.

spec/20-http_request_header_format.md
yurishkuro: -1 to this addition. A component should be free to not send trace context downstream if there was non from upstream.

basti1302: Isn't that somewhat contradictory to what you said [here](https://github.com/w3c/trace-context/issues/390#issuecomment-603356166)? 

How about phrasing it like this then (new parts in bold):

**Note**: Communicating the recording decision downstream via the `sampled` flag obviously requires adding the `traceparent` header to outgoing requests in the first place. In cases where the incoming request did not have the `traceparent` header and the tracing tool decides not to record trace data **and to communicate that decision downstream**, this still entails generating a `trace-id` and `parent-id` (that is, starting a trace), even if those id values are not associated with any actual trace data. **It is also valid behavior to not start a new trace (that is, not adding the traceparent headers to outgoing requests), and accept the fact that the decision to not record trace data cannot be communicated downstream.**

spec/10-overview.md
basti1302: > The section you are changing starts with "interacting with trace context", which seems to imply there is an inbound context. 

Hm, I don't read it that way as starting a new trace is also a form of interacting with the trace context. But anyway, having this in a separate section might be better.

> IMO, i.e. the spec should not dictate starting a new trace.

There seems to disagreement about that? I understood @SergeyKanzhelev comments as being in support of that notion, that to _participate_ also means to _start_ a trace if there is none.

Given that we have the phrase "A tracing tool can choose to change this behavior for each individual request to a component it is monitoring." there is alway the option to act as a non-participating tool for that request.  

It is probably best to remove this change from this PR entirely. I might try to come up with a separate section about how to behave when there is no incoming trace context. I think agreeing on some explicit rules about that situation and adding them to the spec would be good, but it's out of scope for this change.

spec/20-http_request_header_format.md
SergeyKanzhelev: I'm not against the originally proposed text as it still has some room for not doing it. It is still the actor's decision. But the text suggested in the comment is even better.

Overall the feedback was before that we need to minimize the amount of actor's behavior w.r.t. sampling in the normative section as this spec is mostly about the protocols and do not regulate how sampling will work. @yurishkuro are you partially against this section because you want to minimize the amount of details on specific sampling behavior? Maybe good alternative would be to move this notice into the non-normative section [processing model](https://github.com/w3c/trace-context/blob/master/spec/30-processing-model.md)?

basti1302: It says nothing about sampling at all though, does it? "Starting a trace" just means sending headers downstream in this context.

yurishkuro: > Isn't that somewhat contradictory to what you said here?

It is, indeed. New company - new perspective :-). There is a use case that even generating a new random ID is an extra unnecessary overhead for very high QPS services if the trace is not going to be sampled.

spec/10-overview.md
basti1302: I removed this change.

yurishkuro (dismissed):

spec/20-http_request_header_format.md
basti1302: Since Yuri now approved the changes, I guess the new phrasing resolved his concerns. Regarding the suggestion to move this into the processing model: This is very specific to the sampled flag, which is talked about here in detail, so I think it would be better to keep this particular notice in this section.

dyladan (commented):
While not a problem, this seems like a lot of words to add to an already very wordy specification. Since this is non-normative, I think it would be better to extend [Processing Model > no traceparent received](https://github.com/w3c/trace-context/blob/master/spec/30-processing-model.md#no-traceparent-received) with reasoning.

Simply adding a sentence to step (2) to the effect of "This step is important even if the request is not sampled in order to propagate the sampling decision with the traceparent header," should be sufficient.

basti1302:
Since both @SergeyKanzhelev and @dyladan requested to move this to the processing model section, I added it there now, pretty much with the phrasing Daniel suggested.

To be honest, I like this a little less than the previous version because it does no longer make it explicit that having bogus IDs that are not connected to any existing trace data is acceptable. It follows implicitly from the processing model, though, so it might be good enough; and if this version is deemed more acceptable by the rest of the group, I'm fine with it 👍

anuraaga:
Sorry for the flyby - I don't find the wording to be clear since propagating a sampling decision doesn't seem related to generating IDs. Generating invalid IDs to propagate a sampling decision and nothing else seems like a reasonable way to propagate a sampling decision. Really, just one char is required to propagate a non-sampled decision and it's unfortunate w3c doesn't allow this optimization. Ideally it could at least allow "any id" orthogonally from a sampling decision?

yurishkuro:
I think sampling decision does not exist in isolation. If that decision is YES, it is meaningless without IDs. If it is NO, it can be communicated with or without IDs, where without IDs means not sending any header at all.

basti1302:
> I think sampling decision does not exist in isolation. If that decision is YES, it is meaningless without IDs. If it is NO, it can be communicated with or without IDs, where without IDs means not sending any header at all.

I disagree with the last part. You cannot communicate the decision to not sample by _not heading headers_. To any downstream service this behavior will look the same as if there is no instrumented (W3C trace context compliant) service upstream at all.

The whole point of the sampling flag, as i understand it, is so an upstream service can tell downstream services that it decided not to record data and the downstream service can factor that into its own decision whether to record data or not. So not receiving any trace context header or receiving header with sampled=false has a different semantic.

yurishkuro:
Yes, it is semantically different, but still acceptable imo. It really depends on the implementation. Some implementations may choose to start a new trace when not receiving headers, others may treat it as "do not sample" indicator (e.g. when only certain layers in the architecture are allowed to start traces).

danielkhan (approved):
As this is now moved to the processing model which is non-normative, I think it's a fair point. One might still choose to handle this differently without acting against the spec.

danielkhan:
Suggestion: 

> Note: Even if deciding to not sample the request, generating IDs to be able to propagate that sampling decision downstream via the sampled flag may be considered.

basti1302:
Rephrased once more as dicsussed in last meeting.

spec/30-processing-model.md
```diff
@@ -11,7 +11,7 @@ This processing model describes the behavior of a vendor that modifies and forwa
 If no traceparent header is received:
 
 1. The vendor checks an incoming request for a `traceparent` and a `tracestate` header.
-2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request.
+2. Because the `traceparent` header is not received, the vendor creates a new `trace-id` and `parent-id` that represents the current request. (Note: If the vendor does not sample this request and wants to communicate that sampling decision downstream via the `sampled` flag, it is acceptable to create a `trace-id` and `parent-id` that are not associated with any actual trace data. Not communicating the sampling decision downstream is equally acceptable.)
```
yurishkuro: I agree with this, but it may be worth rephrasing using normative words, e.g. "the vendor MAY decide to not sample this request ... and MAY communicate ..."

yurishkuro (approved):

dyladan (approved):

Issue #407, opened 2020-03-27
Clarify use of trace-id in response header


dyladan:
After feedback from @adriancole it became clear that the use-case for trace-id is not well defined in the spec. Update the wording to make it clear that the caller is not meant to adopt the trace id.

yurishkuro:
The use case for returning trace ID is when cloud provider returns an internal trace reference. What is the proposed alternative for that use cae?

dyladan:
There is no proposed alternative. The issue is that the use case you stated wasn't clear to everyone. This issue is just to remind me to take a look through the spec and make sure it's clear that this is the intended use case.

yurishkuro:
My understanding is that the use case is:
  * Client sends a request to a managed service in the cloud. Client may or may not include traceparent.
  * Cloud provider internally starts another trace, and returns its ID in the response header.
  * Client records the response trace ID in its own span/trace, or logs it, so that it can be given to cloud customer support for troubleshooting.

When I say "trace ID", it may also include the span ID, if the cloud provider wants to have more precise reference (perhaps in some messaging / serverless scenarios the reference is not always to the root), but that's a speculation.

Issue #398, opened 2020-03-25
Should response header fields be optional?


dyladan:
https://github.com/w3c/trace-context/pull/365#discussion_r377947060

yurishkuro:
I think they certainly should be optional. We have two use cases involving 3 IDs: self trace ID, self span ID, and proposed parent ID. Both use cases themselves are optional and don't make sense for every service (e.g. for a service in the middle of your own architecture returning a response header is almost always a waste). I expect the two use cases to be at minimum configurable in the SDK, and perhaps even triggered by something in the request.

dyladan:
I also believe they should be optional. Some have mentioned that making them optional adds complexity, but I think it is worthwhile.

dyladan:
AFAIK @SergeyKanzhelev had some concern here?

dyladan:
Discussed at the bi-weekly meeting. The discussion turned from "should they be optional" to "if they're optional, how do you decide which fields to include?"

I will attempt to make a set of rules for how to decide which fields to include. If a set of rules cannot be made reasonably and reliably, the fields cannot be optional.

dyladan:
At last week's meeting, I volunteered to think about a set of rules that could be applied to determine which fields should be returned in the response header. 

## Uses of fields

Here is a quick summary of the fields available and their suggested uses.

### Version

Ensure header is parsed correctly. Always required if the header is included in the response.

### Trace ID

When called by a third party, this may be used to communicate the recorded trace ID to the caller for support purposes.

When called by a load balancer or web browser, this maybe be used to provide a trace ID to the caller so they may participate in the trace.

### Proposed Parent Id

When called by a load balancer or web browser, this maybe be used to provide a span ID to the caller so they may participate in the trace.

### Trace Flags

When called by a third party, this may be used to communicate information about the trace such as the sampled status for support purposes.

When called by a load balancer, this may be used to communicated information such as the sampled status so that the caller may intelligently decide to record its part of the trace or not.

## Suggested decision model

Of course, the header itself and all fields (except version) are always optional. If a service provider wishes not to return the response header for any reason, they are not obligated to do so. The following rules are only a recommended set of rules to encourage interoperation between tracing providers.

### Default behavior

#### Incoming call with traceparent

If the trace ID of the traceparent is used, do not include the header in the response.

If a new trace ID is used (trace restarted), include the version and the trace ID in the response.

#### Incoming call without traceparent 

Return version, trace ID, and trace flags.

[...]

PR #443, opened 2020-12-15
Replace proposed-parent-id with child-id. All fields required.

### LINKED ISSUE #406: Remove proposed-parent-id from response header

### LINKED ISSUE #437: Encourage traceresponse to be provided even when traceparent is used
The current proposal for traceresponse says, "The trace-id field is an optional part of the traceresponse response header. If the request header contains a valid traceparent with a trace-id, and the callee does not use a different trace-id, the callee SHOULD omit the trace-id field from the traceresponse." The proposal includes similar language for proposed-parent-id.

In order to enable consistent troubleshooting, if a response includes traceresponse, we would like it to always contain the trace-id and proposed-parent-id, even if a traceparent was provided on the request.

One use case that drives this request is a customer support use case: we want to give instructions to customers to always provide the traceresponse as part of our troubleshooting process. If the traceresponse is not emitted when a traceparent with a trace-id is provided, getting the right diagnostic information from customers becomes more complex and error prone.

dyladan:
Updates to response header

Fixes #406, #407, #398, #437

- Remove `proposed-parent-id` (Discussion in #406)
	- /cc @adriancole 
- Add `child-id` field
	- Pointed out by @kanwaldeep and others that the span id of the child is required to uniquely identify a request in some support cases where multiple requests share a trace ID.
- Make all fields required (Disussion in #398, #437)
	- Header itself is still optional

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/443.html" title="Last updated on Jan 12, 2021, 9:18 PM UTC (cc567d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/443/3305d0b...dynatrace-oss-contrib:cc567d8.html" title="Last updated on Jan 12, 2021, 9:18 PM UTC (cc567d8)">Diff</a>


http_response_header_optional_fields.md
```diff
@@ -1,61 +0,0 @@
-This document describes a set of rules which may be used to determine which optional fields to include in the HTTP response header.
```
SergeyKanzhelev, line 1: was this file referenced from somewhere like index.html?

spec/21-http_response_header_format.md
```diff
@@ -46,44 +46,34 @@ Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `2
 The following `version-format` definition is used for version `00`.
 
 ``` abnf
-version-format   = [trace-id] "-" [proposed-parent-id] "-" [trace-flags]
+version-format   = trace-id "-" child-id "-" trace-flags
 trace-id         = 32HEXDIGLC  ; 16 bytes array identifier. All zeroes forbidden
-proposed-parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
+child-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
```
SergeyKanzhelev: nit
```suggestion
child-id         = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
```

spec/21-http_response_header_format.md
```diff
@@ -144,16 +131,16 @@ traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-d75597dee50b0cac-01
...
 
-**Note**: In this case, the `proposed-parent-id` was omitted from the response because, being a part of a different trace, it was not necessary for the caller.
+### Tail Sampling Load Balancer
 
-### Load Balancer
+When a service that made a negative sampling decision makes a call to another service, there may be some event during the processing of that request that causes the called service to decide to sample the request. In this case, it may return its updated sampling decision to the caller, the caller may also return the updated sampling decision to its caller, and so on. In this way, as much of a trace as possible may be recovered for debugging purposes even if the original sampling decision was negative.
```
SergeyKanzhelev, line 141: term `negative sampling decision` may not be very clear to a reader

SergeyKanzhelev, line 141: may be this paragraph would work better in request processing section which is non-formative. It sounds like an example rather than a spec. Maybe spec may only say that sampling bit returned in response header may not match the flag that was sent in the request and processing will give this example with the event that happens during the processing in the processing section.

SergeyKanzhelev (approved):
this PR is making spec better, suggestion I added is as applicable to previous iteration. So I'm approving this PR with the comment that we may need to discuss whether we need to move soem examples to the processing non-formative section

http_response_header_optional_fields.md
dyladan: no

briancr-ms (approved):

danielkhan:
I agree with Sergey. Let's wait for 7 days for comments.

spec/21-http_response_header_format.md
dyladan: Created a follow up task

dyladan: Created a follow up task

dyladan:
> I agree with Sergey. Let's wait for 7 days for comments.

Has now been 28 days :) would still appreciate reviews from some more WG members.

spec/21-http_response_header_format.md
```diff
@@ -102,39 +92,36 @@ The current version of this specification (`00`) only supports a single flag cal
...
+The `sampled` flag provides interoperability between tracing systems. It allows tracing systems to communicate recording decisions and enable a better experience for the customer. For example, when a SaaS load balancer service participates in a <a>distributed trace</a>, this service has no knowledge of the tracing vendor used by its callee. This service may produce records of incoming requests for monitoring or troubleshooting purposes. The `sampled` flag can be used to ensure that information about requests that were marked for recording by the callee will also be recorded by the SaaS load balancer service upstream so that the callee can troubleshoot the behavior of every recorded request.
```
yurishkuro:
```suggestion
The `sampled` flag provides interoperability between tracing systems. It allows tracing systems to communicate recording decisions and enable a better experience for the customer. For example, when a SaaS load balancer service participates in a <a>distributed trace</a>, this service has no knowledge of the tracing system used by its callee. This service may produce records of incoming requests for monitoring or troubleshooting purposes. The `sampled` flag can be used to ensure that information about requests that were marked for recording by the callee will also be recorded by the SaaS load balancer service upstream so that the callee can troubleshoot the behavior of every recorded request.
```

spec/21-http_response_header_format.md
```diff
@@ -102,39 +92,36 @@ The current version of this specification (`00`) only supports a single flag cal
...
 
 The `sampled` flag has no restrictions.
 
-The following are a set of suggestions that vendors SHOULD use to increase vendor interoperability.
+The following are a set of suggestions that tracing systems SHOULD use to increase vendor interoperability.
```
yurishkuro:
```suggestion
The following are a set of suggestions that tracing systems SHOULD use to increase interoperability.
```

spec/21-http_response_header_format.md
```diff
@@ -144,16 +131,16 @@ traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-d75597dee50b0cac-01
...
 ```
 
-In this example, a participant in a trace with ID `4bf92f3577b34da6a3ce929d0e0e4736` calls a third party system that collects their own internal telemetry using a new trace ID `1baad25c36c11c1e7fbd6d122bd85db6`. When the third party completes its request, it returns the new trace ID and internal sampling decision to the caller. If there is an error with the request, the caller can include the third party's internal trace ID in a support request.
+In this example, a participant in a trace with ID `4bf92f3577b34da6a3ce929d0e0e4736` calls a third party system that collects their own internal telemetry using a new trace ID `1baad25c36c11c1e7fbd6d122bd85db6`. When the third party completes its request, it returns the new trace ID, the ID of the operation, and internal sampling decision to the caller. If there is an error with the request, the caller can include the third party's internal trace ID in a support request.
 
-**Note**: In this case, the `proposed-parent-id` was omitted from the response because, being a part of a different trace, it was not necessary for the caller.
+### Tail Sampling Load Balancer
```
yurishkuro: Tail sampling has a pretty well defined meaning in the industry already, and it's not what is described here. I would rename the section, maybe "backtrace"

spec/21-http_response_header_format.md
```diff
@@ -144,16 +131,16 @@ traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-d75597dee50b0cac-01
...
 
-### Load Balancer
+When a service that made a negative sampling decision makes a call to another service, there may be some event during the processing of that request that causes the called service to decide to sample the request. In this case, it may return its updated sampling decision to the caller, the caller may also return the updated sampling decision to its caller, and so on. In this way, as much of a trace as possible may be recovered for debugging purposes even if the original sampling decision was negative.
 
-When a request passes through a load balancer, the load balancer may wish to defer a sampling decision to its called service. In this instance, the called service MAY return a `traceresponse` field indicating its sampling decision.
+One example of this might be a load balancer which samples a random subset of requests. If the load balancer origin encounters a problem, it may indicate that the request should be sampled by the load balancer anyway by returning a `traceresponse` with the sampled flag set.
```
yurishkuro: Is "origin" a standard enough term for LBs? I find it exceptionally confusing. How about "destination service"?

yurishkuro (approved):

spec/21-http_response_header_format.md
dyladan: how did you come up with the term backtrace? It isn't immediately obvious to me what the connection is. Backtrace to me would imply stitching the trace using response headers, which is not what we're doing.

spec/21-http_response_header_format.md
dyladan: I've usually heard "origin" terminology but I'm happy to change it. My experience may not reflect common practice.

spec/21-http_response_header_format.md
dyladan: What about "deferred"? We are already using it in other parts of the document.

https://www.w3.org/TR/trace-context/#x3-2-2-5-1-sampled-flag:~:text=If%20a%20component%20deferred%20or%20delayed,trace%20is%20initiated%20by%20this%20component

https://www.w3.org/TR/trace-context/#x3-2-2-5-1-sampled-flag:~:text=Deferred%20sampling

yurishkuro: yes, deferred sampling sounds good here

danielkhan (approved):

PR #455, opened 2021-06-14
Move to GitHub actions

plehegar:
We're retiring our use of travis, switching to use GitHub actions intead


.github/workflows/selt_test.yml
```diff
@@ -0,0 +1,14 @@
+jobs:
```
yurishkuro: `.github/workflows/selt_test.yml` - self, not selt?

test/start.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cd ./test
```
SergeyKanzhelev: is it started from the folder above? Maybe this:

```suggestion
cd "$(dirname "$0")"
```

.github/workflows/selt_test.yml
plehegar: oops, Fixed in https://github.com/w3c/trace-context/pull/455/commits/589c051800651927f8efdfddff48f148df899551

test/start.sh
plehegar: I removed the shell script and moved the cd command back in the yml

yurishkuro (approved):

SergeyKanzhelev (approved):

danielkhan (approved):

PR #458, opened 2021-07-14
Updated tracestate to specify that it should not be used for application level properties

### LINKED ISSUE #439: Clarify how tracestate can or cannot be used by application developers
I'm writing as the author of logging framework for .NET application developers which supports distributed logging including per-request logging, sampled logging, and pluggable correlation protocols (see https://doc.postsharp.net/6.8/distributed-logging and https://doc.postsharp.net/6.8/log-enabling#sampled).

Our logging framework allows application developers to define logging properties, and properties marked as "baggage" are transferred to remote services (in the spirit of OpenTracing and the .NET Activity class). Properties in this framework are name-value pairs where the values are arbitrary objects and eventually rendered into arbitrary strings. It's important to note that these properties are defined by _application developers_, not by infrastructure/framework vendors. A typical piece of baggage could be a user name, for instance.

My questions related to this use case are:

* Should every application property map to a `tracestate` member? Your language seems to indicate that this facility is designed for infrastructure, not for application code.
* If yes, how can a framework know if the member is in "raw" format or in Base64, so that it can deserialize it properly for use by the application developer?
* If no, is it recommendable to serialize all application properties into a single `tracestate` member?
* How are vendors supposed to react to application properties?
* Are there more recommendations for application developers and framework developers?

Thank you.

kalyanaj:
Updated tracestate section to specify that it should not be used for application level properties. Resolves #439.

***
<a href="https://pr-preview.s3.amazonaws.com/kalyanaj/trace-context/pull/458.html" title="Last updated on Aug 31, 2021, 7:23 PM UTC (f647fb3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/458/0d32b21...kalyanaj:f647fb3.html" title="Last updated on Aug 31, 2021, 7:23 PM UTC (f647fb3)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -227,6 +227,8 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
+The `tracestate` HTTP header MUST NOT be used for _application_ level properties such as userId. [W3C baggage](https://github.com/w3c/baggage) provides the specification for defining and propagating application level properties.
+
 ### Header Name
 
 Header name: `tracestate`
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -227,7 +227,7 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
-The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a tracing vendor. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
+The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a telemetry system. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
 
 ### Header Name
 
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -227,7 +227,7 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
-The `tracestate` HTTP header MUST NOT be used for _application_ level properties such as userId. [[BAGGAGE]] MAY be used for defining and propagating application level properties.
+The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a tracing vendor. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
 
 ### Header Name
 
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -227,7 +227,7 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
-The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a telemetry system. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
+The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a tracing system. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
 
 ### Header Name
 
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

Kanwaldeep (approved):

SergeyKanzhelev (changes requested):
specs links must be done differently. Also, normative section cannot refer to WD documents

SergeyKanzhelev:
> specs links must be done differently. Also, normative section cannot refer to WD documents

for example see how `[[!RFC5234]]` is referenced. @plh can help.

I think at this stage we may include this note into the informative header processing section

kalyanaj:
> > specs links must be done differently. Also, normative section cannot refer to WD documents
> 
> for example see how `[[!RFC5234]]` is referenced. @plh can help.
> 
> I think at this stage we may include this note into the informative header processing section

Good points, thanks Sergey for pointing this out. Would "http_header_format_rationale.md" (I am assuming you are referring to that above) be a good non-normative place to add this information?  I will look at that one and will move it there, and will figure out how to reference the spec link.

SergeyKanzhelev:
I meant this one: https://github.com/w3c/trace-context/blob/0d32b21bf4d5e1b8cbbbac71a7124471989443fa/index.html#L103

kalyanaj:
> I meant this one:
> 
> https://github.com/w3c/trace-context/blob/0d32b21bf4d5e1b8cbbbac71a7124471989443fa/index.html#L103

Thanks, I checked the 30-processing-model.md file, but I am not sure if this info about trace-state fits well with the flow there. I was thinking that maybe we should remove the link to the baggage spec from this for now (since it is WD), and instead just retain only the first line i.e. "The `tracestate` HTTP header MUST NOT be used for _application_ level properties such as userId." in the original file (20-http_request_header_format.md) as part of the tracestate header description.

yurishkuro:
> maybe we should remove the link to the baggage spec from this for now (since it is WD)

I wouldn't. What's the harm in referencing it? Could say "provides a proposal" instead of "provides the specification"

SergeyKanzhelev:
> > maybe we should remove the link to the baggage spec from this for now (since it is WD)
> 
> I wouldn't. What's the harm in referencing it? Could say "provides a proposal" instead of "provides the specification"

I think the problem of normative spec referencing work in progress is that work in progress may change direction (potentially). Even links to the non-specifications are not well-recieved, Ideally all "dependencies" needs to be stable so over time the meaning of a spec is not changing.

SergeyKanzhelev (changes requested):
> The `tracestate` HTTP header MUST NOT be used for _application_ level properties such as userId."

My problem is that without the link to spec, that explains more about what "application level properties" are, we need more explanaition in the spec. Why `trace-id` is not application level for instance? Is `application-id` or "cluster" an applicaton level property? I know some vendors want to propagate the "cluster" or "application-id" which help to route traces to the right trace backend "bucket"

spec/20-http_request_header_format.md
```diff
@@ -227,6 +227,8 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
+The `tracestate` HTTP header MUST NOT be used for _application_ level properties such as userId. [W3C baggage](https://github.com/w3c/baggage) provides the specification for defining and propagating application level properties.
```
plehegar:
```suggestion
The `tracestate` HTTP header MUST NOT be used for _application_ level properties such as userId. [[BAGGAGE]] MAY be used for defining and propagating application level properties.
```

plehegar: Note that the reference will work once Baggage is published (this Thursday)

spec/20-http_request_header_format.md
```diff
@@ -227,6 +227,8 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
+The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a tracing vendor. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
```
dyladan: We should avoid the term "vendor" where possible as we may be used by open source tools too. I would use the word "telemetry system" or similar

dyladan (approved):

spec/20-http_request_header_format.md
```diff
@@ -227,6 +227,8 @@ The main purpose of the `tracestate` HTTP header is to provide additional vendor
 
 If the vendor failed to parse `traceparent`, it MUST NOT attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` MUST NOT affect the parsing of `traceparent`.
 
+The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a telemetry system. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
```
SergeyKanzhelev:
```suggestion
The `tracestate` HTTP header MUST NOT be used for any properties that are not defined by a tracing system. [[BAGGAGE]] MAY be used for defining and propagating such application level properties.
```

We call it "tracing systems"

SergeyKanzhelev (approved):

PR #465, opened 2021-08-31
fix(#459): remove confusing vendor/tenant wording

### LINKED ISSUE #459: Confusion between vendor and tenant in tracestate keys
[§ 3.3.1.3.1 Key](https://www.w3.org/TR/trace-context/#key):

> Vendors SHOULD set the tenant ID at the beginning of the key. For example,
`fw529a3039@dt` - `fw529a3039` is a tenant ID and `@dt` is a vendor name.

The example key seems to have the tenant ID at the _end_, not the beginning. Or am I missing something?

dyladan:
This fixes #459 by making the prose underneath the ABNF in the key section more clear.

It is not required to fix this in `main` as the entire paragraph has been removed there.

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/465.html" title="Last updated on Sep 24, 2021, 2:53 PM UTC (eea5d04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/465/92cee7d...dynatrace-oss-contrib:eea5d04.html" title="Last updated on Sep 24, 2021, 2:53 PM UTC (eea5d04)">Diff</a>

### DIFF
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -280,8 +280,7 @@ lcalpha    = %x61-7A ; a-z
 
 **Note**: Identifiers MUST begin with a lowercase letter or a digit, and can only contain lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), and forward slashes (`/`).
 
-For multi-tenant vendor scenarios, an at sign (`@`) can be used to prefix the vendor name. Vendors SHOULD set the tenant ID at the beginning of the key. For example, \
-`fw529a3039@dt` - `fw529a3039` is a tenant ID and `@dt` is a vendor name. Searching for `@dt=` is more robust for parsing (for example, searching for all a vendor's keys).
+In a distributed application where different components may be traced by different instances of the same multi-tenant tracing system, the system SHOULD keep the system ID at the end of the key. For example, `001@xyz` where `001` is the ID of the specific tenant within the tracing system and `xyz` is the tracing system identifier. This makes for fast and robust parsing. For example, by searching for all instances of `@xyz=` a tracing system can easily find all keys made by system `xyz`.
 
 ##### Value
 
--- a/spec/20-http_header_format.md
+++ b/spec/20-http_header_format.md
@@ -270,17 +270,22 @@ A `list-member` contains a key/value pair.
 
 ##### Key
 
-The key is an identifier that describes the vendor.
+The key identifies the tracestate entry.
 
 ``` abnf
-key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
-key = ( lcalpha / DIGIT ) 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" ) "@" lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+key = simple-key / multi-tenant-key
+simple-key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+multi-tenant-key = tenant-id "@" system-id
+tenant-id = ( lcalpha / DIGIT ) 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
+system-id = lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
 lcalpha    = %x61-7A ; a-z
 ```
 
 **Note**: Identifiers MUST begin with a lowercase letter or a digit, and can only contain lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), and forward slashes (`/`).
 
-In a distributed application where different components may be traced by different instances of the same multi-tenant tracing system, the system SHOULD keep the system ID at the end of the key. For example, `001@xyz` where `001` is the ID of the specific tenant within the tracing system and `xyz` is the tracing system identifier. This makes for fast and robust parsing. For example, by searching for all instances of `@xyz=` a tracing system can easily find all keys made by system `xyz`.
+There are two different types of tracestate keys. The first type of key is a simple key used by tracing systems which do not have multiple tenants. Simple keys contain only lowercase alphanumeric characters, underscores, dashes, asterisks, and forward slashes. For example: `my-tracing-system`.
+
+The second type of key is used by multi-tenant tracing systems where each tenant requires a unique tracestate entry. Multi-tenant keys consist of a tenant ID followed by the `@` character followed by a system ID. This allows for fast and robust parsing. For example, tracing system `xyz` can easily find all of its tracestate entries by searching for all instances of `@xyz=`.
 
 ##### Value
 


spec/20-http_header_format.md
```diff
@@ -280,8 +280,7 @@ lcalpha    = %x61-7A ; a-z
 
 **Note**: Identifiers MUST begin with a lowercase letter or a digit, and can only contain lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), and forward slashes (`/`).
 
-For multi-tenant vendor scenarios, an at sign (`@`) can be used to prefix the vendor name. Vendors SHOULD set the tenant ID at the beginning of the key. For example, \
-`fw529a3039@dt` - `fw529a3039` is a tenant ID and `@dt` is a vendor name. Searching for `@dt=` is more robust for parsing (for example, searching for all a vendor's keys).
+In a distributed application where different components may be traced by different instances of the same multi-tenant tracing system, the system SHOULD keep the system ID at the end of the key. For example, `001@xyz` where `001` is the ID of the specific tenant within the tracing system and `xyz` is the tracing system identifier. This makes for fast and robust parsing. For example, by searching for all instances of `@xyz=` a tracing system can easily find all keys made by system `xyz`.
```
kalyanaj: This section currently uses two different terms: "vendor" and "tracing system" - it may be good to use one for consistency. In other words, The section starts with "The key is an identifier that describes the vendor." but then switches to "system ID" with this change. Should we start the section with something like "The key is an identifier of a tracing system."?

kalyanaj:
```suggestion
As part of the key, a multi-tenant tracing system MAY include the identifier of a tenant of the tracing system. In such a case, the key will have two components: the identifier of the tenant and the identifier of the tracing system. The tracing system SHOULD keep its identifier at the end of the key. An example of such a key is `001@xyz` where `001` is the ID of the specific tenant within the tracing system and `xyz` is the tracing system identifier. This makes for fast and robust parsing. For example, by searching for all instances of `@xyz=` a tracing system can easily find all keys made by system `xyz`.
```

kalyanaj: Should we describe more about the @ character: is that an example or are we specifically saying that tracing systems SHOULD use @ as the separator?

SergeyKanzhelev: +1 to this. We only allow one `@` sign for this reason

SergeyKanzhelev: since it's errata, may not ned to change the ABNF definition. But in main we can even call section after the `@` sign as `tracing-system` explicitly

dyladan: In main we currently describe _less_ about the `@` character and we allow more than one. For this change, I can describe more what its meaning is, but it was removed for a reason I assume.

dyladan:
@kalyanaj @SergeyKanzhelev what would you think if I updated it like this:

##### Key

The key identifies the tracestate entry.

``` abnf
key = single-tenant-key / multi-tenant-key
single-tenant-key = lcalpha 0*255( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
multi-tenant-key = tenant-id "@" system-id
tenant-id = ( lcalpha / DIGIT ) 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
system-id = lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
lcalpha    = %x61-7A ; a-z
```

**Note**: Identifiers MUST begin with a lowercase letter or a digit, and can only contain lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), and forward slashes (`/`).

There are two different types of tracestate keys. The first type of key is used by tracing systems which are single-tenant or which share the same entry between all tenants. Single-tenant keys SHOULD contain only lowercase alphanumeric characters, underscores, dashes, asterisks, and forward slashes. For example: `my-tracing-system`. The second type of key is used by multi-tenant tracing systems where each tenant requires a unique tracestate entry. Multi-tenant keys consist of a tenant ID followed by the `@` character followed by a system ID. For example: `001@xyz` and `002@xyz`. This allows for fast and robust parsing. For example, tracing system `xyz` can easily find all of its tracestate entries by searching for instances of `@xyz=`.

spec/20-http_header_format.md
SergeyKanzhelev: >  we allow more than one. 

do we? We only allow one per key, correct?

kalyanaj:
@dyladan : Apologies on the delay in getting back. Here are my thoughts/comments on the above:

1. For the system id, can we standardize on the same syntax/requirements for both cases?  i.e. if 14 characters are sufficient for the multi-tenant case, could it suffice for the single-tenant as well (that way implementations can treat it uniformly for both cases)?  If so, can we describe it as:

single-tenant-key = system-id
multi-tenant-key = tenant-id "@" system-id

2. Terminology: I am not sure if "single-tenant" captures well the case where a system chooses to use "the same entry between all tenants". So, further iterating on the previous change above, instead of specifically defining single-tenant-key and multi-tenant-key, can we simplify it to:

key = tracing-system-identifier / tenant-identifier "@" tracing-system-identifier
tenant-identifier = ( lcalpha / DIGIT ) 0*240( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
tracing-system-identifier = lcalpha 0*13( lcalpha / DIGIT / "_" / "-"/ "*" / "/" )
lcalpha    = %x61-7A ; a-z

spec/20-http_header_format.md
dyladan: No, we don't have any restrictions in `main`

dyladan:
> @dyladan : Apologies on the delay in getting back. Here are my thoughts/comments on the above:
> 
> 1. For the system id, can we standardize on the same syntax/requirements for both cases?  i.e. if 14 characters are sufficient for the multi-tenant case, could it suffice for the single-tenant as well (that way implementations can treat it uniformly for both cases)?  If so, can we describe it as:
> 
> single-tenant-key = system-id
> multi-tenant-key = tenant-id "@" system-id

Probably, but since this is a backport fix to the already released level-1 spec, we can't make any breaking changes. All of the changes i've suggested so far have just been different ways to describe the exact same syntax.

> 1. Terminology: I am not sure if "single-tenant" captures well the case where a system chooses to use "the same entry between all tenants". So, further iterating on the previous change above, instead of specifically defining single-tenant-key and multi-tenant-key, can we simplify it to:

 
> key = tracing-system-identifier / tenant-identifier "@" tracing-system-identifier
> tenant-identifier = ( lcalpha / DIGIT ) 0_240( lcalpha / DIGIT / "_" / "-"/ "_" / "/" )
> tracing-system-identifier = lcalpha 0_13( lcalpha / DIGIT / "_" / "-"/ "_" / "/" )
> lcalpha = %x61-7A ; a-z

I like the simplicity of this, but it is breaking unfortunately. The already released level 1 spec allows for tracing system identifiers that are longer than 14 characters.

kalyanaj (approved):
LGTM.

SergeyKanzhelev (approved):

Issue #467, opened 2021-09-21
Randomness flag bit


bogdandrutu:
Based on the OpenTelemetry work related to sampling traces using a consistent head sampling rate, @jmacd @bogdandrutu and @oertl would like to make a proposal:

In OpenTelemetry we need to know if (and how many) bits from the TraceId are randomly generated. In the https://github.com/open-telemetry/oteps/pull/168 we are proposing to use an "r" value that encodes how many bits are random from the trace-id, but this can be easily encoded using the trace-flags, and this information can be used by other libraries/implementations to do "logs" sampling or any other context specific samplings.

The proposal is to use 1 bit from the TraceFlags that encodes the following information:
* If "0" it means that the "randomness" of the trace-id is unknown - this is for backwards compatibility.
* If "1" it means that the right most 63 bits (this needs to be better specified) are randomly generated and can be used to calculate sampling probabilities. e.g. in case of a trace-id="1234567887654321**53ce929d0e0e4736**" from the right most 16 characters if converted to a binary: **53ce929d0e0e4736** -> "1**01001111001110100100101001110100001110000011100100011100110110**" - all bits are random except the left most bit.

**Why 63 and not 64?**

The main motivation for this is because:
* Languages like golang have a native implementation to generate 63bits random numbers (see https://pkg.go.dev/math/rand#Int63)
* Languages like Java do not have support for unsigned numbers, which makes using the 64th bit very hard when converting to a number in order to calculate probabilities.

The specification can ensure 64bit but will be overkill and unnecessary.

Related/Simplification: This will resolve/help with a large part of https://github.com/w3c/trace-context/issues/463

tedsuo:
It would be absolutely fabulous if TraceContext began supporting some amount guaranteed randomness. This is also a good time to do it, as many organizations which produce these IDs are still in the process of adopting this header.

Referencing a related issue, OTel is considering using the remaining bits as a timestamp: https://github.com/open-telemetry/opentelemetry-specification/issues/1947

But as I mentioned in that issue, what is the possibility that we will want ~128 bits of randomness in the future? I would suggest that we don't assume we can easily bump up the amount of randomness in a later version of the spec.

dyladan:
[...]

Issue #463, opened 2021-08-17
Consider trace context support for probability sampling


jmacd:
The OpenTelemetry project has been working to specify how it propagates information about probability sampling through a couple of OTEP drafts:

[OTEP 168: Specify how to propagate consistent head sampling probability](https://github.com/open-telemetry/oteps/pull/168)
[OTEP 170: Probability sampling: Sampler Name and Adjusted Count attributes](https://github.com/open-telemetry/oteps/pull/170)

The first of these discusses how to propagate head probability so that each Span recorded in a Trace Context that has the `sampled` flag set knows its "adjusted count", which is the inverse of probability. We have proposed to use power-of-two sampling rates, following [research by Otmar Ertl](https://arxiv.org/pdf/2107.07703.pdf), and have come to see the use of a dedicated `tracestate` field as potentially too costly to have on-by-default.

Using `tracestate` means passing around 30 bytes per context, and considering this overhead we would like to see a Version-1 W3C `traceparent` with the addition of a couple of bytes of information. We can do this with 6 or 7 bits of information, ideally, but it will require specifying a lot more about `traceparent` and which bits of the TraceID are truly random.

This issue is a placeholder for raising this discussion in the W3C group.

jmacd:
Since the existing parts of a `traceparent` are  base16 encoded, and whereas the version 0 `traceparent` reads like

```
traceparent: 00-TTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTT-SSSSSSSSSSSSSSSS-FF
```

for T, S and F being half bytes of the TraceID, SpanID, and Flag parts.

The variation discussed by OpenTelemetry would be:

```
traceparent: 01-RRRRRRRRRRRRRRRRTTTTTTTTTTTTTTTT-SSSSSSSSSSSSSSSS-FF-PP
```

where R represents a half byte of true random TraceID (the most significant half), and P represents new information related to the head probability. The 64 available values for PP are recognized as negative base2 logarithm of the sampling probability:

0: an adjusted count of 1 (i.e., probability == 2^0)
1: an adjusted count of 2 (i.e., probability == 2^-1)
2: an adjusted count of 4 (i.e., probability == 2^-2)
...
62: an adjusted count of 2^62
63: an adjusted count of zero

In order to recognize unknown head sampling probability, we would propose a new trace context flag to indicate two things: (a) the most significant 64 bits of the TraceID are true random, (b) the head probability is known.

[...]

PR #468, opened 2021-09-28
Add sampling constant

dyladan:
This draft proposal is an early proposal at solving #463

It does not directly compete with, but is overlapping with, #467

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/468.html" title="Last updated on Sep 28, 2021, 3:04 PM UTC (e7c7b30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/468/ce72c11...dynatrace-oss-contrib:e7c7b30.html" title="Last updated on Sep 28, 2021, 3:04 PM UTC (e7c7b30)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -7,7 +7,7 @@ This section describes the binding of the <a href="#dfn-distributed-traces">dist
 The `traceparent` request header represents the incoming request in a tracing system in a common format, understood by all vendors. Here’s an example of a `traceparent` header.
 
 ``` http
-traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
+traceparent: 01-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01-0602
 ```
 
 The `tracestate` request header includes the parent in a potentially vendor-specific format:
@@ -19,7 +19,7 @@ tracestate: congo=t61rcWkgMzE
 For example, say a client and server in a system use different tracing vendors: Congo and Rojo. A client traced in the Congo system adds the following headers to an outbound HTTP request.
 
 ``` http
-traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
+traceparent: 01-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01-0100
 tracestate: congo=t61rcWkgMzE
 ```
 
@@ -28,7 +28,7 @@ tracestate: congo=t61rcWkgMzE
 The receiving server, traced in the Rojo tracing system, carries over the `tracestate` it received and adds a new entry to the left.
 
 ``` http
-traceparent: 00-0af7651916cd43dd8448eb211c80319c-00f067aa0ba902b7-01
+traceparent: 01-0af7651916cd43dd8448eb211c80319c-00f067aa0ba902b7-01-0100
 tracestate: rojo=00f067aa0ba902b7,congo=t61rcWkgMzE
 ```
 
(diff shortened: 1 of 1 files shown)


spec/20-http_request_header_format.md
```diff
@@ -80,22 +81,25 @@ The dash (`-`) character is used as a delimiter between fields.
...
 ``` abnf
-version-format   = trace-id "-" parent-id "-" trace-flags
-trace-id         = 32HEXDIGLC  ; 16 bytes array identifier. All zeroes forbidden
-parent-id        = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
-trace-flags      = 2HEXDIGLC   ; 8 bit flags. Currently, only one bit is used. See below for details
+version-format         = trace-id "-" parent-id "-" trace-flags [ "-" sampling-constant ]
+trace-id               = 32HEXDIGLC  ; 16 bytes array identifier. All zeroes forbidden
+parent-id              = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
+trace-flags            = 2HEXDIGLC   ; 8 bit flags. Currently, only one bit is used. See below for details
+sampling-constant      = sampling-random-value parent-sampling-constant
```
yurishkuro, line 100:
```suggestion
sampling-constant      = sampling-random-value parent-sampling-rate
```

spec/20-http_request_header_format.md
```diff
@@ -80,22 +81,25 @@ The dash (`-`) character is used as a delimiter between fields.
...
+sampling-random-value  = 2HEXDIGLC   ; geometrically distributed 8-bit random value used for consistent sampling
+parent-sampling-constant = 2HEXDIGLC   ; 8-bit value used to represent the head sampling probability.
```
yurishkuro, line 102: I think the intention of limiting values to 6-bits only was that on the wire the combined constant could be represented with just 3 hexes, not 4. Was this discussed and decided against?

dyladan: I understood the original 6-bit requirement was because of the desire to use a 64-bit int, but due to some restrictions of some languages (e.g. java) it is easier to ignore the sign bit hence 63 bits of data. Then a state for 0 was needed because there is no `x` such that `2^-x = 0` leaving 62 bits. From there, only 6 bits is needed to represent the number of leading 0s in a 62-bit binary number and 62 was the highest required exponent because `2^-62` was the minimum sampling rate.

AFAIK the 6-bit + 6-bit 3-hex-character representation was just a happy coincidence of this earlier decision.

In any case, the decision to use 4 or 3 characters doesn't matter that much to me. 4 is slightly easier to parse (parse 2 hex bytes vs some bit shifting needed) and much easier for a human to read.

spec/20-http_request_header_format.md
dyladan: It is not really the rate. It is a constant used to calculate the probability. For example, a value of 2 works out to a probability of `p = 2^-x = 2^-2 = 0.25`

yurishkuro: the term "parent sampling constant" has no meaning to me. It is still a sampling rate, it's just encoded in a particular manner.

jmacd: The way we have it specified in https://github.com/open-telemetry/oteps/pull/168, this field contains the "adjusted count" specifically encoded by the logarithm for non-zero values and as 63 for the zero case.

spec/20-http_request_header_format.md
jmacd: I agree on both counts.

The use of 4 bytes is simpler to explain and implement, and 3 is definitely acceptable.
One reason to go with 4 instead of 3 is that it makes #467 orthogonal. The random value is only needed when the randomness flag is not set, according to that proposal (@bogdandrutu). Combining this proposal with #467, the `p` value is only needed _when sampled_ and the `r` value is built in--we arrive at a nice outcome:

- With randomness flag set: 2 bytes extra when sampled, 0 bytes extra when not sampled 
- Without randomness flag set: 4 bytes extra when sampled, 2 bytes extra when not sampled
-

jmacd:
[Having now specified how OpenTelemetry could use `tracestate` to convey r-values and p-values for itself](https://github.com/open-telemetry/oteps/pull/168), speaking as a member of that community, it would lead to a significantly better outcome for users and vendors if we are able to quickly and by consensus adopt a version-1 trace context format like this one combined with the proposal in #467.

Keeping in mind the original purpose of this work--which is to count spans and logs that happen in traced contexts and translate them into estimated metrics about how many associated spans and logs are happening in all contexts--we're better off with on-by-default implementations of probability sampling. The combination of these proposals leads to easier integration for span-to-metrics pipelines at a cost of 2 bytes per sampled context and 0 bytes per unsampled context.

PR #474, opened 2021-11-09
Add the random trace id flag

### LINKED ISSUE #467: Randomness flag bit

dyladan:
Resolves #467 

As discussed in the previous working group meeting, this PR adds a trace flag to signal to downstream services that part of the trace id is random. Randomness in the trace ID is desirable because it may be used for things like consistent trace sampling, sharding, and more.

/cc @bogdandrutu author of #467 
/cc @jmacd @oertl

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/474.html" title="Last updated on Nov 9, 2021, 9:08 PM UTC (f65d9a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/474/5ddf6db...dynatrace-oss-contrib:f65d9a5.html" title="Last updated on Nov 9, 2021, 9:08 PM UTC (f65d9a5)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/474.html" title="Last updated on Apr 13, 2022, 1:38 PM UTC (2276524)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/474/bfcc82c...dynatrace-oss-contrib:2276524.html" title="Last updated on Apr 13, 2022, 1:38 PM UTC (2276524)">Diff</a>


spec/60-trace-id-format.md
```diff
@@ -22,6 +22,9 @@ concerns](#privacy-considerations) of exposing unwanted information. Randomness
 also allows tracing vendors to base sampling decisions on `trace-id` field value
 and avoid propagating an additional sampling context.
 
+If the `random-trace-id` flag is set, the least significant 7 bytes of the
```
yurishkuro: The terminology "least significant" does not apply to a byte sequence. "right-most" is fine.

yurishkuro:
If this resolves #467, please state this in the description.

Does this not require a bump to the header's version number?

spec/20-http_request_header_format.md
```diff
@@ -176,6 +174,28 @@ There are two additional options that vendors MAY follow:
...
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+##### Random Trace ID Flag
+
+<!--
+
+TODO: how many random bytes are needed?
+7 was chosen as it can be efficiently represented as a 64-bit signed or unsigned integer.
+8 would require an unsigned long which is not supported by some languages (like Java).
+63 bits would be possible, but would require a more complex description that may be more difficult to understand.
```
bogdandrutu: 63 was not only because of java. Also golang has this limitation https://pkg.go.dev/math/rand#Int63

spec/20-http_request_header_format.md
```diff
@@ -176,6 +174,28 @@ There are two additional options that vendors MAY follow:
...
+
+TODO: Which specific bytes should be random?
+The least significant bytes were chosen because some tracing systems are known to use the most significant
+portion of the trace id for non-random data such as a timestamp component.
+
+TODO: Do we want to place any restrictions on the randomness or is saying "MUST be random" enough?
+As an example, RFC4122 (UUID v4) simply states "Set all the other bits to randomly (or pseudo-randomly) chosen values."
```
bogdandrutu: I am ok with same terminology as RFC4122

bogdandrutu:
> Does this not require a bump to the header's version number?

No, since this adds just a flag to the trace-flags and is backwards compatible.

spec/20-http_request_header_format.md
dyladan: It's ok. This is just a comment to help us make a decision. Once we make a decision we can document the reasons.

dyladan:
> > Does this not require a bump to the header's version number?
> 
> No, since this adds just a flag to the trace-flags and is backwards compatible.

I agree with Bogdan here.

spec/60-trace-id-format.md
yurishkuro:
```suggestion
If the `random-trace-id` flag is set, the right-most 7 bytes of the
```

yurishkuro (approved):

spec/20-http_request_header_format.md
```diff
@@ -176,6 +174,28 @@ There are two additional options that vendors MAY follow:
...
+The right-most bytes were chosen because some tracing systems are known to use the left-most
+portion of the trace id for non-random data such as a timestamp component.
+
+TODO: Do we want to place any restrictions on the randomness or is saying "MUST be random" enough?
+As an example, RFC4122 (UUID v4) simply states "Set all the other bits to randomly (or pseudo-randomly) chosen values."
+https://datatracker.ietf.org/doc/html/rfc4122#section-4.4
+
+-->
+
+When set, the second least significant bit (second from the right), denotes that the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
```
basti1302: *nitpick:* I know this is the same phrasing that we use for `sampled`, but when actually reading it, I find the phrasing a bit awkward. 

How about something like this: 

> The second least significant bit of the `trace-flags` field denotes the `random-trace-id` flag. If that flag is set, the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).

Additionally, this section currently does not introduce the spelling `random-trace-id` (lower case, kebap case) but we refer to it by that spelling later on. Thus having that exact spelling here in the definition might be good, I think.

basti1302 (approved):
LGTM, I left one nitpick/phrasing suggestion. 

(And I assume the two open TODOs need to be decided and removed before merging).

yurishkuro:
Note that while this resolves #467, it does not address #463 because it makes the operator choose between using externally generated ids (sometimes needed for legacy/compatibility reasons) and having randomness in the context.

dyladan:
It addresses part of #463 which is that the trace id can be used to ensure sampling is consistent between services. It does not address the part about propagating the probability itself which is intentional.

yurishkuro:
> It addresses part of #463 which is that the trace id can be used to ensure sampling is consistent between services.

It "addresses #463" only by introducing another limitation.

dyladan:
I'm afraid we might be talking past each other. I'm not entirely sure I understand what you're arguing.

yurishkuro:
Sorry, let me try again. #463 presents a high level problem - we want to have access to a single random number across all nodes in the trace. This PR says you can have it but ONLY if your trace-id is random(-ish). So this PR only solves the high level problem for a subset of systems that can also meet the additional restriction of random IDs. In contrast, the proposal in #463 solves the problem for everyone.

dyladan:
> Sorry, let me try again. #463 presents a high level problem - we want to have access to a single random number across all nodes in the trace. This PR says you can have it but ONLY if your trace-id is random(-ish). So this PR only solves the high level problem for a subset of systems that can also meet the additional restriction of random IDs. In contrast, the proposal in #463 solves the problem for everyone.

OK I see what you're saying, but in order to implement #463 you would still need a random number at the head of the trace. If the restriction is because of the random requirement, why does it not affect the other proposal? Sorry if this is obvious I just want to make sure I fully understand.

yurishkuro:
I don't think the need to generate a random number is a blocker, having a random trace-id is. There are situations where traces must reuse a 3rd-party identifier as trace-id on which no randomness guarantees can be made (same reason why traceparent spec does not require random trace-id, we only use SHOULD).

dyladan:
Thanks for the explanation. Would you rather have the proposal in #463 (or something like it)? We discussed it at a meeting and even started making a draft prototype #468 but some people thought it was too breaking of a change.

edit: it's obviously possible to have both, but with #463/#468 there is less need for this

yurishkuro:
I am not opposed to the new bit proposal, I like it, we just need to be aware that it does not solve the problem for everyone.

dyladan:
I have added the `SHOULD` wording and some wording about why users should be motivated to use the flag. Please take a look and review when you can.

spec/20-http_request_header_format.md
```diff
@@ -138,8 +142,6 @@ boolean sampled = (traceFlags & FLAG_SAMPLED) == FLAG_SAMPLED;
 
 ##### Sampled flag
 
-The current version of this specification (`00`) only supports a single flag called `sampled`.
```
kalyanaj, line 141: Under the "trace-flags" section, it might be useful for the readers if we add a line that says:

The current version of this specification (`00`) supports only two flags: `sampled` and `random-trace-id`.

spec/20-http_request_header_format.md
```diff
@@ -176,6 +178,15 @@ There are two additional options that vendors MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+##### Random Trace ID Flag
+
+When set, the second least significant bit (second from the right), denotes that the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
```
kalyanaj: Do we want to add a qualifier "at least" here?  i.e. ".....denotes that at least the right-most 7 bytes ..." instead of ".....denotes that the right-most 7 bytes ...".

spec/20-http_request_header_format.md
kalyanaj: Agree that it will be good to introduce the "random-trace-id" term/spelling here in this section since we use it in the trace-id-format text.

spec/60-trace-id-format.md
```diff
@@ -22,6 +22,9 @@ concerns](#privacy-considerations) of exposing unwanted information. Randomness
 also allows tracing vendors to base sampling decisions on `trace-id` field value
 and avoid propagating an additional sampling context.
 
+If the `random-trace-id` flag is set, the right-most 7 bytes of the
```
kalyanaj: Similar to my other comment, do we want to add the "at least" qualifier here as well?

spec/20-http_request_header_format.md
```diff
@@ -176,6 +180,17 @@ There are two additional options that vendors MAY follow:
...
 
+##### Random Trace ID Flag
+
+The second least significant bit of the trace-flags field denotes the random-trace-id flag.
+If that flag is set, the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
+If the flag is not set, the trace ID MAY still be randomly (or pseudo-randomly) generated.
+When unset, the trace ID may be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
+
+When at least the right-most 7 bytes of the `trace-id` are randomly (or pseudo-randomly) generated, the random trace ID flag SHOULD be set to `1`.
+This allows downstream consumers know that the right-most 7 bytes of the trace ID are sufficiently random, and enables them to implement features such as trace sampling or database sharding based on these bytes.
```
yurishkuro: > know that the right-most 7 bytes of the trace ID are sufficiently random

redundant

yurishkuro:
```suggestion
This allows downstream consumers to implement features such as trace sampling or database sharding based on these bytes.
```

spec/20-http_request_header_format.md
```diff
@@ -116,6 +120,8 @@ Vendors MUST ignore the `traceparent` when the `parent-id` is invalid (for examp
 
 #### trace-flags
 
+The current version of this specification (`00`) supports only two flags: `sampled` and `random-trace-id`.
```
yurishkuro, line 123: Do we not need a new version if we're adding a new flag?

spec/20-http_request_header_format.md
```diff
@@ -176,6 +180,17 @@ There are two additional options that vendors MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+##### Random Trace ID Flag
+
+The second least significant bit of the trace-flags field denotes the random-trace-id flag.
+If that flag is set, the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
```
```suggestion
If that flag is set, at least the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
```

spec/20-http_request_header_format.md
dyladan: It is backwards compatible so it is not required. We can decide to bump the version if we feel it is different enough.

kalyanaj (approved):

spec/20-http_request_header_format.md
```diff
@@ -100,13 +100,17 @@ trace-flags      = 2HEXDIGLC   ; 8 bit flags. Currently, only one bit is used. S
...
 #### trace-id
 
-This is the ID of the whole trace forest and is used to uniquely identify a <a href="#dfn-distributed-traces">distributed trace</a> through a system. It is represented as a 16-byte array, for example, `4bf92f3577b34da6a3ce929d0e0e4736`. All bytes as zero (`00000000000000000000000000000000`) is considered an invalid value.
+This is the ID of the whole trace forest and is used to uniquely identify a <a href="#dfn-distributed-traces">distributed trace</a> through a system.
+It is represented as a 16-byte array, for example, `4bf92f3577b34da6a3ce929d0e0e4736`.
+All bytes as zero (`00000000000000000000000000000000`) is considered an invalid value.
 
-If the `trace-id` value is invalid (for example if it contains non-allowed characters or all zeros), vendors MUST ignore the `traceparent`.
+The value of `trace-id` SHOULD be globally unique.
+It is RECOMMENDED that implementers randomly (or pseudo-randomly) generate the `trace-id` field, and set the corresponding [random trace id flag](#random-trace-id-flag).
```
SergeyKanzhelev: it is not a concern for this PR. The conformance section that is automatically generated doesn't mention the "RECOMMENDED" keyword, even though it exists in the spec the section refers to.

image

I wonder if this can be fixed, @plehegar do you know?

If this cannot be fixed we can switch to SHOULD as they are interchangeable with RECOMMENDED as per my understanding

spec/20-http_request_header_format.md
```diff
@@ -176,6 +180,17 @@ There are two additional options that vendors MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+##### Random Trace ID Flag
+
+The second least significant bit of the trace-flags field denotes the random-trace-id flag.
+If that flag is set, at least the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
+If the flag is not set, the trace ID MAY still be randomly (or pseudo-randomly) generated.
+When unset, the trace ID may be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
```
SergeyKanzhelev:
```suggestion
When unset, the trace ID MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
```

spec/20-http_request_header_format.md
```diff
@@ -116,6 +120,8 @@ Vendors MUST ignore the `traceparent` when the `parent-id` is invalid (for examp
 
 #### trace-flags
 
+The current version of this specification (`00`) supports only two flags: `sampled` and `random-trace-id`.
+
```
SergeyKanzhelev, line 124: cannot make comment below, on line 135 we can change abstract example to a specific example with the random-trace-id flag. Saying that 0 and 2 represent not sampled, 1 and 3 - sampled

SergeyKanzhelev (approved):
a few suggestions added

spec/20-http_request_header_format.md
dyladan: Changed to use SHOULD

SergeyKanzhelev (approved):

kalyanaj (approved):

PR #475, opened 2021-11-11
Specify expected behaviour when no traceparent is received

### LINKED ISSUE #442: Expected behavior with regard to starting a trace should be spelled out explicitly
A bit of a follow up from #390 and #433 and discussions around that in the last workshop and meeting.

The spec does not explicitly spell out the expected behavior for tracers when no traceparent header is incoming. Should a tracer start a trace by creating a trace ID and send `traceparent` downstream? 

IMO, the current wording in the non-normative section kind of implies that tracers are actually expected to start a trace if no traceparent is incoming. But apparently the current consensus is that "a tracer SHOULD start a trace if no traceparent is incoming" is already too strong here. In which case we should make that consensus explicit, too. 

Maybe we could add something to the "If no traceparent header is received:" case.

basti1302:
Resolves #442.

As discussed in the previous working group meeting, this PR adds a clarification around the expected behaviour when no traceparent headers is received.

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/475.html" title="Last updated on Dec 8, 2021, 8:19 AM UTC (3e68405)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/475/3081b03...instana:3e68405.html" title="Last updated on Dec 8, 2021, 8:19 AM UTC (3e68405)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -334,6 +334,8 @@ The version of `tracestate` is defined by the version prefix of `traceparent` he
 
 ## Mutating the traceparent Field
 
+A vendor receiving a request without a `traceparent` header SHOULD generate `traceparent` headers for outbound requests, effectively starting a new trace. A possible reason to not add `traceparent` to outbound requests is when the vendor decides to not sample this request.
+
 A vendor receiving a `traceparent` request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing it to outgoing requests.
 
 If the value of the `traceparent` field wasn't changed before propagation, `tracestate` MUST NOT be modified as well. Unmodified header propagation is typically implemented in pass-through services like proxies. This behavior may also be implemented in a service which currently does not collect distributed tracing information.
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -334,7 +334,7 @@ The version of `tracestate` is defined by the version prefix of `traceparent` he
 
 ## Mutating the traceparent Field
 
-A vendor receiving a request without a `traceparent` header SHOULD generate `traceparent` headers for outbound requests, effectively starting a new trace. A possible reason to not add `traceparent` to outbound requests is when the vendor decides to not sample this request.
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

kalyanaj (approved):

dyladan (commented):
To me the wording seems a little confusing. "SHOULD send the `traceparent` header" implies there is already a header to send, which there is not in this case. I would say something along the lines of "SHOULD generate traceparent headers for outgoing requests"

If the vendor does not sample the request, they should still send a traceparent header in most cases. If they don't the sampling decision they made is not propagated and the next component in the trace might make a different decision.

basti1302:
> I would say something along the lines of "SHOULD generate traceparent headers for outgoing requests"

I agree that this is better, will change accordingly.

basti1302:
Discussed in the meeting: Make example for not sending the header to include performance considerations because normally even when not sampling you should send the header.

basti1302:
@dyladan I made the wording around when you might not want to generate `traceparent` more precise as discussed.

spec/20-http_request_header_format.md
```diff
@@ -334,6 +334,8 @@ The version of `tracestate` is defined by the version prefix of `traceparent` he
 
 ## Mutating the traceparent Field
 
+A vendor receiving a request without a `traceparent` header SHOULD generate `traceparent` headers for outbound requests, effectively starting a new trace. A possible reason to not add `traceparent` to outbound requests is for performance reasons when the vendor decides to not sample this request. (Note that for most scenarios, vendors are expected to generate the header even when not sampling, to propagate the sampling decision downstream.)
```
```suggestion
A vendor receiving a request without a `traceparent` header SHOULD generate `traceparent` headers for outbound requests, effectively starting a new trace. A possible reason for not doing this could be a performance sensitive scenario when the vendor decides to not sample a request. Note that for most scenarios, vendors are expected to generate the header even when not sampling, to propagate the sampling decision downstream.
```

kalyanaj (approved):

bogdandrutu (approved):

dyladan (approved):

PR #477, opened 2021-11-11
Specify uniqueness of tracestate keys as the sender's responsibility

### LINKED ISSUE #369: Clarify handling of duplicate keys in tracestate
Discussing on gitter last week it was not clear from the spec how to handle duplicate keys in a received tracestate.

The tests require the tracestate is dropped in such a case: https://github.com/w3c/trace-context/blob/98f210efd89c63593dce90e2bae0a1bdcb986f51/test/test.py#L536

The example in the spec about going through a system twice is also a bit confusing, https://w3c.github.io/trace-context/#combined-header-value

I think it would make more sense, even though it is an example of what shouldn't be used, to say it shouldn't be `congo=congosSecondPosition,rojo=rojosFirstPosition,congo=congosFirstPosition` simply because the newer entries should come first.

### LINKED ISSUE #446: Clarify tracestate keys MUST be unique
Currently, uniqueness of tracestate keys is not defined anywhere. The closest it gets is the section on updating a key. However, there's nothing that says a key can't be added twice, at least not that I can find.

Code written poorly could easily merge together 2 tracestate headers and result in duplicate values. Consider messaging where you have a "current trace" and you are reading incoming state also.

This should be addressed with a MUST somewhere, ex MUST not duplicate keys or MUST ignore duplicate keys, reading the first.

basti1302:
This ~partially~ resolves #446.

This adds a new rule to "Mutating the tracestate Field" that makes it
clear that a sender must not create duplicated tracestate keys. It is a
non-breaking change because the section "Combined Header Value" already
had called that out with "Only one entry per key is allowed". Basically
this adds that rule to the section it actually belongs into.

Having that as the responsibility of the sender seems to be consensus in #446.

~There seems to be no consensus yet whether an implementation should
clean up a tracestate value it receives by discarding duplicate keys
that it does not own. Thus, that aspect is not considered in this
change.~

Update: Consensus has been reached that an implementation MAY discard duplicate key but is not obliged to. This aspect has been added to this change as well now.

Finally, this change fixes the test suite that was too strict and demanded that the whole tracestate should be dropped when duplicated keys are present. Thus, this PR fixes #369, too.

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/477.html" title="Last updated on Dec 14, 2021, 10:48 AM UTC (3590d0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/477/3081b03...instana:3590d0a.html" title="Last updated on Dec 14, 2021, 10:48 AM UTC (3590d0a)">Diff</a>


spec/20-http_request_header_format.md
```diff
@@ -297,7 +297,7 @@ The `tracestate` value is the concatenation of trace graph key/value pairs
 
 Example: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`
 
-Only one entry per key is allowed because the entry represents that last position in the trace. Hence vendors must overwrite their entry upon reentry to their tracing system.
```
basti1302, line 300: "because the entry represents that last position in the trace." had been called out [here](https://github.com/w3c/trace-context/issues/446#issuecomment-759073256) by @yurishkuro as not making a lot of sense, and I agree. Hence, I removed that part.

Also, I think update could be a better phrase instead of overwrite because we also use that in the section "Mutating the traceparent Field". 

Finally "upon reentry to their tracing system" is really unwieldy, IMO.

spec/20-http_request_header_format.md
```diff
@@ -353,6 +355,6 @@ Vendors receiving a `tracestate` request header MUST send it to outgoing request
 
 Following are allowed mutations:
 
-- **Add a new key/value pair**. The new key/value pair SHOULD be added to the beginning of the list.
+- **Add a new key/value pair**. The new key/value pair SHOULD be added to the beginning of the list. Adding a key/value pair MUST NOT result in the same key being present multiple times.
```
basti1302, line 356: This had been called out in "Combined Header Value" already, but it actually belong here, IMO.

spec/20-http_request_header_format.md
```diff
@@ -297,7 +297,7 @@ The `tracestate` value is the concatenation of trace graph key/value pairs
 
 Example: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`
 
-Only one entry per key is allowed because the entry represents that last position in the trace. Hence vendors must overwrite their entry upon reentry to their tracing system.
+Only one entry per key is allowed. Hence vendors MUST update their entry instead of adding it twice when propagating a modfied `tracestate` header downstream.
```
yurishkuro: * the word "hence" not needed
* instead of speaking of updating/adding, I would rather say vendors MUST ensure that the entry with their vendor key is only present once in the header sent downstream.

SergeyKanzhelev: +1 to just say that vendor is responsible for not duplicating keys it owns

spec/20-http_request_header_format.md
SergeyKanzhelev: I like how this formulated. It eliminates concern that if platform doesn't deal with tracestate keys, just propagates it as is - it is not responsible for the duplicates. We may want to spell it out explicitly

spec/20-http_request_header_format.md
basti1302: Well, that is exactly what I have added down in the "Mutating the traceparent Field" section -> https://github.com/w3c/trace-context/pull/477/files#diff-3832602716774745d030f22592f4c5b7d8f316e4d44be0165282244f9d6f14d9R358

Do we just want to duplicate that phrasing here? Seems redundant. How about doing away with the second sentence entirely? So just "Only one entry per key is allowed." and that's it. We also have the cross-reference 

> See [Mutating the tracestate Field](#mutating-the-tracestate-field) for details. 

so I think spelling out specifics here too is not required, the actual normative rules belong in the "Mutating" section and are already phrased as you requested.

WDYT?

spec/20-http_request_header_format.md
basti1302: See this part of my PR comment:

> There seems to be no consensus yet whether an implementation should
> clean up a tracestate value it receives by discarding duplicate keys
> that it does not own. Thus, that aspect is not considered in this
> change.

And my message here https://github.com/w3c/trace-context/issues/446#issuecomment-966451599

TL;DR: I agree with you that this is how it should work (vendors should not need to care about foreign keys at all) but I am not sure if this is consensus, so for now I was reluctant to add this aspect to this PR.

spec/20-http_request_header_format.md
SergeyKanzhelev: Yes, perhaps just delete the entire phrase including "Only one entry per key is allowed."

dyladan: I would keep the first sentence.

spec/20-http_request_header_format.md
dyladan: I agree that a tracing system should be responsible for deduping their own keys and shouldn't need to analyze or dedupe other keys.

spec/20-http_request_header_format.md
basti1302: I kept the first sentence, it leads nicely into the following example, but I removed the duplication about not adding your own header twice.

spec/20-http_request_header_format.md
basti1302: I've added that aspect as well now: https://github.com/w3c/trace-context/pull/477/commits/ccbabd6087a3ee9e65889fa54a23f32240436b06

spec/20-http_request_header_format.md
```diff
@@ -353,6 +353,6 @@ Vendors receiving a `tracestate` request header MUST send it to outgoing request
...
 - **Update an existing value**. The value for any given key can be updated. Modified keys SHOULD be moved to the beginning (left) of the list.
-- **Delete a key/value pair**. Any key/value pair MAY be deleted. Vendors SHOULD NOT delete keys that were not generated by them. The deletion of an unknown key/value pair will break correlation in other systems. This mutation enables two scenarios. The first is that proxies can block certain `tracestate` keys for privacy and security concerns. The second scenario is a truncation of long `tracestate`s.
+- **Delete a key/value pair**. Any key/value pair MAY be deleted. Vendors SHOULD NOT delete keys that were not generated by them. The deletion of an unknown key/value pair will break correlation in other systems. This mutation enables three scenarios. The first is that proxies can block certain `tracestate` keys for privacy and security concerns. The second scenario is a truncation of long `tracestate`s. Finally, vendors MAY also discard duplicate keys that were not generated by them, however, they are not obliged to do that.
```
```suggestion
- **Delete a key/value pair**. Any key/value pair MAY be deleted. Vendors SHOULD NOT delete keys that were not generated by them. The deletion of an unknown key/value pair will break correlation in other systems. This mutation enables three scenarios. The first is that proxies can block certain `tracestate` keys for privacy and security concerns. The second scenario is a truncation of long `tracestate`s. The third scenario is vendors MAY discard duplicate keys that were not generated by them.
```

basti1302: "The third scenario is vendors MAY discard" sounds a bit awkward IMO. I would prefer the current phrasing, to be honest.

kalyanaj: The current phrasing of "Finally, vendors..." is fine with me. However, the phrase ", however, they are not obliged to do that" feels redundant since it is conveyed by the MAY semantics in the previous sentence, so it feels we can remove that part to make it tighter.

basti1302: Good call, I have removed that part.

kalyanaj (approved):

basti1302:
I have added another commit that fixes the broken test case that originally sparked this discussion in #369.

basti1302:
@dyladan @yurishkuro @SergeyKanzhelev I think all comments have been addressed by now and the PR also includes a fix for the test case related to duplicate keys in the tracestate header. Could you please re-review?

spec/20-http_request_header_format.md
```diff
@@ -293,19 +293,19 @@ chr      = %x20 / nblk-chr
...
 
-The `tracestate` value is the concatenation of trace graph key/value pairs
+The `tracestate` value is the concatenation of trace graph key/value pairs.
 
 Example: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`
 
-Only one entry per key is allowed because the entry represents that last position in the trace. Hence vendors must overwrite their entry upon reentry to their tracing system.
-
-For example, if a vendor name is Congo and a trace started in their system and then went through a system named Rojo and later returned to Congo, the `tracestate` value would not be:
+Only one entry per key is allowed. For example, if a vendor name is Congo and a trace started in their system and then went through a system named Rojo and later returned to Congo, the `tracestate` value would not be:
```
yurishkuro, line 300: "is allowed" is not a normative language. I would change it to "There SHOULD be only one entry per key".

yurishkuro (approved):

spec/20-http_request_header_format.md
basti1302: The normative language around this is in the section "Mutating the tracestate Field":
> Adding a key/value pair MUST NOT result in the same key being present multiple times.

I would restrict the normative description of that aspect to the "Mutating" section because it describes _what an implementation is allowed to do_ with the header and what not. I do not see any added benefit by duplicating this aspect in normative language here. "Only one entry per key is allowed." is given in the context of the example (see the line above) and in my understanding, examples are never normative.

bogdandrutu (approved):

dyladan (approved):

Issue #481, opened 2022-01-04
Incorrect randomness claim in 50-privacy.md


dyladan:
`Privacy of traceparent field` section incorrectly states that `trace-id` is randomly generated. Section should be modified such that PII in trace ids is discouraged or forbidden, and randomness suggested as a possible solution.

kalyanaj:
Leaving this issue open to decide if we can backport this (https://github.com/w3c/trace-context/pull/482) to level-1 of the spec. 

I feel the wording in #482 ("“MUST NOT contain PII”) is sufficiently backwards compatible to backport, since the Level 1 spec *indirectly* talks about not exposing user-identifiable information.

@plehegar , based on a discussion on this in a recent working group meeting, we wanted to get your thoughts on this.

kalyanaj:
Assigned to Philippe per our discussion in the DT working group meeting today, thanks Philippe!

dyladan:
This is fixed in #495 but I will leave this open until it is published

kalyanaj:
Closing this one as we have a separate issue to track republishing of level 1 spec to include this fix.

PR #482, opened 2022-01-04
Remove incorrect randomness claim from privacy considerations

dyladan:
Related to #481 

Fix the contradiction in `main`, this fix will need to be backported to `level-1` before #481 is closed.

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/482.html" title="Last updated on Jan 5, 2022, 4:12 PM UTC (9c76e0d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/482/3081b03...dynatrace-oss-contrib:9c76e0d.html" title="Last updated on Jan 5, 2022, 4:12 PM UTC (9c76e0d)">Diff</a>

### DIFF
--- a/spec/50-privacy.md
+++ b/spec/50-privacy.md
@@ -6,7 +6,7 @@ Vendors MUST assess the risk of header abuse. This section provides some conside
 
 ## Privacy of traceparent field
 
-The `traceparent` field MUST NOT contain any personally identifiable information. One way to achieve this is to randomly generate all trace IDs. If a random number generator leverages any user identifiable information like IP address as seed state, this information may be exposed. Random number generators MUST NOT rely on any information that can potentially be user-identifiable.
+The `traceparent` field MUST NOT contain any personally identifiable information. One way to achieve this is to randomly generate all trace IDs using a random number generator that does not expose any personally identifiable information. Any random number generator used for generating trace IDs MUST NOT rely on any information as input or seed state that can potentially be personally identifiable.
 
 Another privacy risk of the `traceparent` field is the ability to correlate requests made as part of a single transaction. A downstream service may track and correlate two or more requests made in a single transaction and may make assumptions about the identity of the caller of a request based on information from another request.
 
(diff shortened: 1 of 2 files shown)


bogdandrutu (approved):

yurishkuro (approved):

SergeyKanzhelev (approved):

spec/50-privacy.md
```diff
@@ -6,7 +6,7 @@ Vendors MUST assess the risk of header abuse. This section provides some conside
 
 ## Privacy of traceparent field
 
-The `traceparent` field is comprised of randomly-generated numbers. If a random number generator leverages any user identifiable information like IP address as seed state, this information may be exposed. Random number generators MUST NOT rely on any information that can potentially be user-identifiable.
+The `traceparent` field MUST NOT contain any personally identifiable information. One way to achieve this is to randomly generate all trace IDs. If a random number generator leverages any user identifiable information like IP address as seed state, this information may be exposed. Random number generators MUST NOT rely on any information that can potentially be user-identifiable.
```
kalyanaj:
```suggestion
The `traceparent` field MUST NOT contain any personally identifiable information. One way to achieve this is to randomly generate all trace IDs using a random number generator that doesn't expose any personally identifiable information. For instance, if a random number generator uses IP address as seed state, it may expose this information. Hence, any random number generator used for generating trace IDs MUST NOT rely on any information that can potentially be personally identifiable.
```

dyladan: The use of the term "For instance" implies that you are showing an example of something, but the following sentence is not an example of the previous sentence. Reading the sentence again, I also think we don't need to provide IP as an example, because we don't want to define what might be personally identifiable. How about this?

```suggestion
The `traceparent` field MUST NOT contain any personally identifiable information. One way to achieve this is to randomly generate all trace IDs using a random number generator that does not expose any personally identifiable information. Any random number generator used for generating trace IDs MUST NOT rely on any information as input or seed state that can potentially be personally identifiable.
```

kalyanaj: Thanks, this text looks good to me.

kalyanaj (approved):

plehegar:
those changes are editorial and can be applied to Trace Context 1.

PR #500, opened 2022-10-11
Add test case for random flag propagation

dyladan:
This adds a test case which asserts that the random flag is propagated. It only runs if the newly added SPEC_LEVEL variable is 2 or greater (default 2).

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -79,12 +79,30 @@ Content-Type: application/json
 		HARNESS_BIND_HOST  the host/address which the test harness binds to (default to HARNESS_HOST)
 		HARNESS_BIND_PORT  the port which the test harness binds to (default to HARNESS_PORT)
 		SERVICE_ENDPOINT   your test service endpoint (no default value)
+		STRICT_LEVEL       the level of test strictness (default 2)
+		SPEC_LEVEL         the minimum version of the Trace Context specification being tested (default 2)
 
 	Example:
+		# Run all tests
 		python test.py http://127.0.0.1:5000/test
+		# Run one test
 		python test.py http://127.0.0.1:5000/test TraceContextTest.test_both_traceparent_and_tracestate_missing
+		# Run one test suite
 		python test.py http://127.0.0.1:5000/test AdvancedTest
+		# Run two test suites
+		python test.py http://127.0.0.1:5000/test AdvancedTest TraceContext2Test
+		# Combinations of the above
 		python test.py http://127.0.0.1:5000/test AdvancedTest TraceContextTest.test_both_traceparent_and_tracestate_missing
+		python test.py http://127.0.0.1:5000/test AdvancedTest TraceContextTest.test_both_traceparent_and_tracestate_missing TraceContext2Test.test_propagates_random_flag
+		python test.py http://127.0.0.1:5000/test AdvancedTest TraceContext2Test TraceContextTest.test_both_traceparent_and_tracestate_missing
+
+	Available Test Suites:
+		TraceContextTest
+			Trace Context Level 1 support
+		AdvancedTest
+			Advanced Trace Context Level 1 support
+		TraceContext2Test
+			Trace Context Level 2 support
 	```
 * Alternatively, you can use the Python [unit testing framework](https://docs.python.org/3/library/unittest.html) module to run the test.
 	```
--- a/test/test.py
+++ b/test/test.py
@@ -13,13 +13,16 @@ server = None
 def environ(name, default = None):
 	if not name in os.environ:
 		if default:
-			os.environ[name] = default
+			return default
 		else:
 			raise EnvironmentError('environment variable {} is not defined'.format(name))
 	return os.environ[name]
 
-STRICT_LEVEL = int(environ('STRICT_LEVEL', '2'))
-SPEC_LEVEL = int(environ('SPEC_LEVEL', '2'))
+DEFAULT_SPEC_LEVEL = '1'
+DEFAULT_STRICT_LEVEL = '2'
+
+STRICT_LEVEL = int(environ('STRICT_LEVEL', DEFAULT_STRICT_LEVEL))
+SPEC_LEVEL = int(environ('SPEC_LEVEL', DEFAULT_SPEC_LEVEL))
 print('STRICT_LEVEL: {}'.format(STRICT_LEVEL))
 print('SPEC_LEVEL:   {}'.format(SPEC_LEVEL))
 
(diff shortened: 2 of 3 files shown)


test/test.py
```diff
@@ -876,6 +878,20 @@ def test_multiple_requests_with_illegal_traceparent(self):
...
+	def test_propagates_random_flag(self):
+		'''
+		harness sends a request with a traceparent with the random flag set
+		expects a valid traceparent from the output header, with the random flag set
+		'''
+		traceparent, tracestate = self.make_single_request_and_get_tracecontext([
+			['traceparent', '00-12345678901234567890123456789012-1234567890123456-02'],
+		])
+		self.assertEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')
+		self.assertNotEqual(traceparent.parent_id.hex(), '1234567890123456')
```
kalyanaj, line 904: Is the check for parent ID meant to be an "assertEqual" check instead of "assertNotEqual"?  Looking at the parent ID being compared here, it seems to match the parentID specified in line 889.

dyladan: The request from the service being tested should have a new parent ID, not reuse the incoming one

dyladan: Trace ID should be equal and Trace Parent should be not equal. This test is correct. See a similar example here https://github.com/w3c/trace-context/blob/84b583d86ecb7005a9eab8fed86ab7117b050b48/test/test.py#L123-L132

kalyanaj: Got it, thanks for the clarification.

kalyanaj (approved):

test/test.py
```diff
@@ -876,6 +878,20 @@ def test_multiple_requests_with_illegal_traceparent(self):
...
+		self.assertTrue(traceparent.trace_flags & (1 << 1) == (1 << 1))
```
basti1302: _nitpick:_ slightly hard to read IMO. Maybe like this?
 
```suggestion
		self.assertEqual(traceparent.trace_flags & (1 << 1), 1 << 1)
```

Or this:

```suggestion
		self.assertEqual(traceparent.trace_flags & 2, 2)
```

basti1302 (approved):

test/test.py
dyladan: What about this:

```patch
diff --git a/test/test.py b/test/test.py
index 97d9991..1f8d93f 100644
--- a/test/test.py
+++ b/test/test.py
@@ -114,6 +114,12 @@ class TestBase(unittest.TestCase):
 		headers = self.make_request(headers)[0]['headers']
 		return (self.get_traceparent(headers), self.get_tracestate(headers))
 
+	def assert_bit_set(self, value, n):
+		'''
+		assert that the nth position bit is set where n == 0 is the least significant bit
+		'''
+		self.assertEqual(value & (1 << n - 1), (1 << n - 1))
+
 class TraceContextTest(TestBase):
 	def test_both_traceparent_and_tracestate_missing(self):
 		'''
@@ -890,7 +896,7 @@ class TraceContext2Test(TestBase):
 		])
 		self.assertEqual(traceparent.trace_id.hex(), '12345678901234567890123456789012')
 		self.assertNotEqual(traceparent.parent_id.hex(), '1234567890123456')
-		self.assertTrue(traceparent.trace_flags & (1 << 1) == (1 << 1))
+		self.assert_bit_set(traceparent.trace_flags, 2)
 
 if __name__ == '__main__':
 	if len(sys.argv) >= 2:
```

basti1302: That's splendid 👍

test/test.py
```diff
@@ -19,7 +19,9 @@ def environ(name, default = None):
 	return os.environ[name]
 
 STRICT_LEVEL = int(environ('STRICT_LEVEL', '2'))
+SPEC_LEVEL = int(environ('SPEC_LEVEL', '2'))
```
dyladan: @kalyanaj @basti1302 

Should we test level 2 by default or leave it to testers to manually bump to level 2 testing? My fear is that automated test suites for things like OTel which pull the test suite from git might fail for existing level 1 implementations.

basti1302: Good question. I don't think there is an ideal choice here.

How about this:
* We default to level 1 for now, implementors have to explicitly set the spec level to 2 if they already want to run level 2 tests.
* When level 2 has reached the candidate recommendation stage we set the default level to 2.
    * Yes, this will break the existing test suites of level 1 implementations (assuming they pull the test suite from git via the main branch instead of pinning a specific commit). I think at that point in time, this is actually a good signal for those folks to make them aware that there _is_ a level 2 of the spec.
    * The fix is easy: Add the env var SPEC_LEVEL=1 until you update the implementation to level 2 compliance. We need to point out that env var  very clearly but I think the readme and the help output of the test suite already do a pretty good job. I we want to go the extra mile we could log something at the start of the the test suite when the env var is absent. Something like:
    
```
if SPEC_LEVEL is unset:
    print('Warning: The environment variable SPEC_LEVEL has not been set. Defaulting to specification level {}. Consider setting SPEC_LEVEL explicitly if your implementation is based on a different specification level.'.format(DEFAULT_SPEC_LEVEL))
```

dyladan: Changed to level 1 default and added the suggested warning. I'm still concerned that users that already have it running in CI won't see it but at least the fix is easy when tests begin to fail in the future.

test/test.py
```diff
@@ -892,15 +920,35 @@ def test_multiple_requests_with_illegal_traceparent(self):
 	HARNESS_BIND_PORT  the port which the test harness binds to (default to HARNESS_PORT)
 	SERVICE_ENDPOINT   your test service endpoint (no default value)
 	STRICT_LEVEL       the level of test strictness (default 2)
+	SPEC_LEVEL         the minimum version of the Trace Context specification being tested (default 1)
```
reyang, line 923: Minor: consider `SPEC_VERSION` instead of `SPEC_LEVEL`?

test/README.md
```diff
@@ -79,12 +79,30 @@ Content-Type: application/json
 		HARNESS_BIND_HOST  the host/address which the test harness binds to (default to HARNESS_HOST)
 		HARNESS_BIND_PORT  the port which the test harness binds to (default to HARNESS_PORT)
 		SERVICE_ENDPOINT   your test service endpoint (no default value)
+		STRICT_LEVEL       the level of test strictness (default 2)
+		SPEC_LEVEL         the minimum version of the Trace Context specification being tested (default 2)
```
reyang, line 83: sort alphabetically

TimothyMothra:
@dyladan @kalyanaj 
The default value for SPEC_LEVEL in the Readme is 2, but the default in test.py is 1.

PR #511, opened 2022-12-01
improve description of version field in traceparent request header 

### LINKED ISSUE #507: Version number relationship to ASCII and UTF-8
3.2.2.1 version
https://www.w3.org/TR/trace-context-2/#version

> version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
>
> The value is US-ASCII encoded (which is UTF-8 compliant).
>
> Version (version) is 1 byte representing an 8-bit unsigned integer. Version ff is invalid. The current specification assumes the version is set to 00

The ABNF is `2HEXDIGLC` (which is two lowercase hex digits, e.g. \[0-9a-f\]). Mentioning US-ASCII or UTF-8 here is misleading and/or confusing, since non-ASCII characters are not permitted and EBCDIC would be unexpected 😉. It is the case that ASCII is a subset of UTF-8, but pointing this out doesn't really add anything to your spec. 

The text about the "version" being "1 byte" is also confusing, since the hex digits are two-bytes long. I would probably change your description to say:

> version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
>
> Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string. Version `ff` is invalid. The value can be stored as a single byte. This specification assumes the version is set to 00.

basti1302:
This fixes the confusing wording in the description of the version field, which previously didn't clearly distinguish between a possible internal representation (1 byte) and the format on the wire (two ascii chars). New wording (mostly) contributed by @aphillips.

fixes #507

This needs to be backported to level 2 once it has been merged to `main`.

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/511.html" title="Last updated on Dec 7, 2022, 7:28 AM UTC (83f43d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/511/f9d9e05...instana:83f43d6.html" title="Last updated on Dec 7, 2022, 7:28 AM UTC (83f43d6)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -83,9 +83,7 @@ The dash (`-`) character is used as a delimiter between fields.
 version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
 ```
 
-The value is US-ASCII encoded (which is UTF-8 compliant).
-
-Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `ff` is invalid. The current specification assumes the `version` is set to `00`.
+Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This specification assumes the version is set to 0 (`"00"`).
 
 #### version-format
 
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -83,7 +83,7 @@ The dash (`-`) character is used as a delimiter between fields.
 version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
 ```
 
-Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This specification assumes the version is set to 0 (`"00"`).
+Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This document specifies version 0 (`"00"`) of the `traceparent` header.
 
 #### version-format
 


kalyanaj (approved):

spec/20-http_request_header_format.md
```diff
@@ -83,9 +83,7 @@ The dash (`-`) character is used as a delimiter between fields.
 version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
 ```
 
-The value is US-ASCII encoded (which is UTF-8 compliant).
-
-Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `ff` is invalid. The current specification assumes the `version` is set to `00`.
+Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This specification assumes the version is set to 0 (`"00"`).
```
kalyanaj: To avoid ambiguity, it may be worth calling out that this is an ASCII case insensitive string (with a link to the corresponding spec), similar to what you have added in PR #513.

basti1302: It isn't, though, is it? We specify 
* `version = 2HEXDIGLC ` and
* `HEXDIGLC = DIGIT / "a" / "b" / "c" / "d" / "e" / "f" ; lowercase hex character`

So upper case letters `A` to `F` are invalid, aren't they?

kalyanaj: Thanks, I missed that earlier; in that case, the current wording sounds good to me.

dyladan: Same comment as https://github.com/w3c/trace-context/pull/512#discussion_r1041486068

dyladan (approved):

spec/20-http_request_header_format.md
basti1302: Good point. Fixed now.

basti1302:
For the record, the backport PR for level 2 is here: https://github.com/w3c/trace-context/pull/517

PR #512, opened 2022-12-01
improve description of version field in traceresponse header 

basti1302:
This fixes the confusing wording in the description of the version field, which previously didn't clearly distinguish between a possible internal representation (1 byte) and the format on the wire (two ascii chars).

This change ports the new description of the version field in the request header over to the response header and aligns both descriptions. See https://github.com/w3c/trace-context/pull/511.

In contrast to #511, this will not be backported to level 2 (level 2 does not have the response header).

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/512.html" title="Last updated on Dec 7, 2022, 7:29 AM UTC (ad4d9d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/512/f9d9e05...instana:ad4d9d0.html" title="Last updated on Dec 7, 2022, 7:29 AM UTC (ad4d9d0)">Diff</a>

### DIFF
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -34,12 +34,10 @@ The dash (`-`) character is used as a delimiter between fields.
 #### version
 
 ``` abnf
-version         = 2HEXDIGLC   ; this document assumes version 00. Version 255 is forbidden
+version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
 ```
 
-The value is US-ASCII encoded (which is UTF-8 compliant).
-
-Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `255` is invalid. The current specification assumes the `version` is set to `00`.
+Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This specification assumes the version is set to 0 (`"00"`).
 
 #### version-format
 
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -37,7 +37,7 @@ The dash (`-`) character is used as a delimiter between fields.
 version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
 ```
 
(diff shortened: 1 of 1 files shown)


spec/21-http_response_header_format.md
```diff
@@ -34,12 +34,10 @@ The dash (`-`) character is used as a delimiter between fields.
...
 
 ``` abnf
-version         = 2HEXDIGLC   ; this document assumes version 00. Version 255 is forbidden
+version         = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
 ```
 
-The value is US-ASCII encoded (which is UTF-8 compliant).
-
-Version (`version`) is 1 byte representing an 8-bit unsigned integer. Version `255` is invalid. The current specification assumes the `version` is set to `00`.
+Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This specification assumes the version is set to 0 (`"00"`).
```
kalyanaj, line 40: To avoid ambiguity, it might be worth mentioning that this is a case-insensitive ASCII string.

kalyanaj (approved):

spec/21-http_response_header_format.md
dyladan: Do we "assume" version is set to 00 or do we require it? We can keep the existing wording if it is easier but I think something like this is more accurate.

```suggestion
Version (`version`) is an 8-bit unsigned integer value, serialized as an ASCII string with two characters. Version 255 (`"ff"`) is invalid. This document specifies version 0 (`"00"`) of the traceparent header.
```

dyladan (approved):
LGTM % wording nit

spec/21-http_response_header_format.md
basti1302: Good point. This section is about the `traceresponse` header though. I'll fix that in a follow up commit.

basti1302: (For context, this is not yet the backport to level-2, I just ported the changes for traceparent also to the same fields for traceparent.)

basti1302: For the record, I commited this change and then rewrote the commit to say of the `traceresponse` header.

dyladan (approved):

basti1302:
For the record: Since this is for the response header, it is not applicable to be backported to level 2.

Issue #508, opened 2022-10-23
Why should keys be restricted to lowercase ASCII?


aphillips:
3.3.2.2.1 Key
https://www.w3.org/TR/trace-context-2/#key

> A key MUST begin with a lowercase letter or a digit and contain up to 256 characters including lowercase letters (a-z), digits (0-9), underscores (_), dashes (-), asterisks (*), forward slashes (/), and at signs (@).

The definitions here should be more rigorously laid out. The `key` is part of the `tracestate` header's value and is described as "...an identifier that describes the vendor". The identifier is restricted to lowercase ASCII letters, digits, and the punctuation shown above in the quote.

Allowing keys/values that are non-ASCII is one possibility here, since vendors presumably can be named anything. Restricting the *value* to an ASCII token is understandable because of the need for encoding efficiency, but the text should be clearer about the reasoning behind the ASCII restriction for the key. 

Note that key names like this are sometimes machine generated and a restriction to ASCII can be irksome when non-ASCII Unicode characters might be present in the data. The pre-computed portions probably should be well-known (ASCII !) tokens, but it's unclear why the names are otherwise restricted.

kalyanaj:
Thanks for the review and for filing this issue. We discussed this in the Distributed Tracing working group meeting today and below captures the summary from that discussion.

Since this is Level 2 of the spec, the primary reason is back compatibility with existing Level 1 standard.

Some of the original reasons (from Level 1) for restricting the keys and values:

1. Processing efficiency: For example, an implementation could split on commas to get the list of entries.
2. If we extended the allowed character list, we would have had to have specific escape characters or disallowed characters (to meet the processing efficiency goal) which would have added more complexity to the specification.

We will update the specification to include the above rationale.

At this point, the main reason to not revisit this decision is backwards compatibility since Level 1 is already a standard.

aphillips:
> Some of the original reasons (from Level 1) for restricting the keys and values:
> 1.  Processing efficiency: For example, an implementation could split on commas to get the list of entries.
> 2.  If we extended the allowed character list, we would have had to have specific escape characters or disallowed characters (to meet the processing efficiency goal) which would have added more complexity to the specification.

I think these reasons are questionable.

The ASCII comma is still a comma. Disallowing a small list of characters (comma, space, etc.) because they are syntax-meaningful would allow implementations to split on commas. There is no need to introduce escapes unless you want to allow restricted characters (comma, space) into the namespace.

[...]

PR #515, opened 2022-12-01
add reasoning for the restriction of tracestate keys

basti1302:
Motivated by the request in #508

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/515.html" title="Last updated on Dec 1, 2022, 9:24 AM UTC (6c341a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/515/f9d9e05...instana:6c341a5.html" title="Last updated on Dec 1, 2022, 9:24 AM UTC (6c341a5)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -300,6 +300,8 @@ lcalpha    = %x61-7A ; a-z
 
 A `key` MUST begin with a lowercase letter or a digit and contain up to 256 characters including lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), forward slashes (`/`), and at signs (`@`).
 
+This restriction ensures that the `tracestate` header can be parsed efficiently into individual list members by splitting the header value at each comma, and splitting the list members an implementation is interested in at the equals sign.
+
 ###### Value
 
 The value is an <a href="#opaque">opaque</a> string containing up to 256 printable ASCII [[!RFC0020]] characters (i.e., the range 0x20 to 0x7E) except comma (,) and (=). The string must end with a character which is not a space (0x20). Note that this also excludes tabs, newlines, carriage returns, etc. All leading spaces MUST be preserved as part of the value. All trailing spaces are considered to be optional whitespace characters not part of the value. Optional trailing whitespace MAY be excluded when propagating the header.


spec/20-http_request_header_format.md
```diff
@@ -300,6 +300,8 @@ lcalpha    = %x61-7A ; a-z
 
 A `key` MUST begin with a lowercase letter or a digit and contain up to 256 characters including lowercase letters (`a`-`z`), digits (`0`-`9`), underscores (`_`), dashes (`-`), asterisks (`*`), forward slashes (`/`), and at signs (`@`).
 
+This restriction ensures that the `tracestate` header can be parsed efficiently into individual list members by splitting the header value at each comma, and splitting the list members an implementation is interested in at the equals sign.
```
aphillips, line 303: As noted in my [comment](https://github.com/w3c/trace-context/issues/508#issuecomment-1335454580), this reasoning is incorrect. Nothing about Unicode text, particularly UTF-8, is less-efficient for parsing, including looking for comma/equals signs.

While there really is no *good* reason (from an I18N point-of-view) for this restriction, I understand that it *feels* simpler to restrict to ASCII. Thinking about it at length, I'd suggest the following instead:

> This restriction on the `tracestate` header ensures that even very low-level and simple computing devices can process, validate, and display headers without having to worry about mid-character truncation or certain classes of display issues due to the use of e.g. bidi controls.

basti1302: I understand why you say that the phrase I added is actually incorrect, even if that (flawed) thinking might have been the original reason for the design choice to limit the character set the way we do. 

But to be honest I'm skeptical about your suggested phrasing as well:
* You say there is "no good reason" for this restriction. Well, we don't need to make up a reason. We already [established](https://github.com/w3c/trace-context/issues/508#issuecomment-1333468843) that by now the main reason to keep this restriction is _backwards compatibility_.  
* _Displaying_ values from the header to users is not a use case this spec is concerned with. The [Problem Statement](https://www.w3.org/TR/trace-context/#problem-statement) and [Solution](https://www.w3.org/TR/trace-context/#solution) section clearly call out what this spec is about: Trace context propagation. The trace context is of no concern to users, it is a mechanism for observability tooling. Displaying individual key-value pairs to users might be something that is relevant for the baggage spec but not for this spec.

I see two ways forward here:
* We leave this unchanged and do not add any reasoning at all. Yes, the restriction is not ideal from an i18n point of view, but it cannot be easily changed now. Adding a reasoning doesn't change that. Adding a partially made up reason just for the sake of adding a reason doesn't add value IMO.
* If you insist on us adding a reasoning here, I think we could meet halfway with either of those phrasings: 

> This restriction on the `tracestate` header ensures that even low-level computing devices can process and validate it.

or

> This restriction on the `tracestate` header ensures that even low-level computing devices can process and validate it, without having to worry about mid-character truncation.

aphillips: Actually, I don't insist that you add a reason--just that, if you do have a reason, that it make sense.

I don't buy that the header values are not displayed. While this spec may not be concerned with the display of values, you do need to think about the larger trace ecosystem. The keys have meaningful structure, presumably because people will use them for various things. Otherwise you could just generate completely opaque values.

I would suggest that you just drop this PR? I'd be satisfied with that.

basti1302: Fine, then let's do that. :-)  I assume we can close #508 then?

aphillips: Yep. And I'll add a _satisfied_ comment to that issue when you do.

basti1302: 👍

basti1302:
Closing this after the discussion in https://github.com/w3c/trace-context/pull/515#pullrequestreview-1202739206

spec/20-http_request_header_format.md
basti1302: @aphillips done.

PR #533, opened 2023-04-12
Added random trace id flag to traceresponse header

### LINKED ISSUE #532: Incorporate Level 2 changes (random trace id flag) into traceresponse header in Level 3
Incorporate Level 2 changes (random trace id flag) into traceresponse header in Level 3.

kalyanaj:
Resolves #532 .

***
<a href="https://pr-preview.s3.amazonaws.com/kalyanaj/trace-context/pull/533.html" title="Last updated on May 19, 2023, 12:21 AM UTC (a963ece)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/533/a4e20e7...kalyanaj:a963ece.html" title="Last updated on May 19, 2023, 12:21 AM UTC (a963ece)">Diff</a>

### DIFF
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -53,13 +53,13 @@ trace-flags      = 2HEXDIGLC   ; 8 bit flags. See below for details
 
 The format and requirements for this are the same as those of the trace-id field in the `traceparent` request header.
 
-For details, see the trace-id section under [traceparent Header Field Values](#traceparent Header Field Values).
+For details, see the trace-id section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
 
 #### child-id
 
 This is the ID of the operation of the callee (in some tracing systems, this is known as the `span-id`, where a `span` is the execution of a client request) and is used to uniquely identify an operation within a trace. It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
 
-Vendors MUST ignore the `traceresponse` when the `child-id` is invalid (for example, if it contains non-lowercase hex characters).
+Vendors MUST ignore the `traceresponse` header when the `child-id` is invalid (for example, if it contains non-lowercase hex characters).
 
 #### trace-flags
 
@@ -102,7 +102,7 @@ The second least significant bit of the trace-flags field denotes the random-tra
 
 The format and requirements for this are the same as those of the random-trace-id flag in the trace-flags field in the `traceparent` request header.
 
-For details, see the trace-flags section under [traceparent Header Field Values](#traceparent Header Field Values).
+For details, see the trace-flags section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
 
 
 ##### Other Flags
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -63,11 +63,11 @@ Vendors MUST ignore the `traceresponse` header when the `child-id` is invalid (f
 
 #### trace-flags
 
-Similar to the `trace-flags` field in the `traceparent` request header, this field is hex-encoded.
+Similar to the `trace-flags` field in the `traceparent` request header, this is a hex-encoded <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a callee handled the trace. The same requirement to properly mask the bit field value when interpreting it applies here as well.
 
 The current version of this specification (`00`) supports only two flags: `sampled` and `random-trace-id`.
 
-This is an <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a callee handled the trace. These flags are recommendations given by a callee rather than strict rules to follow for three reasons:
+These flags are recommendations given by a callee rather than strict rules to follow for three reasons:
 
 1. An untrusted callee may be able to abuse a tracing system by setting these flags maliciously.
 2. A callee may have a bug which causes the tracing system to have a problem.
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -21,7 +21,6 @@ In order to increase interoperability across multiple protocols and encourage su
 
 ### traceresponse Header Field Values
 
-
 This section uses the Augmented Backus-Naur Form (ABNF) notation of [[!RFC5234]], including the DIGIT rule from that document. The `DIGIT` rule defines a single number character `0`-`9`.
 
 ``` abnf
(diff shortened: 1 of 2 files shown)


spec/21-http_response_header_format.md
```diff
@@ -52,18 +51,28 @@ trace-flags      = 2HEXDIGLC   ; 8 bit flags. Currently, only one bit is used. S
 
 #### trace-id
```
yurishkuro, line 53: the notion of trace-id in the response header should be identical to trace-in in `traceparent`. Why can we not link there instead of re-defining and potentially diverging?

kalyanaj: Good point, sure I have taken a pass at updating the description for a couple of fields (trace-id and one of the flags) to link to the corresponding fields in traceparent header. We likely want to keep the description for the sampled flag independently to express the callee/caller semantics easily.

SergeyKanzhelev (approved):

spec/21-http_response_header_format.md
```diff
@@ -47,47 +46,37 @@ The following `version-format` definition is used for version `00`.
...
 
-This is the ID of the operation of the callee (in some systems known as the span id) and is used to uniquely identify an operation within a trace. It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
+This is the ID of the operation of the callee (in some tracing systems, this is known as the `span-id`, where a `span` is the execution of a client request) and is used to uniquely identify an operation within a trace. It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
+
+Vendors MUST ignore the `traceresponse` header when the `child-id` is invalid (for example, if it contains non-lowercase hex characters).
 
 #### trace-flags
 
-An <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides additional information about how the callee handled the trace such as sampling, trace level, etc. These flags are recommendations given by the callee rather than strict rules to follow for three reasons:
+Similar to the `trace-flags` field in the `traceparent` request header, this field is hex-encoded.
```
basti1302: _suggestion:_

Maybe we can combine the two comments on the format (hex encode, 8-bit) here, instead of splitting them across two paragraphs with the list of available flags in the middle? 

We could add the 8-bit aspect here and shorten the paragraph below accordingly. 

```suggestion
Similar to the `trace-flags` field in the `traceparent` request header, this is a hex-encoded <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a callee handled the trace. The same requirement to properly mask the bit field value when interpreting it applies here as well.
```

The bit 

> The same requirement to properly mask the bit-field value when interpreting it applies here as well.

is probably debatable, as we have this spelled out in trace-parent/trace-flags already. I personally feel it is worth mentioning it again.

spec/21-http_response_header_format.md
```diff
@@ -47,47 +46,37 @@ The following `version-format` definition is used for version `00`.
...
+
+The current version of this specification (`00`) supports only two flags: `sampled` and `random-trace-id`.
+
+This is an <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a callee handled the trace. These flags are recommendations given by a callee rather than strict rules to follow for three reasons:
```
basti1302: _suggestion_: 

If we want to go with the suggestion above, we should remove the format aspects here.

```suggestion
These flags are recommendations given by a callee rather than strict rules to follow for three reasons:
```

basti1302 (approved):
Looks good to me. I have added two small optional suggestions.

spec/21-http_response_header_format.md
```diff
@@ -107,6 +96,15 @@ There are two additional options that tracing systems MAY follow:
 - A component that makes a deferred or delayed recording decision may communicate the priority of a recording by setting `sampled` flag to `1` for a subset of requests.
 - A component may also fall back to probability sampling and set the `sampled` flag to `1` for the subset of requests.
 
+##### Random Trace ID Flag
+
+The second least significant bit of the trace-flags field denotes the random-trace-id flag.
+
+The format and requirements for this are the same as those of the random-trace-id flag in the trace-flags field in the `traceparent` request header.
+
+For details, see the trace-flags section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
```
basti1302: _issue:_

Thinking about this again, one more aspect came to my mind.

I think it is pretty clear how participants are ought to behave that _start_ a trace. If they generate the trace id randomly, they set the flag in the response header to signal the randomness upstream. 

But what about participants which continue a trace that has been started upstream? They receive a value for the flag in the request header from the upstream participant. They also potentially receive multiple response headers from requests they send downstream. What if the flag values are not identical? This should not happen in practice I think, but we still should spell out explicit rules here IMO.

It might be fairly obvious I think: When continuing a trace we always need to respect the decision of the participant that generated the trace id. Only that participant knows how it generated the id. Thus, a participant that continues the trace MUST set the flag to the same value as in the incoming request traceparent header. Even if it is obvious, I think we should still state this explicitly.

That said, the benefit of the flag in the response header is questionable anyway. You are telling the upstream participant something it already knows anyway. But of course we need it for consistency, since the flag has value in the response header when the participant started the trace.

basti1302 (changes requested):

spec/21-http_response_header_format.md
kalyanaj: Thanks Bastian. Regarding "You are telling the upstream participant something it already knows anyway. ", there is a case where the upstream participant didn't generate the trace-id (i.e., the trace was started by the downstream participant), and in that case the upstream participant would benefit from this random-trace-id flag in the response header. I will add a note to this effect.

Apart from the above, given that you have already addressed how the flag should be set on trace continuation, do you see the need for clarifying anything else here?

basti1302: I was thinking about some explicit normative language like the following:

* A participant that starts or restarts the trace by generating a new `trace-id` MUST apply the rules for (re)starting a trace stated in the section on the random trace ID for `traceparent` for the flag value in the `traceresponse` header as well.
* A participant that uses the same `trace-id` in `traceresponse` as it received in the incoming `traceparent` header MUST also set the `random-trace-id` flag to the same value that was found in that `traceparent` header.
* If a participant did not receive an incoming `traceparent` header, but it received a `traceresponse` header from downstream, and it wants to communicate the `trace-id` from that `traceresponse` header upstream, it MUST use the same `random-trace-id` flag value as in that `traceresponse` header.

Side note: The last bullet point refers to the non-normative example in the section "Load Balancer Deferred Sampling".

Alternatively, we can also drop this point for now and revisit it later. That is merge the PR as-is, and iterate on the normative wording later on.

basti1302 (approved):

spec/21-http_response_header_format.md
kalyanaj: Thanks Bastian. Filed #540 to track this suggestion.

basti1302: Thanks for moving that out into its own issue. 👍

PR #536, opened 2023-05-04
explicitly describe rules for randomness flag when continuing a trace

basti1302:
When implementing level 2, I found that there is little guidance on how to handle the new random-trace-id flag when continuing a trace. IMO, the rules we added so far mostly implicitly only apply when starting a trace. One could argue that the rules around how to handle the flag when continuing a trace follow implicitly from the rules for starting a trace, or from the stated intention. I think this is not good enough. I think we should spell this out explicitly.

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/536.html" title="Last updated on May 9, 2023, 7:32 PM UTC (698f4a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/536/a08b822...instana:698f4a1.html" title="Last updated on May 9, 2023, 7:32 PM UTC (698f4a1)">Diff</a>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -182,12 +182,18 @@ There are two additional options that vendors MAY follow:
 
 ##### Random Trace ID Flag
 
-The second least significant bit of the trace-flags field denotes the random-trace-id flag.
-If that flag is set, at least the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
-If the flag is not set, the trace ID MAY still be randomly (or pseudo-randomly) generated.
-When unset, the trace ID MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
+The second least significant bit of the trace-flags field denotes the `random-trace-id` flag.
 
-When at least the right-most 7 bytes of the `trace-id` are randomly (or pseudo-randomly) generated, the random trace ID flag SHOULD be set to `1`.
+When starting or restarting a trace (that is, when the participant generates a new `trace-id`), the following rules apply:
+* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be random (or pseudo-random).
+* If the flag is not set, the `trace-id` MAY still be randomly (or pseudo-randomly) generated.
+* When unset, the `trace-id` MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
+* When at least the right-most 7 bytes of the `trace-id` are randomly (or pseudo-randomly) generated, the `random-trace-id` flag SHOULD be set to `1`.
+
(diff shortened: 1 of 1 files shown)


spec/20-http_request_header_format.md
```diff
@@ -184,12 +184,18 @@ There are two additional options that vendors MAY follow:
...
-If that flag is set, at least the right-most 7 bytes of the trace ID MUST be random (or pseudo-random).
-If the flag is not set, the trace ID MAY still be randomly (or pseudo-randomly) generated.
-When unset, the trace ID MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
+The second least significant bit of the trace-flags field denotes the `random-trace-id` flag.
+
+When starting or restarting a trace (that is, when the participant generates a new `trace-id`), the following rules apply:
+* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be random (or pseudo-random).
+* If the flag is not set, the `trace-id` MAY still be randomly (or pseudo-randomly) generated.
+* When unset, the `trace-id` MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
+* When at least the right-most 7 bytes of the `trace-id` are randomly (or pseudo-randomly) generated, the `random-trace-id` flag SHOULD be set to `1`.
```
basti1302, line 191: These four bullet points are what we had before. The only change I made was to replace "trace ID" with `trace-id` etc., to be in line with the new [style guide](https://github.com/w3c/distributed-tracing-wg/blob/main/STYLE.md#referencing-fields).

spec/20-http_request_header_format.md
```diff
@@ -184,12 +184,18 @@ There are two additional options that vendors MAY follow:
...
+
+When continuing a trace (that is, the incoming HTTP request had the `traceparent` header and the participant uses the same `trace-id` in the `traceparent` header on outgoing requests), the following rules apply:
+* If the flag is set in the incoming `traceparent` header, it MUST also be set in all outgoing `traceparent` headers which use the same `trace-id`.
+* If the flag is unset in the incoming `traceparent` header, it MUST also be unset in any outgoing `traceparent` headers which use the same `trace-id`.
```
basti1302, line 195: This part is new. It reflects my understanding of how things need to be handled when continuing a trace. It also aligns with the [test case](https://github.com/w3c/trace-context/blob/d7ad23a21504d44a809f66b01c82e5a263e6c68e/test/test.py#L893-L905) we have for propagating the flag downstream.

basti1302:
If we add a distinction like this, it might have an impact on what we say about the same flag in the [response header](https://github.com/w3c/trace-context/pull/533/files#diff-724c97a1367f014afb98eec8dce24d15beaf75925c4ecec8af9955852c915d19R99-R105). But this should probably be a separate discussion, and a separate PR if changes are required there as well.

dyladan (approved):

kalyanaj (approved):

PR #537, opened 2023-05-09
Always run tests

dyladan:
Do not filter tests based on path. Tests take ~10 seconds so running always is no blocker to anyone


basti1302 (approved):
thanks!

SergeyKanzhelev (approved):

PR #547, opened 2023-10-16
Update tracestate duplicate keys in test_data.json

pellared:
Follows https://github.com/w3c/trace-context/pull/477

Relates to https://github.com/w3c/trace-context/issues/548


pellared:
@basti1302 @yurishkuro @dyladan @kalyanaj PTAL

test/test_data.json
```diff
@@ -42,8 +42,8 @@
...
-	{"headers": [["traceparent", "00-12345678901234567890123456789012-1234567890123456-00"], ["tracestate", "foo=1,foo=1"]], "is_traceparent_valid": true, "is_tracestate_valid": false},
-	{"headers": [["traceparent", "00-12345678901234567890123456789012-1234567890123456-00"], ["tracestate", "foo=1,foo=2"]], "is_traceparent_valid": true, "is_tracestate_valid": false},
+	{"headers": [["traceparent", "00-12345678901234567890123456789012-1234567890123456-00"], ["tracestate", "foo=1,foo=1"]], "is_traceparent_valid": true, "is_tracestate_valid": true},
+	{"headers": [["traceparent", "00-12345678901234567890123456789012-1234567890123456-00"], ["tracestate", "foo=1,foo=2"]], "is_traceparent_valid": true, "is_tracestate_valid": true},
 	{"headers": [["traceparent", "00-12345678901234567890123456789012-1234567890123456-00"], ["tracestate", "foo=1"], ["tracestate", "foo=1"]], "is_traceparent_valid": true, "is_tracestate_valid": false},
 	{"headers": [["traceparent", "00-12345678901234567890123456789012-1234567890123456-00"], ["tracestate", "foo=1"], ["tracestate", "foo=2"]], "is_traceparent_valid": true, "is_tracestate_valid": false},
```
basti1302: I think this change is correct. For context, it seems like self_test.py _writes_ this file. Actually, for me it also changes the next two cases to `"is_tracestate_valid": true` when I run this on `main` -- they are basically the same cases, just that the headers are sent as multiline headers. @pellared would you mind changing lines 47 and 48 accordingly?

@dyladan would you agree? Also, it seems the file is only ever written to, by self_test.py. Should it rather be in `.gitignore` instead of being under source control?

pellared: Done https://github.com/w3c/trace-context/pull/547/commits/85899d0177363ba816eaf899f86843b456c3607c

basti1302 (approved):

test/test_data.json
basti1302: I have created #553 as an alternative for removing the file from version control. We should either merge this PR here or #553. Still would like to get @dyladan's input on this.

PR #555, opened 2023-12-12
explict rules for traceresponse/random-trace-id when continuing a trace

### LINKED ISSUE #540: Add more normative text for flags processing in traceresponse
@basti1302 had suggestions in #533 for normative language related to flags in traceresponse, filing this issue to track it:

From @basti1302 :

I was thinking about some explicit normative language like the following:

- A participant that starts or restarts the trace by generating a new trace-id MUST apply the rules for (re)starting a trace stated in the section on the random trace ID for traceparent for the flag value in the traceresponse header as well.

- A participant that uses the same trace-id in traceresponse as it received in the incoming traceparent header MUST also set the random-trace-id flag to the same value that was found in that traceparent header.

- If a participant did not receive an incoming traceparent header, but it received a traceresponse header from downstream, and it wants to communicate the trace-id from that traceresponse header upstream, it MUST use the same random-trace-id flag value as in that traceresponse header.

Side note: The last bullet point refers to the non-normative example in the section "Load Balancer Deferred Sampling".

basti1302:
Fixes https://github.com/w3c/trace-context/issues/540.

***
<a href="https://pr-preview.s3.amazonaws.com/instana/trace-context/pull/555.html" title="Last updated on Dec 12, 2023, 7:52 PM UTC (5c7f3f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/555/2178be0...instana:5c7f3f5.html" title="Last updated on Dec 12, 2023, 7:52 PM UTC (5c7f3f5)">Diff</a>

### DIFF
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -107,7 +107,7 @@ When a participant starts or restarts a trace (that is, when the participant gen
 
 A participant that continues a trace started upstream &mdash; that is, if the participant uses the `trace-id` value from an incoming `traceparent` header in its own `traceresponse` header &mdash; MUST set the `random-trace-id` flag in the `traceresponse` header to the same value that was found in the incoming `traceparent` header.
 
-A participant that continues a trace started downstream &mdash; that is, if the participant uses the `trace-id` value from a `traceresponse` header it has receieved &mdash; MUST set the `random-trace-id` flag in its own `traceresponse` header to the same value that was found in the `traceresponse` header from which the `trace-id` was taken.
+A participant that continues a trace started downstream &mdash; that is, if the participant uses the `trace-id` value from a `traceresponse` header it has received &mdash; MUST set the `random-trace-id` flag in its own `traceresponse` header to the same value that was found in the `traceresponse` header from which the `trace-id` was taken.
 
 ##### Other Flags
 
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -53,7 +53,7 @@ trace-flags      = 2HEXDIGLC   ; 8 bit flags. See below for details
 
 The format and requirements for this are the same as those of the trace-id field in the `traceparent` request header.
 
-For details, see the trace-id section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
+For details, see the trace-id section under [traceparent Header Field Values](#traceparent-header-field-values).
 
 #### child-id
 
@@ -105,7 +105,7 @@ the specification for this flag.
 
 The format and requirements for this are the same as those of the random-trace-id flag in the trace-flags field in the `traceparent` request header.
 
-For details, see the trace-flags section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
+For details, see the trace-flags section under [traceparent Header Field Values](#traceparent-header-field-values).
 
 
 ##### Other Flags
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -103,10 +103,11 @@ The second least significant bit of the trace-flags field denotes the random-tra
 If a trace was started by a downstream participant and it responds with the `traceresponse` HTTP header, an upstream participant can use this flag to determine if the `trace-id` was generated as per
 the specification for this flag.
 
-The format and requirements for this are the same as those of the random-trace-id flag in the trace-flags field in the `traceparent` request header.
+When a participant starts or restarts a trace (that is, when the participant generates a new `trace-id`), the requirements for this flag are the same as those for the random-trace-id flag in the trace-flags field in the `traceparent` request header. For details, see the section [Random Trace ID Flag](#random-trace-id-flag).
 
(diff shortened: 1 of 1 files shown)


spec/21-http_response_header_format.md
```diff
@@ -53,7 +53,7 @@ trace-flags      = 2HEXDIGLC   ; 8 bit flags. See below for details
 
 The format and requirements for this are the same as those of the trace-id field in the `traceparent` request header.
 
-For details, see the trace-id section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
+For details, see the trace-id section under [traceparent Header Field Values](#traceparent-header-field-values).
```
basti1302, line 56: This is unrelated the the PR, but this link was broken (md anchors are always lower case only)

spec/21-http_response_header_format.md
```diff
@@ -63,7 +63,7 @@ Vendors MUST ignore the `traceresponse` header when the `child-id` is invalid (f
 
 #### trace-flags
 
-Similar to the `trace-flags` field in the `traceparent` request header, this is a hex-encoded <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a callee handled the trace. The same requirement to properly mask the bit field value when interpreting it applies here as well.
+Similar to the [`trace-flags` field)(#trace-flags) in the `traceparent` request header, this is a hex-encoded <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a callee handled the trace. The same requirement to properly mask the bit field value when interpreting it applies here as well.
```
basti1302: I just added a link here to the traceparent flags section. Since we reference that section, linking to it wouldn't hurt either I think.

spec/21-http_response_header_format.md
```diff
@@ -103,10 +103,11 @@ The second least significant bit of the trace-flags field denotes the random-tra
 If a trace was started by a downstream participant and it responds with the `traceresponse` HTTP header, an upstream participant can use this flag to determine if the `trace-id` was generated as per
 the specification for this flag.
 
-The format and requirements for this are the same as those of the random-trace-id flag in the trace-flags field in the `traceparent` request header.
+When a participant starts or restarts a trace (that is, when the participant generates a new `trace-id`), the requirements for this flag are the same as those for the random-trace-id flag in the trace-flags field in the `traceparent` request header. For details, see the section [Random Trace ID Flag](#random-trace-id-flag).
```
basti1302, line 106: 1. I dropped the phrase "the format" -- since section is about an individual flag, there is no need to explain the format. We state that the traceresponse flags field is a hex encoded 8-bit field at the start of the trace-flags section. Flags don't have an individual format.
2. I added a new constraint to this sentence: The rules from traceparent/random-trace-id flags are now only applied when the participant starts the trace. In my opinion, we cannot just refer to traceparent/random-trace-id for the situation when continuing a trace, because the a trace can be continued from upstream or from downstream, and different rules need to be applied. See below.

spec/21-http_response_header_format.md
```diff
@@ -103,10 +103,11 @@ The second least significant bit of the trace-flags field denotes the random-tra
...
 
-For details, see the trace-flags section under [traceparent Header Field Values](#traceparent-Header-Field-Values).
```
basti1302, line 108: I removed this link. Instead, this section now links directly to the random-trace-id section in the request header section. This section has all the rules and information that are relevant here. IMO the only reason to link to the whole traceparent-header-field-values section would be if we also wanted to include this line from the trace-id subsection: 

https://github.com/w3c/trace-context/blob/main/spec/20-http_request_header_format.md?plain=1#L108

> If the right-most 7 bytes are randomly (or pseudo-randomly) generated, the corresponding [random trace id flag](#random-trace-id-flag) SHOULD be set.

But that particular rule is [repeated](https://github.com/w3c/trace-context/blob/main/spec/20-http_request_header_format.md?plain=1#L191) in the random-trace-id section as well.

yurishkuro (approved):

spec/21-http_response_header_format.md
```diff
@@ -103,10 +103,11 @@ The second least significant bit of the trace-flags field denotes the random-tra
...
+A participant that continues a trace started upstream &mdash; that is, if the participant uses the `trace-id` value from an incoming `traceparent` header in its own `traceresponse` header &mdash; MUST set the `random-trace-id` flag in the `traceresponse` header to the same value that was found in the incoming `traceparent` header.
 
+A participant that continues a trace started downstream &mdash; that is, if the participant uses the `trace-id` value from a `traceresponse` header it has receieved &mdash; MUST set the `random-trace-id` flag in its own `traceresponse` header to the same value that was found in the `traceresponse` header from which the `trace-id` was taken.
```
kalyanaj:
```suggestion
A participant that continues a trace started downstream &mdash; that is, if the participant uses the `trace-id` value from a `traceresponse` header it has received &mdash; MUST set the `random-trace-id` flag in its own `traceresponse` header to the same value that was found in the `traceresponse` header from which the `trace-id` was taken.
```

kalyanaj (approved):

jcopi (approved):

PR #558, opened 2024-01-30
fix: ensure random part of trace id is uniformly distributed

### LINKED ISSUE #544: Random part of trace ID should be uniformly distributed
Currently, it is not mentioned that the random part of the trace ID (when the random flag is set) should be uniformly distributed. Compare https://w3c.github.io/trace-context/#trace-id, https://w3c.github.io/trace-context/#random-trace-id-flag, and https://w3c.github.io/trace-context/#randomness-of-trace-id. To be useful for sampling, it should be explicitly stated that the random part should follow a uniform distribution, meaning that all possible 2^56 values are equally likely. Otherwise, if we cannot rely on the uniform distribution, we would again need to hash the trace ID, which is exactly what we want to avoid with the random part.

FYI @PeterF778 @jmacd

dyladan:
Partially addresses #544

***

### :boom: Error: 502 Bad Gateway :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 1, 2024, 9:04 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fdynatrace-oss-contrib%2Ftrace-context%2Fcd8ca6f6d66804abae0cb0c491e585ce32651079%2Findex.html%3FisPreview%3Dtrue)

```
error code: 502
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/trace-context%23558.)._
</details>

### DIFF
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -185,7 +185,7 @@ There are two additional options that vendors MAY follow:
 The second least significant bit of the trace-flags field denotes the `random-trace-id` flag.
 
 When starting or restarting a trace (that is, when the participant generates a new `trace-id`), the following rules apply:
-* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be random (or pseudo-random).
+* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be selected randomly (or pseudo-randomly) from a uniform distribution of all `2^56` possible values.
 * If the flag is not set, the `trace-id` MAY still be randomly (or pseudo-randomly) generated.
 * When unset, the `trace-id` MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
 * When at least the right-most 7 bytes of the `trace-id` are randomly (or pseudo-randomly) generated, the `random-trace-id` flag SHOULD be set to `1`.
--- a/spec/20-http_request_header_format.md
+++ b/spec/20-http_request_header_format.md
@@ -185,7 +185,7 @@ There are two additional options that vendors MAY follow:
 The second least significant bit of the trace-flags field denotes the `random-trace-id` flag.
 
 When starting or restarting a trace (that is, when the participant generates a new `trace-id`), the following rules apply:
-* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be selected randomly (or pseudo-randomly) from a uniform distribution of all `2^56` possible values.
+* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be selected randomly (or pseudo-randomly) with uniform distribution over the interval [0..2^56-1].
 * If the flag is not set, the `trace-id` MAY still be randomly (or pseudo-randomly) generated.
 * When unset, the `trace-id` MAY be generated in any way that satisfies the requirements of the [trace ID format](#trace-id).
 * When at least the right-most 7 bytes of the `trace-id` are randomly (or pseudo-randomly) generated, the `random-trace-id` flag SHOULD be set to `1`.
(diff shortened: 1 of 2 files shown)


kalyanaj (approved):

spec/20-http_request_header_format.md
```diff
@@ -185,7 +185,7 @@ There are two additional options that vendors MAY follow:
 The second least significant bit of the trace-flags field denotes the `random-trace-id` flag.
 
 When starting or restarting a trace (that is, when the participant generates a new `trace-id`), the following rules apply:
-* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be random (or pseudo-random).
+* If that flag is set, at least the right-most 7 bytes of the `trace-id` MUST be selected randomly (or pseudo-randomly) from a uniform distribution of all `2^56` possible values.
```
PeterF778: My concern is that this language seems to exclude random generators with periods smaller than 2^56 (such as java.util.Random).

oertl: Generators with such a short period should not be used here anyway. java.util.Random has already been superseded in Java 8 by java.util.SplittableRandom with a period of 2^64.

dyladan: @PeterF778 the distribution of the values must be known in advance in order for the flag to be useful. The random bit has already been defined to affect 7 bytes. This PR is just clarifying the exact meaning of that requirement. If you want to make an argument that 56 bits is too many and we should reduce the requirement, I think that should be done in a separate issue.

PeterF778: No, it was not my intention. I just would like to see a weaker condition: if X is a random value described by the lower 7 bytes (0 <= X < 2^56), and N is a number from [0..2^56-1], then P(X<N) =~ N / 2^56. 
This can be also formulated simply as "the right-most 7 bytes must be selected randomly with uniform distribution over the interval [0..2^56-1].

PeterF778: > Generators with such a short period should not be used here anyway. java.util.Random has already been superseded in Java 8 by java.util.SplittableRandom with a period of 2^64.

I still think that requiring long periods for the random number generators used to generate trace-id is an unnecessary burden for all platforms. The longer period the better, but what are really the downsides of using a generator with a relatively low period of 2^32 (at least for the purpose of consistent probability sampling).

oertl: A downside is that sampling with probabilities in the order of 2^(-32) (1 out of 4B) or less would not work. And it is not completely unrealistic to use such sampling probabilities. I would even expect that the actual sampling probability would differ significantly from the selected probability, even for much larger probabilities (e.g. 1 out of 100M).

dyladan: @oertl the wording suggested is what we're trying to achieve anyway so I don't think I have a problem with it:

"the right-most 7 bytes must be selected randomly with uniform distribution over the interval [0..2^56-1]"

WDYT?

oertl: @dyladan I think the wording as suggested is fine.

oertl: > I still think that requiring long periods for the random number generators used to generate trace-id is an unnecessary burden for all platforms.

I do not think that it is a burden to use a reasonable pseudorandom number generator like splitmix64 (used by java.util.SplittableRandom) which only require a few operations to generate a 64-bit random value (compare https://prng.di.unimi.it/splitmix64.c).

kalyanaj (approved):

jcopi (approved):

PR #560, opened 2024-02-27
Use server-timing for trace context response

dyladan:
This is a first draft of the server-timing response header. It is not meant to be a final version, but a place to start discussion. It is mostly a direct translation with the following exceptions:

### Version is optional

In the original version, all fields were required in order to simplify parsing. Now that we are a small component of a more complex header, that simplicity is lost, and the parsing simplification is no longer beneficial.

### Flags are optional

The reasoning is the same as above. Additionally, some servers may not want to reveal information like the sampled flag to an untrusted client in order to avoid abuse.

### Terminology Fixes

Changed ambiguous or nonstandard terms like "callee" and "caller" to established terms like "server" and "client."

***
<a href="https://pr-preview.s3.amazonaws.com/dynatrace-oss-contrib/trace-context/pull/560.html" title="Last updated on Aug 25, 2025, 8:29 PM UTC (88ba098)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/560/bc3c536...dynatrace-oss-contrib:88ba098.html" title="Last updated on Aug 25, 2025, 8:29 PM UTC (88ba098)">Diff</a>

### DIFF
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -4,56 +4,55 @@ This section describes the binding of the distributed trace context to a metric 
 
 ## Trace Context Metric
 
-The trace context metric identifies a completed request in a tracing system. It has four params:
-
-* `tid` - required
-* `cid` - required
-* `v` - optional
-* `flags` - optional
+The trace context metric identifies a completed request in a tracing system.
+It consists of a {{PerformanceServerTiming/description}}.
+{{PerformanceServerTiming/duration}} is reserved for future use.
 
 Example server timing header with trace context metric:
 
 ```
-server-timing: trace;tid=0af7651916cd43dd8448eb211c80319c,cid=b7ad6b7169203331
+server-timing: trace;desc=00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
 ```
 
 ### Metric Name
 
 Metric name: `trace`
 
-The metric name is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive). That is, `trace`, `Trace`, and `TRACE` are considered the same metric.
+The metric {{PerformanceServerTiming/name}} is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive). That is, `trace`, `Trace`, and `TRACE` are considered the same metric.
 
 In order to increase interoperability across multiple protocols and encourage successful integration, tracing systems SHOULD encode the metric name as [ASCII lowercase](https://infra.spec.whatwg.org/#ascii-lowercase).
 
-### Trace Context Metric Param Values
+### Metric Description Format
 
 This section uses the Augmented Backus-Naur Form (ABNF) notation of [[!RFC5234]], including the DIGIT rule from that document. The `DIGIT` rule defines a single number character `0`-`9`.
 
 ```abnf
-HEXDIGLC = DIGIT / "a" / "b" / "c" / "d" / "e" / "f" ; lowercase hex character
-tid      = 32HEXDIGLC
-cid      = 16HEXDIGLC
-flags    = 2HEXDIGLC
-v        = 2HEXDIGLC ; this document assumes version 00. Version ff is forbidden
+HEXDIGLC         = DIGIT / "a" / "b" / "c" / "d" / "e" / "f" ; lowercase hex character
+value            = version "-" version-format
+version          = 2HEXDIGLC   ; this document assumes version 00. Version ff is forbidden
+version-format   = trace-id "-" child-id "-" trace-flags
+trace-id         = 32HEXDIGLC  ; 16 bytes array identifier. All zeroes forbidden
+child-id         = 16HEXDIGLC  ; 8 bytes array identifier. All zeroes forbidden
+trace-flags      = 2HEXDIGLC   ; 8 bit flags.
 ```
 
-#### Trace ID (`tid`)
+#### trace-id
 
 The format and requirements for this are the same as those of the `trace-id` field in the `traceparent` request header. This is a required parameter.
 
 For details, see the `trace-id` section under [traceparent Header Field Values](#traceparent-header-field-values).
 
-#### Child ID (`cid`)
+#### child-id
 
 This is the span ID of the server operation. It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. An all-zero child ID (`0000000000000000`) is an invalid value. Tracing systems MUST ignore the trace context metric when the child id is invalid (for example, if it contains non-lowercase hex characters).
 
 For details, see the `span-id` section under [traceparent Header Field Values](#traceparent-header-field-values).
 
-#### Version (`v`)
+#### version
 
-Version (`v`) is an 8-bit unsigned integer value, serialized as an ASCII string with two hexadecimal characters. Version 255 (`ff`) is invalid. This document specifies version 0 (`00`) of the trace context metric. The version field is optional; if omitted, the version is `00`.
+Version is an 8-bit unsigned integer value, serialized as an ASCII string with two hexadecimal characters. Version 255 (`ff`) is invalid. This document specifies version 0 (`00`) of the trace context metric.
 
-#### Trace Flags (`flags`)
+#### trace-flags
 
 Similar to the [`trace-flags` field](#trace-flags) in the `traceparent` request header, this is a hex-encoded <a data-cite='!BIT-FIELD#firstHeading'>8-bit field</a> that provides information about how a child handled the trace. The same requirement to properly mask the bit field value when interpreting it applies here as well.
 
@@ -66,8 +65,6 @@ These flags are recommendations given by a server, rather than strict rules for 
 3. Different load between services might force one or more participants to discard part or all of a trace.
 
 You can find more in the section [Security considerations](#security-considerations) of this specification.
-
-The trace flags param is optional. If it omitted, the value of all flags is unknown.
 
 ##### Sampled flag
 
--- a/spec/21-http_response_header_format.md
+++ b/spec/21-http_response_header_format.md
@@ -1,113 +1,107 @@
-# Trace Context HTTP Response Headers Format
+# Trace Context Server Timing Metric Format
 
-This section describes the binding of the distributed trace context to the `traceresponse` HTTP header.
+This section describes the binding of the distributed trace context to a metric in the Server Timing HTTP header.
 
-## Traceresponse Header
+## Trace Context Metric
 
-The `traceresponse` HTTP response header field identifies a completed request in a tracing system. It has four fields:
+The trace context metric identifies a completed request in a tracing system. It has four params:
 
-* `version`
-* `trace-id`
-* `child-id`
-* `trace-flags`
+* `tid` - required
+* `cid` - required
+* `v` - optional
+* `flags` - optional
 
-### Header Name
+Example server timing header with trace context metric:
 
-Header name: `traceresponse`
+```
+server-timing: trace;tid=0af7651916cd43dd8448eb211c80319c,cid=b7ad6b7169203331
+```
 
-The header name is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive). That is, `TRACERESPONSE`, `TraceResponse`, and `traceresponse` are considered the same header. The header name is a single word; it does not contain any delimiters such as a hyphen.
+### Metric Name
 
-In order to increase interoperability across multiple protocols and encourage successful integration, tracing systems SHOULD encode the header name as [ASCII lowercase](https://infra.spec.whatwg.org/#ascii-lowercase).
+Metric name: `trace`
 
-### traceresponse Header Field Values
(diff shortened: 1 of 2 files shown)


spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,107 @@
...
-The header name is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive). That is, `TRACERESPONSE`, `TraceResponse`, and `traceresponse` are considered the same header. The header name is a single word; it does not contain any delimiters such as a hyphen.
+```
+server-timing: trace;tid=0af7651916cd43dd8448eb211c80319c,cid=b7ad6b7169203331
+```
 
-In order to increase interoperability across multiple protocols and encourage successful integration, tracing systems SHOULD encode the header name as [ASCII lowercase](https://infra.spec.whatwg.org/#ascii-lowercase).
+### Metric Name
 
-### traceresponse Header Field Values
+Metric name: `trace`
```
dmathieu, line 19: As someone who hasn't attended the meetings (where I suppose this was discussed), this is a bit obscure. What does this name mean? Here, it's `trace`, but should it always be that? What would be other good values?

SergeyKanzhelev: server-timing is a set of metrics. This is one of them and we are trying to reserve the name `trace`.

Alternative naming proposal is here: https://github.com/w3c/trace-context/issues/561

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,107 @@
...
 
-### Header Name
+Example server timing header with trace context metric:
 
-Header name: `traceresponse`
-
-The header name is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive). That is, `TRACERESPONSE`, `TraceResponse`, and `traceresponse` are considered the same header. The header name is a single word; it does not contain any delimiters such as a hyphen.
+```
+server-timing: trace;tid=0af7651916cd43dd8448eb211c80319c,cid=b7ad6b7169203331
+```
```
basti1302, line 15: With respect to making the trace context available to client side JS, does that really work if we use arbitrary key value pairs like that? The way I used `Server-Timing` in the past was to always strictly stick with the `desc` and `dur` attributes from the [PerformanceServerTiming Interface](https://www.w3.org/TR/server-timing/#the-performanceservertiming-interface).

See also:
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing#performanceservertiming_interface

I would have expected that we were going to define _one_ metric (`trace`, `span`, `tracecontext`, `traceresponse`, which ever name we settle on) and put all the information we need into the `desc` field of that metric. That way, I believe it would be correctly exposed by browsers via the performance API. (Side note: We probably do not have a use for the `dur` field, so we would just omit it; or it could later be used to for the span duration.)

For reference, that is also what OpenTelemetry is discussing: https://github.com/open-telemetry/opentelemetry-specification/issues/3811:

```
Server-Timing: traceparent;desc="00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"
```

It is basically the previous `traceresponse` format, in the `desc` field of a `Server-Timing` metric. 

Of course, I might be missing something here. If it is possible to transport arbitrary key-value pairs like suggested in this PR and have them exposed to client side JS via the ServerTiming API (and browser support for that is on par with the `desc` attribute approach), the suggestion in this PR is cleaner.

basti1302: For reference, here is a quick test app I used:

```
const http = require('node:http');
const hostname = '127.0.0.1';
const port = 3000;

const server = http.createServer((req, res) => {
  res.statusCode = 200;
  res.setHeader('Content-Type', 'text/html');
  res.setHeader('Server-Timing', 'trace;tid=0af7651916cd43dd8448eb211c80319c,cid=b7ad6b7169203331');
  res.end(`
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>Server-Timing Test Page</title>
  <script>
    const metrics = performance.getEntriesByType('navigation')[0]['serverTiming'];
    console.log(JSON.stringify(metrics, null, 2));
  </script>
</head>
<body>
Ohai!
</body>
</html>
`);
});

server.listen(port, hostname, () => {
  console.log(`Server running at http://${hostname}:${port}/`);
});
```

This results in the following output in the browser's console (in Chrome):

```
[
  {
    "name": "trace",
    "duration": 0,
    "description": ""
  },
  {
    "name": "cid",
    "duration": 0,
    "description": ""
  }
]
```

So `trace` becomes a metric without any values, `cid` becomes a metric without any values and `tid` gets discarded.

Changing the setHeader line to `res.setHeader('Server-Timing', 'traceresponse;desc="00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-03"')` yields the following:

```
[
  {
    "name": "traceresponse",
    "duration": 0,
    "description": "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-03"
  }
]
```

dyladan: I think you're right. I'll update the PR

dyladan: Done in ca90939974154fbe77a80ad2c1316da3b1ea11b2

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,104 @@
-# Trace Context HTTP Response Headers Format
+# Trace Context Server Timing Metric Format
 
-This section describes the binding of the distributed trace context to the `traceresponse` HTTP header.
+This section describes the binding of the distributed trace context to a metric in the Server Timing HTTP header.
```
johnbley: I'd suggest `HTTP response header` (adding "response") to be super-clear.

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,104 @@
...
+The format and requirements for this are the same as those of the `trace-id` field in the `traceparent` request header. This is a required parameter.
 
-For details, see the trace-id section under [traceparent Header Field Values](#traceparent-header-field-values).
+For details, see the `trace-id` section under [traceparent Header Field Values](#traceparent-header-field-values).
 
 #### child-id
 
-This is the ID of the operation of the callee (in some tracing systems, this is known as the `span-id`, where a `span` is the execution of a client request) and is used to uniquely identify an operation within a trace. It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. All bytes as zero (`0000000000000000`) is considered an invalid value.
+This is the span ID of the server operation. It is represented as an 8-byte array, for example, `00f067aa0ba902b7`. An all-zero child ID (`0000000000000000`) is an invalid value. Tracing systems MUST ignore the trace context metric when the child id is invalid (for example, if it contains non-lowercase hex characters).
```
johnbley: I don't normally like to bikeshed names too much, but to me this descriptive text suggests that the name of this field should be changed to `span-id`.  It would read much more cleanly and sidestep any questions of semantics ("child of.... what?")

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,104 @@
...
 
 The current version of this specification (`00`) supports only two flags: `sampled` and `random-trace-id`.
 
-These flags are recommendations given by a callee rather than strict rules to follow for three reasons:
+These flags are recommendations given by a server, rather than strict rules for the client to follow, for three reasons:
 
-1. An untrusted callee may be able to abuse a tracing system by setting these flags maliciously.
-2. A callee may have a bug which causes the tracing system to have a problem.
-3. Different load between calling and called services might force one or more participants to discard part or all of a trace.
+1. An untrusted server may be able to abuse a tracing system by setting these flags maliciously.
```
johnbley, line 63: I like these examples of thinking through security implications.  I might add that simply exposing the trace id (and span id) itself to clients may present a risk in some scenarios.  Do we need text explicitly stating that using the response header (`server-timing: trace`) is optional for all participants and/or that compliant software SHOULD make emitting it configurable?

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,104 @@
...
   [Security considerations](#security-considerations) SHOULD be applied to protect from abusive or malicious use of this flag.
-- If a component deferred or delayed the decision and only a subset of telemetry will be recorded, the `sampled` flag from the incoming `traceparent` header should be used if it is available. It should be set to `0` as the default option when the trace is initiated by this component.
-- If a component receives a `0` for the `sampled` flag on an incoming request, it may still decide to record a trace. In this case it SHOULD return a `sampled` flag `1` on the response so that the caller can update its sampling decision if required.
+- If the server deferred or delayed the decision and only a subset of telemetry will be recorded, the `sampled` flag from the incoming `traceparent` header should be used if it is available. It should be set to `0` as the default option when the trace is initiated by this server.
+- If the server receives a `0` for the `sampled` flag on an incoming request, it may still decide to record a trace. In this case it SHOULD return a `sampled` flag `1` on the response so that the client can update its sampling decision if required.
```
johnbley, line 83: I like that this embodies the "I'm going to tell you, the client, how I actually traced this, insofar as I know it" spirit.

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,104 @@
...
+A participant that continues a trace started upstream &mdash; that is, if the participant uses the `trace-id` value from an incoming `traceparent` header in its own trace context server timing metric &mdash; MUST set the `random-trace-id` flag in the trace context server timing metric to the same value that was found in the incoming `traceparent` header.
 
-A participant that continues a trace started downstream &mdash; that is, if the participant uses the `trace-id` value from a `traceresponse` header it has received &mdash; MUST set the `random-trace-id` flag in its own `traceresponse` header to the same value that was found in the `traceresponse` header from which the `trace-id` was taken.
+A participant that continues a trace started downstream &mdash; that is, if the participant uses the `trace-id` value from a trace context server timing metric it has received &mdash; MUST set the `random-trace-id` flag in its own trace context server timing metric to the same value that was found in the trace context server timing metric from which the `trace-id` was taken.
```
johnbley, line 102: Perhaps a `Client Interpretation` section on how participating clients may choose to "extend" the trace by producing data with the same `trace-id` (as some browser instrumentation might do for page load), may choose to link the two traces in a system-appropriate way (e.g., as OTel span links or as simple Zipkin tags like `linked.traceId`), may choose to ignore it, or may choose to log or otherwise respond when its intended trace propagation was not honored ("I sent you a `traceparent` but your response shows you started a new trace")?

spec/21-http_response_header_format.md
dyladan: I think we will want to think through the security and privacy implications thoroughly before publishing. I'd rather handle that in its own dedicated issue to ensure we're giving it the consideration it deserves.

spec/21-http_response_header_format.md
dyladan: IMO that's a little bit out of scope since different tracing systems work differently. Since the server starts a span with no parent, we can't know how creating another span with no parent would affect the tracing system. I also don't want to track down the behavior of every tracing system since they could change. I would rather simply describe the format of the header and let the tracing systems decide for themselves the best way to handle it. For exmaple, one system might use a span link while another might allow multiple root spans in a single trace.

spec/21-http_response_header_format.md
```diff
@@ -1,113 +1,104 @@
...
+1. An untrusted server may be able to abuse a tracing system by setting these flags maliciously.
+2. A server may have a bug which causes the tracing system to have a problem.
+3. Different load between services might force one or more participants to discard part or all of a trace.
 
 You can find more in the section [Security considerations](#security-considerations) of this specification.
 
 ##### Sampled flag
 
-When set, the least significant bit (right-most), denotes that the callee may have recorded trace data. When unset, the callee did not record trace data out-of-band.
+When set, the least significant bit (right-most), denotes that the server may have recorded trace data. When unset, the server did not record trace data out-of-band.
```
balajikris, line 71: when we say LSB is right-most, how do we say that, does it not depend on the endian-ness?

spec/21-http_response_header_format.md
balajikris: John, purely for my learning, could you explain how exposing traceid could lead to a compromise?

balajikris (approved):

spec/21-http_response_header_format.md
johnbley: Objectively, the traceid/spanid is "just" some random bytes that aren't tied to any "really" sensitive information.  So, exposing it is not a compromise itself per se, but some organizations might view it as a risk.  Any leakage of information across a trust boundary might be undesired for somebody.  A far-fetched example: if an outsider has the ability to send requests into your system *and* has the ability to delete system monitoring records by traceid (but not the ability to search/query logged requests), then telling them the traceid for their specific request in the response means that they can combine their two abilities into being able to send malicious requests and then promptly remove the records for them.  Again, this seems far fetched - "if they already have the ability to delete system monitoring records we've already lost" - but security-sensitive organizations want control over such scenarios and try to analyze (and minimize) risk globally.  All I'm really advocating for here is that the owner of the tracing system be able to configure this functionality on/off, for security or any other reason.

SergeyKanzhelev (approved):

PR #580, opened 2026-03-15
Test multiple Python versions

martincostello:
- Update tests to run a matrix for 3.11-3.13.
- Update GitHub Actions to their latest versions.

### DIFF
--- a/.github/workflows/self_test.yml
+++ b/.github/workflows/self_test.yml
@@ -6,13 +6,23 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - '3.11'
+          - '3.12'
+          - '3.13'
+          - '3.14'
+
     name: Run the python tests
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Setup python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v6
         with:
-          python-version: 3.11.2
+          python-version: ${{ matrix.python-version }}
           architecture: x64
       - run: pip install aiohttp
       - run: cd test && python self_test.py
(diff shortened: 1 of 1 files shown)


.github/workflows/self_test.yml
```diff
@@ -6,14 +6,42 @@ on:
...
       - run: pip install aiohttp
       - run: cd test && python self_test.py
         shell: bash
+
+  self-test:
+    name: Run the python tests
+    needs: [ build ]
+    if: ${{ !cancelled() }}
+    env:
+      TESTS_SUCCESS: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
```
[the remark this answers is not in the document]
martincostello: This isn't true for a matrix.

.github/workflows/self_test.yml
```diff
@@ -6,14 +6,42 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          if [ "$TESTS_SUCCESS" == "true" ]
+          then
+            echo 'Self tests successful ✅'
+          else
+            echo 'Self tests failed ❌'
+            exit 1
+          fi
```
dyladan, line 47: What's the reason to gather the results of the other tests? If any one test fails it will be visible anyway

dyladan (commented):
Seems fine to me. Some projects use this as a test suite for their project and it is good to know they can use multiple versions. I would not make any promises about which python versions do and do not work though outside of making a reasonable effort to support recent versions.

.github/workflows/self_test.yml
martincostello: There's only one GitHub required status to manage _and_ it doesn't need to be changed for this PR.

Otherwise, the repo settings need to be changed for the additional test legs and re-configured over time as versions are added/removed.

dyladan (approved):

zacharycmontoya (approved):
LGTM as well

Kielek (approved):

maraisr (approved):

PR #585, opened 2026-07-06
Update tracestate example for header format

JoakimBrannstrom:
Clarified example of tracestate header formatting to ensure only the most recent position is included.

***
<a href="https://pr-preview.s3.amazonaws.com/JoakimBrannstrom/trace-context/pull/585.html" title="Last updated on Jul 6, 2026, 12:48 PM UTC (50d861e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trace-context/585/acab820...JoakimBrannstrom:50d861e.html" title="Last updated on Jul 6, 2026, 12:48 PM UTC (50d861e)">Diff</a>


spec/20-http_request_header_format.md
```diff
@@ -320,7 +320,7 @@ Example: `vendorname1=opaqueValue1,vendorname2=opaqueValue2`
 
 Tracing tools are not supposed to add the same header multiple times. For example, if a vendor name is Congo and a trace started in their system and then went through a system named Rojo and later returned to Congo, the `tracestate` value would not be:
 
-`congo=congosFirstPosition,rojo=rojosFirstPosition,congo=congosSecondPosition`
+`congo=congosSecondPosition,rojo=rojosFirstPosition,congo=congosFirstPosition`
```
zacharycmontoya, line 323: I don't see a reason to change this tracestate value, since the one several lines later in the doc is the correct one. This line is supposed to show a wrong example, so the previous value works just as well as the newly proposed value

JoakimBrannstrom: I know that this example shows an invalid tracestate, but it was confusing when I read it, since the order of creation should be from right to left. 
I believe it would show the problem in a better way by making the suggested change.

As a side-note, accepting PR's (if they are not flat out wrong) builds more engagement and makes the community stronger :)

